PR #1, opened 2010-11-18
Pt 6318273 replace merkle

jonmeredith:
For Andy/Dizzy mostly, but any of you repl lovers can enjoy.

These are the changes made for our client merged into the new framework using a helper process off the tcp client/servers for doing vnode folding/hash diffs.

-- Jon


slfritchie:
Jon, I've only the two questions posted in the respective commits.

jonmeredith:
For the first comment on c8d95f3, I think the undefined their_merkle_fn behavior is safe - you should only ever receive one {Ref, converted} message when you enter merkle_recv.

The files are cleaned up as we go for disk space reasons.  The fullsync creates 3xnum partitions of files.  We could keep them around I suppose.  The terminate cleans up if somebody drops the connection.  Although this is duplicated too as a new temp dir is created on startup and the old one removed.

slfritchie:
Jon -- it looks like I'd forgotten a couple of words, it should read: Is there ^a^ reason ^not^ to delete the file, like this clause does?

slfritchie:
Via IM ... server-side uses the open+create, unlink, use trick.  Client-side ... the new cleanup_partition() seems like it should take care of stragglers.  That wraps up my comments on comments.  Merge it?

PR #2, opened 2010-12-01
Bz882 cluster info

slfritchie:
Reviewer: anyone

In the name of laziness and time before code cutoff, I only added the
same stats that "riak-repl status" spits out, using the same trick for
riak_repl_console.erl that I used for riak_kv_console.erl.


beerriot:
+1 looks good and works for me

PR #3, opened 2010-12-02
Pt6529745 pause fullsync

jonmeredith:
Added a riak-admin pause-fullsync and riak-admin resume-fullsync command so fullsync replication can be disabled during unwanted periods.


jonmeredith:
Scott says "The pause/resume commit seems sane, too.  "

PR #7, opened 2010-12-23
Remove connector process

jonmeredith:
This patch moves the reconnect/failover functionality that was in riak_repl_connector into riak_repl_tcp_client.
It now uses the supervisor to make sure that the client is trying to reconnect on failure and is restarted rather than the monitor/ETS tables it was using before.

The problem seen at McAfee of multiple client connections being set up should no longer be possible.

Whoever reviews the patch, I'd recommend making sure
- add/remove sites works (and tears down any existing connections - which did not happen before)
- reconnection works after the leader server goes down
- riak-repl status looks good, I've improved it to add client stats and added the site names to both.
- doing generally nasty things to it and we do not leak file handles, generate temporary file garbage on
  the filesystem that does not get cleaned up on restart.

Cheers, Jon.


slfritchie:
I read over all the commit diffs, nothing to complain about.  Will start testing after lunch.

slfritchie:
No memory, disk, or file descriptor leaks.  I need to investigate if there's any reason why behavior that looks like BZ 759 is worse with these patches.

PR #8, opened 2011-01-03
Leader reg name change

jonmeredith:
Fix for the gen_leader messages reported in bz://960


dizzyd:
+1, based on code survey.

slfritchie:
Just a bit of housekeeping: is this pull request going to survive?  Discussion yesterday suggested that this patch would be abandoned?

jonmeredith:
The patch should survive so we can do rolling upgrades.

dizzyd:
+1.

PR #9, opened 2011-01-04
Repl by capability

jonmeredith:
Added detection of capability by version and changed bounded queue to only be enabled if 0.14.0 and above.


jonmeredith:
I used list_to_integer for rc1 suffixes, otherwise you get this...

1> list_to_integer("123rc1").
*\* exception error: bad argument
     in function  list_to_integer/1
        called as list_to_integer("123rc1")

jonmeredith:
Resubmitted as https://github.com/basho/riak_repl/pull/10

PR #10, opened 2011-01-04
Repl by capability

jonmeredith:
Resubmitted https://github.com/basho/riak_repl/pull/9 against riak_repl-0.14 now github is feeling better.

Selectively enables q_ack behavior depending on versions and makes forward compatible change to pass a capability list with the peerinfo.


dizzyd:
+1.

PR #11, opened 2011-01-04
Clear controller on new candidates

jonmeredith:
When the new repl leader gen_server gets a new list of candidates, it clears the leader name and restarts the gen_leader process.  The riak_repl_controller needs to be told so that it correctly sets up the client supervisor (otherwise you can have multiple client connections).


dizzyd:
+1.

PR #17, opened 2011-04-20
BZ1074 Don't close the repl server socket if the election is still going

Vagabond:
Previous to this commit if, for whatever reason, the gen_leader election
was taking less than a small amount of time and a replication connection
was established the repl_tcp_server would instantly close the
connection. This in turn led to the connecting client exiting and
hitting the supervisor's max restarts and having the supervisor restart.

This change makes the code wait up to 60 seconds for the result of an
election to to become known, checked in 5 second intervals. Until we
know what the status of the election is, we do not consume any data off
the socket (we put the socket in passive mode). If the timeout is hit,
we exit as before, but if the election completes we either start doing
replication or refer the client to the leader. A new starting state was
added to the FSM, send_peerinfo which is where the FSM waits until it
knows who the leader is. Once we know who the leader is we put the
socket in {active, once} mode and continue as before.

### DIFF
--- a/src/riak_repl_tcp_server.erl
+++ b/src/riak_repl_tcp_server.erl
@@ -18,7 +18,8 @@
          handle_info/3, 
          terminate/3, 
          code_change/4]).
--export([wait_peerinfo/2,
+-export([send_peerinfo/2,
+         wait_peerinfo/2,
          merkle_send/2,
          merkle_build/2,
          merkle_xfer/2,
@@ -52,7 +53,8 @@
           diff_errs,                                % errors retrieving different keys
           q :: undefined | bounded_queue:bounded_queue(),
           pending :: undefined | non_neg_integer(),
-          max_pending :: undefined | pos_integer()
+          max_pending :: undefined | pos_integer(),
+          election_timeout :: reference()          % reference for the election timeout
          }
        ).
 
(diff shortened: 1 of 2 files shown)


src/riak_repl_tcp_server.erl
```diff
@@ -82,30 +84,41 @@ set_socket(Pid, Socket) ->
...
+                          L#repl_listener.nodename =:= Node],
+    hd(NodeListeners).
+
+send_peerinfo(timeout, #state{socket=Socket, sitename=SiteName} = State) ->
     OurNode = node(),
     case riak_repl_leader:leader_node()  of
         undefined -> % leader not elected yet
-            stop;
+            %% check again in 5 seconds
+            erlang:send_after(5000, self(), timeout),
```
jonmeredith: This message gets delivered after the handshake has completed and crashes other states (e.g. merkle_build).
Perhaps it should be renamed election_retry and only delivered if in the send_peerinfo state.

The other option would be to use send_event_after and ignore it in all the other states (shame there is no send_all_state_event_after).

src/riak_repl_tcp_server.erl
```diff
@@ -82,30 +84,41 @@ set_socket(Pid, Socket) ->
...
     OurNode = node(),
     case riak_repl_leader:leader_node()  of
         undefined -> % leader not elected yet
-            stop;
+            %% check again in 5 seconds
+            erlang:send_after(5000, self(), election_wait),
+            {next_state, send_peerinfo, State};
         OurNode ->
-            send(Socket, {peerinfo, PeerInfo});
+            erlang:cancel_timer(State#state.election_timeout),
```
jonmeredith, line 105: Even if you cancel the election timeout there's a chance it could fire and leave the message in the inbox before it is cancelled.
Could fix by setting election_timeout to undefined in the final state and only delivering the handle_info message if it is /= undefined.

jonmeredith:
+1 merge.

Tested with basho_expect against riak_ee-0.14.2rc4 with the following commits applied

From the monitor branch
9ff4fd2cb014ec6fc3a6d7ef35d27e3dfb78d5ef
8b23a2df88dbfca54cb682e30e0f81190b0baf81

From the wait on election branch
49d735352426d2acacfb8acdd0b7a431b23fd567
4a709ec1bcc41928477cec46cc8c26fe27152ba6
57c10d15108fc6bea47766cc0b591e33f4ded9fb

Vagabond:
Merged

PR #19, opened 2011-07-26
Reconcile split brain

Vagabond:


jonmeredith:
I set up 4 VMs the one that was elected as leader I disconnected the virtual NIC and waited until I saw disconnected messages on all the nodes then re-enabled the NIC.   Once the nodes reconnected there were two leaders, i.e. no re-election took place.  This sounds very similar to your physical plugging test - it isn't a 'fiddle with the cookie' kind of a test.

PR #20, opened 2011-08-01
Az573 use kv fsm for repl

rzezeski:
Instead of having what is essentially a copy of the KV put coordinator in repl use the KV put coordinator directly.


slfritchie:
Ryan: basho_expect's Run-systest-riak-repl-2x2 test passes, +1 to merge, thanks.

PR #21, opened 2011-08-04
Bz1146 reconcile split brain 0.14

Vagabond:


src/gen_leader.erl
```diff
@@ -740,6 +769,56 @@ loop(#server{parent = Parent,
                         false ->
                             loop(Server,Role,E,Msg)
                     end;
+                {election} ->
+                    %% Told to do an election because of a leader conflict.
+                    Mide = E,
+                    E1 = startStage1(Mide),
+                    safe_loop(Server, candidate, E1, Msg);
+                  {checklead, Node} ->
```
jonmeredith: Indent pedantry - somebody needs electric indent :)

jonmeredith:
+1 merge in combination with repl leader eqc test enhancements.
Only issue I can see is that a {send_checklead} could sit in the mailbox waiting for safe_loop to complete, but that's probably  not a big deal.

PR #22, opened 2011-08-04
repl_leader_eqc fixes for riak_repl-0.14

jonmeredith:
Fixes to the leadership eqc test made during bz://1146, zd://773.

Gives additional time for elections to take place.
Ensures election completed after restarting nodes.
Does not expect elections to happen when all candidate nodes are down.
Fixes toggle_type to change between candidate/worker correctly rather than just promoting to candidate.

Against 0.14 branch.

### DIFF
--- a/test/repl_leader_eqc.erl
+++ b/test/repl_leader_eqc.erl
@@ -40,7 +40,7 @@
         eqc:on_output(fun(Str, Args) -> io:format(user, Str, Args) end, P)).
 
 -define(DBG(Fmt,Args),ok).
-%-define(DBG(Fmt,Args),io:format(user, Fmt, Args)).
+%% -define(DBG(Fmt,Args),io:format(user, Fmt, Args)).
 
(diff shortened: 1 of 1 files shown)


test/repl_leader_eqc.erl
```diff
@@ -539,7 +577,9 @@ register_receiver() ->
                                 ok
                         end
                 end),
+    Leader1 = riak_repl_leader:leader_node(),
     Res = riak_repl_leader:add_receiver_pid(Pid),
-    {Res, Pid}.
+    Leader2 = riak_repl_leader:leader_node(),
```
Vagabond, line 582: Why do you get the leader node twice here?

jonmeredith, line 582: I did it out of paranoia that the election could happe before/after my add_receiver_pid call so I wanted to make extra sure it was safe.  There was a counterexample it fixed.

PR #23, opened 2011-08-04
repl_leader_eqc fixes against master

jonmeredith:
Fixes to the leadership eqc test made during bz://1146, zd://773.

Gives additional time for elections to take place.
Ensures election completed after restarting nodes.
Does not expect elections to happen when all candidate nodes are down.
Fixes toggle_type to change between candidate/worker correctly rather than just promoting to candidate.

Against master.  'git am' didn't automatically merge the second commit correctly so there will be a minor difference in commit ids.


jonmeredith:
+1 merge with repl eqc test fixes.

PR #24, opened 2011-08-08
Make replication Search aware

rzezeski:
az551

This commit enables replication to successfully replicate all
Search related data no matter how indexing has occured.  That is,
whether it was an object stored in a bucket with the search hook
installed, a local document installed via `search-cmd` or a solr
document replication will see it and replicate it.  This also goes
for deleting a document.

The key to getting this to work is to teach replication to recognize
Search "proxy objects."  A proxy object (PO) is one that encapsultes all
the postings for an indexed document (whether it be a KV object or
local file) so that they may be deleted later.  When a replication
client sees a PO come across the wire it knows that the postings
inside it must be enetered into the local index.  If the incoming
object is a tombstone then replication knows it needs to
remove the entries.

Things are only slightly more complicated for an indexed KV object.
In the case of a fullsync there is a chance that there could be a great
delay between the replication of the KV oject and the PO.  Furtermore,
you want to replicate the PO before it's corresponding KV object to
stay consistent with how it works on a single cluster.  This means that
during fullsync replication will perform checks to determine if the object
it is about to replicate is a member of a pair of KV/PO objects that must
be replicated together.  If it is then fullsync will make sure to
replicate both and send them in the preferred order.

Initially I had planned on using some form of caching to determine
if an object is a PO or not but I have decided against this.  First,
the Search hook could be installed onto a bucket during fullsync
which causes the cache to become invalid.  Second, I'm not convinced
the code to determine PO status is dominant.  Third, it would complicate
the code causing more opportunity for failure.  Fourth, it's in
the works to completely rewrite replication.

### DIFF
--- a/src/riak_repl_tcp_client.erl
+++ b/src/riak_repl_tcp_client.erl
@@ -401,14 +401,14 @@ update_site_ips(TheirReplConfig, SiteName) ->
     ok.    
 
 do_repl_put(Obj, State=#state{ack_freq = undefined}) -> % q_ack not supported
-    riak_repl_util:do_repl_put(Obj),
+    spawn(riak_repl_util, do_repl_put, [Obj]),
     State;
 do_repl_put(Obj, State=#state{count=C, ack_freq=F}) when (C < (F-1)) ->
-    riak_repl_util:do_repl_put(Obj),
+    spawn(riak_repl_util, do_repl_put, [Obj]),
     State#state{count=C+1};
 do_repl_put(Obj, State=#state{socket=S, ack_freq=F}) ->
     send(S, {q_ack, F}),
-    riak_repl_util:do_repl_put(Obj),
+    spawn(riak_repl_util, do_repl_put, [Obj]),
     State#state{count=0}.
     
 %% Decide when it is time to leave the merkle_recv state and whether
--- a/src/riak_repl_tcp_server.erl
+++ b/src/riak_repl_tcp_server.erl
@@ -82,7 +82,7 @@ status(Pid, Timeout) ->
 
 set_socket(Pid, Socket) ->
     gen_fsm:sync_send_all_state_event(Pid, {set_socket, Socket}).
-    
+
 init([SiteName]) ->
     {ok, send_peerinfo, #state{sitename=SiteName}}.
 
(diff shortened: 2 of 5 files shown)


src/riak_repl_tcp_client.erl
```diff
@@ -401,14 +401,14 @@ update_site_ips(TheirReplConfig, SiteName) ->
     ok.    
 
 do_repl_put(Obj, State=#state{ack_freq = undefined}) -> % q_ack not supported
-    riak_repl_util:do_repl_put(Obj),
+    spawn(riak_repl_util, do_repl_put, [Obj]),
```
jonmeredith, line 404: Why are these spawn's here?  It will create processes outside the supervision tree that will be very had to track down.
Instead you could start children with riak_kv_put_fsm_sup like riak_client does.  That way we'll be able to track them in the process tree.
There's also the issue of knowing if repl puts are succeeding or failing - but we may not have been tracking that before.

jonmeredith, line 404: Also, spawning the repl puts means that you lose the back pressure the acking system is designed to provide.  If the local cluster can't keep up with the put load then you'll back up.

src/riak_repl_tcp_server.erl
```diff
@@ -431,14 +431,68 @@ maybe_send(RObj, ClientVC, Socket) ->
...
+    case P of
+        [] -> false;
+        _ ->
+            case lists:any(fun search_pre_hook/1, P) of
+                true -> true;
+                false -> false
+            end
+    end.
+
+search_pre_hook({struct, [{<<"mod">>, <<"riak_search_kv_hook">>},_]}) -> true;
```
jonmeredith, line 497: What if search is not the first hook?

src/riak_repl_tcp_server.erl
```diff
@@ -431,14 +431,68 @@ maybe_send(RObj, ClientVC, Socket) ->
...
+    end.
+
+check_and_send(Sock, {diff_obj, O}=Data) ->
+    B = riak_object:bucket(O),
+    K = riak_object:key(O),
+    {ok, C} = riak:local_client(),
+    PO = is_proxy_object(B),
+    SE = is_search_enabled(C, B),
+
+    if PO ->
```
jonmeredith: Maybe better written as 

<pre>
if 
  PO andalso SE ->  send_search(po, O, B, K, C, Sock);
  SE -> send_search(kv, O, B, K, C, Sock);
  true ->  send(Sock, term_to_binary(Data))
end
</pre>

jonmeredith: You could even change send_search to take a true/false for the is proxy object and just do this (for simple folk like me who have have to train their brains to read the catchall true in if statements)

<pre>
case SE of 
  true ->  send_search(PO, O, B, K, C, Sock);
  false ->  send(Sock, term_to_binary(Data))
end
</pre>

jonmeredith: Then you could add {ok, C} =riak_client().  Inside the true case.  I'm stopping now :)

src/riak_repl_tcp_client.erl
jonmeredith, line 404: I see the spawns were in the original code.  Bad bad bad. So much for the backpressure argument - we can make the put async. The supervision stuff still stands.

src/riak_repl_util.erl
```diff
@@ -44,8 +44,46 @@ get_partitions(_Ring) ->
 
 do_repl_put(Object) ->
     ReqId = erlang:phash2(erlang:now()),
-    spawn(riak_kv_put_fsm, start,
-          [ReqId, Object, 1, 1, ?REPL_FSM_TIMEOUT, self(), [disable_hooks]]).
+    B = riak_object:bucket(Object),
+    K = riak_object:key(Object),
+    Opts = [disable_hooks, {update_last_modified, false}],
+
+    case binary:matches(B, <<"_rsid_">>) of
```
jonmeredith: this will do more work that you need - looking for _rsid_ anywhere in the binary.  This matches the start of the binary for less work.

``` erlang
case B of
  << "_rsid_", Idx/binary>> ->
    matching_case;
  _ ->
    non_matching_case
end
```

src/riak_repl_tcp_client.erl
rzezeski, line 404: So which do you want to do?  Keep the spawn's or make do_repl_put sync?

src/riak_repl_tcp_server.erl
rzezeski, line 497: Notice this is called via `lists:any`.

src/riak_repl_tcp_server.erl
rzezeski: Actually, I can't move `{ok, C} = riak:local_client()` because `is_search_enabled` relies on it.  I like the other refactorings.

src/riak_repl_tcp_client.erl
jonmeredith, line 404: We can keep the spawns - I'm concerned it would impact performance otherwise.  Need to add the sup.

src/riak_repl_tcp_server.erl
jonmeredith, line 497: derp.  missed that.  my red pen got out of control.

jonmeredith:
Pull request looks good - I've been able to manually index search-only and search/kv data and seen it replicate.
I just need to verify there are no performance regressions.

Currently the check for search being enabled retrieves the bucket properties, which put also does.  There's a performance optimization to change the put FSM to accept bucket parameters in it's options, but I think we can handle that in a separate pull request.

jonmeredith:
Finishing up review - I was able to trigger the {error, notfound} case during fullsync and triggered a crash with no write load on the cluster.  It wasn't reproducible, but happening once is enough that it should be explicitly handled.  I believe it could happen during fullsync although would still be hard.

Also, looking at the send proxy object case, I think we should try and send the K/V object afterwards if it exists to ensure consistency.

After those two changes it looks good to merge.

jonmeredith:
I eat my words on the send proxy case.  It's fine.

jonmeredith:
+1 merge.

PR #25, opened 2011-08-25
Change repl to use a fixup_bucket function

Vagabond:


src/riak_repl.erl
```diff
@@ -18,7 +19,38 @@ install_hook() ->
...
+strip_postcommit(BucketProps) ->
+    %% Get the current postcommit hook
+    case proplists:get_value(postcommit, BucketProps, []) of
+        X when is_list(X) ->
+            CurrentPostcommit=X;
+        {struct, _}=X ->
+            CurrentPostcommit=[X]
+    end,
     
+    %% Add kv/search hook - make sure there are not duplicate entries
```
rzezeski: What's up with the search reference?

rzezeski:
+1 to merge

src/riak_repl.erl
```diff
@@ -18,7 +19,38 @@ install_hook() ->
...
-                        proplists:delete(postcommit, DefaultBucketProps)),
-    riak_core_bucket:append_bucket_defaults([{postcommit, [?REPL_HOOK]}]),
+                        proplists:delete(repl, DefaultBucketProps)),
+    riak_core_bucket:append_bucket_defaults([{repl, true}]),
     ok.
+
+fixup(_Bucket, BucketProps) ->
+    case proplists:get_value(repl, BucketProps) of
+        true ->
+            CleanPostcommit = strip_postcommit(BucketProps),
```
jonmeredith: Clean/Update portions are common to both cases - but duplication isn't huge.

jonmeredith:
Looks good to me - fiddled around with the HTTP interface to get/set repl properties.
+1 merge after I've checked the core/search stuff.

PR #27, opened 2011-09-06
Add infinity timeouts on all gen_server2:calls - they are all local.

jonmeredith:
With very large keys-per-partition update_many2 has been timing out and throwing errors, crashing riak_repl_fullsync_helper.

zd://715


slfritchie:
+1 to merge, thanks!

PR #28, opened 2011-10-04
Repl fullsync resume

Vagabond:
Tracks the outstanding partitions in a fullsync in case the fullsync aborts before it completes. Also adds some extra logging.

### DIFF
--- a/src/riak_repl_tcp_client.erl
+++ b/src/riak_repl_tcp_client.erl
@@ -262,11 +262,14 @@ merkle_diff({Ref, diff_done}, State=#state{our_kl_ref = Ref}) ->
                                    our_kl_ref = Ref})}.
 
 handle_info({tcp_closed, Socket}, _StateName, #state{socket = Socket} = State) ->
+    lager:info("Connection to site ~p closed", [State#state.sitename]),
     cleanup_and_stop(State);
 handle_info({tcp_closed, _Socket}, StateName, State) ->
     %% Ignore old sockets - e.g. after a redirect
     {next_state, StateName, State};
-handle_info({tcp_error, Socket, _Reason}, _StateName,  #state{socket = Socket} = State) ->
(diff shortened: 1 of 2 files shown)


src/riak_repl_tcp_client.erl
```diff
@@ -262,11 +262,14 @@ merkle_diff({Ref, diff_done}, State=#state{our_kl_ref = Ref}) ->
                                    our_kl_ref = Ref})}.
 
 handle_info({tcp_closed, Socket}, _StateName, #state{socket = Socket} = State) ->
+    lager:info("Connection to site ~p closed", [State#state.sitename]),
```
jonmeredith, line 265: Should be connection _for_ site instead of _to_ site on the client messages - limit confusion on site naming 100% as much as possible.

src/riak_repl_tcp_server.erl
```diff
@@ -84,7 +84,17 @@ set_socket(Pid, Socket) ->
     gen_fsm:sync_send_all_state_event(Pid, {set_socket, Socket}).
 
 init([SiteName]) ->
-    {ok, send_peerinfo, #state{sitename=SiteName}}.
+    State0 = #state{sitename=SiteName},
+    %% check if there's a previously incomplete fullsync
+    State = case application:get_env({progress, SiteName}) of
+        {ok, Partitions} when is_list(Partitions) ->
+            lager:notice("Resuming incomplete fullsync, ~p partitions remain",
```
jonmeredith: Knowing the site name would be useful in case of multi-site repl.

jonmeredith:
do_cancel_fullsync should also clean up progress - otherwise if the sync is cancelled and then the remote site reconnects it will resume at the original point rather than the start.  (Doesn't really matter, just principle of least surprise stuff).

jonmeredith:
+1 merge based on diff against #30

PR #30, opened 2011-10-04
Az794 repl fullsync resume 0.14

Vagabond:
az794 zd823

Also adds some extra logging.


jonmeredith:
+1 merge / send hotfix

PR #31, opened 2011-11-01
Fix repl between 1.0 and 0.14 clusters

Vagabond:
This prevents future changes to the ring from breaking if both repl
clusters are not running the same version of riak.


Vagabond:
Oh, I verified repl works between 0.14->1.0->1.0->0.14 with this change.

jonmeredith:
+1 merge.

Verified 0.14server &1.0 client, 1.0 server & 0.14 client and 1.0 client/server.

PR #32, opened 2011-11-02
Repl improvements phase 1

Vagabond:
Fix the supervisor tree
Audit modules for bad naming
Remove riak_repl_controller


src/riak_repl_leader.erl
```diff
@@ -87,6 +92,7 @@ helper_pid() ->
 
 init([]) ->
     process_flag(trap_exit, true),
+    erlang:send_after(0, self(), update_leader),
```
massung, line 95: This seems like a tiny little hack. My only concern is that there might actually be a race condition here. Is there any other way to update the leader after all the "initialization" is done some other way?

src/riak_repl_leader.erl
```diff
@@ -99,6 +105,9 @@ handle_call({add_receiver_pid, _Pid}, _From, State) ->
 handle_call(leader_node, _From, State) ->
     {reply, State#state.leader_node, State};
 
+handle_call(is_leader, _From, State) ->
+    {reply, State#state.i_am_leader, State};
+
```
massung, line 110: This also feels like another race condition. I have no solution, and perhaps it's always been this way, but it very much feels like - since each machine tracks its own leadership status - that one could easily setup a situation where there is no leader for a period of time or there are multiple leaders for a period of time (if leadership status is handed off). Are there [negative] consequences to this happening?

Vagabond, line 110: It's not the end of the world if there are multiple leaders, or no leaders. This can actually happen in normal operation with gen_leader anyway, so "fixing" it here (if it was even possible) isn't going to get you anything.

src/riak_repl_listener.erl
```diff
@@ -43,17 +46,26 @@ new_connection(Socket, State) ->
             connection_error(Reason, "unknown", State)
     end.
 
-stop(Pid) when is_pid(Pid)  ->
+stop(Pid) when is_pid(Pid) ->
     gen_server:cast(Pid, stop).
 
 close_all_connections() ->
     [exit(P, kill) || {_, P, _, _} <- supervisor:which_children(riak_repl_server_sup)].
```
massung, line 54: More of a question w/ naming. Should a "listener" be allowed to kill all the processes?

Vagabond, line 54: So, this is called when the current node loses its leader status. You don't want existing repl sessions to continue, so they all get killed. This code hasn't actually changed from before.

massung:
Everything appears to be fine. However, I did get a consistent crash when trying to get the configuration from a [new] ring multiple times. I'm trying to create repro steps for this, but I haven't gotten it to happen again. Symptom:

{ok,R}=riak_core_ring_manager:get_my_ring().
C=riak_repl_ring:get_repl_config(R).
*\* exception error: {badrecord,chstate_v2}
    in function  riak_core_ring:get_meta/2
    in call from riak_repl_ring:get_repl_config/1

Once I have a repro I'll follow-up.

massung:
Gonna +1. I can't get the above to happen again. Tried everything I initially did when learning to setup repl for the review and more and still couldn't get it to happen. My guess is that there is a bug out there (I'll keep looking), but the issue appears to be in riak_core_ring instead of repl, and I'm fairly confident that this PR is safe.

PR #33, opened 2011-12-05
Replication rework

Vagabond:
This major rework of the replication code provides the following benefits:
- Per bucket replication
- Generic hooks for per-object replication decisions (eg search ensuring the proxy object is replicated along with the KV object).
- Split fullsync and realtime replication into different modules, so fullsync can be pluggable.
- Implement a new 'keylist' fullsync strategy.
- Make the TCP send/recv buffers tunable.

Please treat this as a full code review, don't just look at the changes, because these are very extensive modifications.


beerriot:
Something that would help reviewing, and I think would also be a big improvement in its own right, would be an English readme describing how this version of replication works.  The `replication.org` file in `riak_ee/doc` gives something of a high-level explanation, but what I'm hoping for is something a little bit more specific about the guts: what each module does, where apps hook into repl, details of the inter-cluster protocol, etc.  This can all be learned by reading the code (though more doc'ing there could also help), but I think an overall doc could help both in learning, and also in spotting inconsistencies between how we think it works, and how it actually does.

Vagabond:
Yes, this is actually something I meant to do, but I forgot. Will type something up.

Vagabond:
Added documentation, and opened a new P/R on riak_ee for the details on the replication hooks in the replication.org file.

src/riak_repl_tcp_server.erl
```diff
@@ -147,373 +189,61 @@ wait_peerinfo({peerinfo, TheirPeerInfo, Capability},
...
-            case lists:any(fun search_pre_hook/1, P) of
-                true -> true;
-                false -> false
-            end
+send_peerinfo(#state{socket=Socket, sitename=SiteName} = State) ->
+    OurNode = node(),
+    case riak_repl_leader:leader_node()  of
+        undefined -> % leader not elected yet
+            %% check again in 5 seconds
+            erlang:send_after(5000, self(), election_wait),
```
beerriot, line 229: It looks like the code to handle this message has been removed.  Grepping for `election_wait` finds only this line.

src/riak_repl_tcp_server.erl
```diff
@@ -147,373 +191,61 @@ wait_peerinfo({peerinfo, TheirPeerInfo, Capability},
...
+            {noreply, State};
+        OurNode ->
+            erlang:cancel_timer(State#state.election_timeout),
+            %% this switches the socket into active mode
+            Props = riak_repl_fsm_common:common_init(Socket),
+            PI = proplists:get_value(my_pi, Props),
+            send(Socket, {peerinfo, PI,
+                [bounded_queue, {fullsync_strategies,
+                        app_helper:get_env(riak_repl, fullsync_strategies,
+                            [?LEGACY_STRATEGY])}]}),
```
beerriot: This call might return `{error, not_leader}` if the leader has changed since the call to `riak_repl_leader:leader_node/0`.

src/riak_repl_tcp_server.erl
```diff
@@ -147,373 +191,61 @@ wait_peerinfo({peerinfo, TheirPeerInfo, Capability},
...
+    OurNode = node(),
+    case riak_repl_leader:leader_node()  of
+        undefined -> % leader not elected yet
+            %% check again in 5 seconds
+            erlang:send_after(5000, self(), election_wait),
+            {noreply, State};
+        OurNode ->
+            erlang:cancel_timer(State#state.election_timeout),
+            %% this switches the socket into active mode
+            Props = riak_repl_fsm_common:common_init(Socket),
```
beerriot: The return value of `send/2` is ignored here, but could be either `ok` or `{error, Reason}`.

src/riak_repl_fullsync_helper.erl
```diff
@@ -125,13 +149,16 @@ handle_call({make_keylist, Partition, Filename}, From, State) ->
...
         true ->
-            {ok, FP} = file:open(Filename, [read, write, binary, raw, delayed_write]),
+            {ok, FP} = file:open(Filename, [read, write, binary, delayed_write]),
             Self = self(),
             Worker = fun() ->
                              %% Spend as little time on the vnode as possible,
                              %% accept there could be a potentially huge message queue
                              Folder = fun(K, V, MPid) -> 
-                                              gen_server2:cast(MPid, {kl, K, hash_object(V)}),
+                                              %gen_server2:cast(MPid, {kl, K, hash_object(V)}),
```
beerriot: This code should be removed if unused.  Also, it seems to be the only place casting the `{kl, K, H}` tuple, so if it's truly unused, the handle_cast clause for this tuple should be removed as well.

src/riak_repl_fullsync_helper.erl
```diff
@@ -321,10 +354,18 @@ open_couchdb(Filename) ->
     {ok, Fd, Btree}.
 
 itr_new(File, Tag) ->
+    case itr_fresh(File, Tag) of
```
beerriot: It looks like this is the only call to `itr_fresh/2`. Why add the extra fun wrapper if it's just going to be evaluated immediately like it used to be?

beerriot:
Further comments on `riak_repl_fullsync_helper`, from full review, not related to changes in this PR:
- The process is `gen_server2` behavior, but `gen_server` calls are made in a few places
- The `{make_keylist, Partition, Filename}` clause of `handle_call` does uses both a `gen_server:reply` and a `{reply, Reply, State}` tuple, which seems like potential for trouble. (Why did I also think it was necessary to return a `noreply` result _before_ calling `gen_server:reply`?)
- The use of R/remote and L/local in `diff_hash/2` is confusing when used in the new `keylist` strategy, since the code is running on the "server" side of repl, but the Remote argument is the server's keylist.

src/riak_repl_util.erl
```diff
@@ -44,48 +52,33 @@ get_partitions(_Ring) ->
...
     K = riak_object:key(Object),
-    Opts = [asis, disable_hooks, {update_last_modified, false}],
+    case repl_helper_recv(Object) of
+        ok ->
+            ReqId = erlang:phash2(erlang:now()),
+            B = riak_object:bucket(Object),
+            K = riak_object:key(Object),
+            Opts = [asis, disable_hooks, {update_last_modified, false}],
+
+            riak_kv_put_fsm_sup:start_put_fsm(node(), [ReqId, Object, 1, 1,
```
beerriot, line 64: Should this link to the Pid that is returned, so the worker tears down if the put fsm barfs?  I guess the timeout in the receive will trigger otherwise…

src/riak_repl_util.erl
```diff
@@ -93,6 +86,111 @@ reap(ReqId, B, K) ->
                                       [ReqId, B, K, 1, ?REPL_FSM_TIMEOUT,
                                        self()]).
 
+wait_for_response(ReqId) ->
+    receive
+        {ReqId, _} ->
+            ok
+    after 60000 ->
```
beerriot, line 95: Why 60sec instead of `?REPL_FSM_TIMEOUT`?

src/riak_repl_util.erl
```diff
@@ -93,6 +86,111 @@ reap(ReqId, B, K) ->
                                       [ReqId, B, K, 1, ?REPL_FSM_TIMEOUT,
                                        self()]).
 
+wait_for_response(ReqId) ->
+    receive
+        {ReqId, _} ->
```
beerriot, line 93: Should the reply be checked, with maybe a `lager:debug` if the put failed?

src/riak_repl_tcp_server.erl
```diff
@@ -1,144 +1,188 @@
...
+            {noreply, State};
+        cancel ->
+            {noreply, State}
+    end;
+handle_info({repl, RObj}, State) ->
+    case riak_repl_util:repl_helper_send_realtime(RObj, State#state.client) of
+        Objects when is_list(Objects) ->
+            %% enqueue all the objects the hook asked us to send
+            NewState = lists:foldl(fun(O, S) ->
+                        enqueue(term_to_binary({diff_obj, O}), S)
```
beerriot: This seems like it's likely to break the guarantees made to repl helpers/hooks about replicating some objects before others.  The `bounded_queue` module dequeues items to make space for new items when its limit is reached.  So, if the repl helper/hook returned a sufficiently large list of objects, or a list of sufficiently-large-size objects, and the repl link was unable to keep up, the objects early in the in that list would be dropped on the floor, while the objects later in the list would be replicated.

That is:

```
22> Q0 = bounded_queue:new(3).       
{bq,3,0,{[],[]},0}
23> Q1 = bounded_queue:in(Q0, <<1>>).
{bq,3,1,{[<<1>>],[]},0}
24> Q2 = bounded_queue:in(Q1, <<2>>).
{bq,3,2,{[<<2>>],[<<1>>]},0}
25> Q3 = bounded_queue:in(Q2, <<3>>).
{bq,3,3,{[<<3>>,<<2>>],[<<1>>]},0}
% Q3 is at capacity

26> bounded_queue:in(Q3, <<4>>). % add one more small entry
{bq,3,3,{[<<4>>,<<3>>],[<<2>>]},1}
% The first entry, <<1>>, has been lost

27> bounded_queue:in(Q3, <<5,6,7,8>>). % add one larger-than-capacity entry
{bq,3,4,{[<<5,6,7,8>>],[]},3}
% All previous entries have been lost
```

Vagabond: This is an excellent point, do you have any suggestions for solving it? Maybe just sending those objects directly and only enqueueing the original object?

beerriot: The only functional workaround I can think of is to add all of the "dependencies" and the root object as one, combined element in the queue.  The combined element may be large enough to kick other things out, but they would either all make it to send time, or none of them would.

The nonfunctional "workaround" was to change `bounded_queue` so that it refuses to enqueue new items, instead of dropping old ones, when its threshold is reached.  It prevents replicating an object before its dependencies, but may also prevent _ever_ replicating an object (because its dependencies fill the queue).

Vagabond: Yes, I had the same solution come to me in a dream.

Vagabond: Alright, I pushed a patch implementing this, with a test (and some manual confirmation it works).

beerriot:
+1 merge (see AZ ticket for testing details/omissions)

src/bounded_queue.erl
```diff
@@ -43,31 +43,38 @@
...
 in(BQ=#bq{m=Max, q=Q, d=D}, Item) when is_binary(Item) ->
-    ItemSize = size(Item),
+    ItemSize = element_size(Item),
     case ItemSize > Max of
         true ->
             BQ#bq{q=queue:from_list([Item]),s=ItemSize,d=queue:len(Q)+D};
         false ->
             make_fit(BQ, Item, ItemSize)
+    end;
+in(BQ=#bq{m=Max, q=Q, d=D}, [H|_T] = Items) when is_binary(H) ->
```
jonmeredith, line 62: Could do with a comment to explain the list is added as an item in the queue, rather than adding a list of items to the queue individually.

PR #36, opened 2012-01-09
Add worker pool for doing GETs

Vagabond:
Also, add tunables for setting the get/put worker pool sizes, and make repl wait for riak_kv to start.


jonmeredith:
As discussed with Diz, we're making a special exception for this 'performance bug' after the code freeze.

jonmeredith:
+1 merge

PR #37, opened 2012-01-30
Various minor cleanups to repl

Vagabond:
- Don't fetch vector clocks when they aren't needed
- Switch all lager:notices to infos
- Log the sitename in all fullsync messages
- Various improvements in cleaning up stale files
- Avoid ring_trans if there's nothing to do
- Comment on the algorithim used for choosing the strategy


jonmeredith:
+1 merge

PR #38, opened 2012-01-30
Avoid process leaks

Vagabond:


Vagabond:
When Grant did 100 million key repl testing he was seeing 'out of processes' errors. I found that the worker pools weren't being cleaned up correctly if the tcp connection stopped normally (eg. on network error) because {stop, normal, State} doesn't stop linked children. Use the terminate clause to make sure the fullsync worker process and the client PUT pool exit along with the TCP connection.

Also, fixed a silly bug that would make 'riak-repl status' fail when there was no fullsync process.

jonmeredith:
Got this error when I took down the server connection

```
14:22:48.181 [error] gen_server <0.1574.0> terminated with reason: {normal,{gen_fsm,sync_send_all_state_event,[<0.1603.0>,stop]}}
14:22:48.182 [notice] Starting replication with site {repl_site,"c2",[{"127.0.0.1",9011}],{0,0,0}}
14:22:48.183 [error] CRASH REPORT Process <0.1574.0> with 0 neighbours crashed with reason: {normal,{gen_fsm,sync_send_all_state_event,[<0.1603.0>,stop]}}
14:22:48.186 [error] Supervisor riak_repl_client_sup had child "c2" started with riak_repl_tcp_client:start_link("c2") at <0.1574.0> exit with reason {normal,{gen_fsm,sync_send_all_state_event,[<0.1603.0>,stop]}} in context child_terminated
```

Tracked to a bug in the poolboy:handle_sync_event return - Reply/ExitReason are transposed in the return tuple.

jonmeredith:
Poolboy fix did change things, now I get a bunch of this message on the server side.  It's really quite unhappy.

https://gist.github.com/db61a8f71e1f894df24b

jonmeredith:
+1 merge.

PR #39, opened 2012-02-01
Fix repl EQC test broken by refactoring

Vagabond:


jtuple:
Might want to consider cleaning up all the commented out code in the diff, but otherwise. Code looks fine, and eunits pass for me on OS X (top level "rebar eunit") and on Linux (running all deps eunit individually).

+1 merge

PR #40, opened 2012-02-07
Option for doing direct vnode gets

Vagabond:


jonmeredith:
Initial testing with single vnode get is good.  Need to test with it disabled and verify performance impact.

Vagabond:
I added a commit to make the keylist diff batch size configurable, and lowered the default. This fixed the 'dead socket' issues on HP cloud and shut up most of the 'busy port' messages.

jonmeredith:
+1 merge

PR #41, opened 2012-02-09
Add backpressure to keylist construction.

Vagabond:
This improves the speed of keylist generation to be at least on-par with 1.0.

### DIFF
--- a/src/riak_repl_fullsync_helper.erl
+++ b/src/riak_repl_fullsync_helper.erl
@@ -147,7 +147,7 @@ handle_call({make_keylist, Partition, Filename}, From, State) ->
     OwnerNode = riak_core_ring:index_owner(Ring, Partition),
     case lists:member(OwnerNode, riak_core_node_watcher:nodes(riak_kv)) of
         true ->
-            {ok, FP} = file:open(Filename, [read, write, binary, delayed_write]),
+            {ok, FP} = file:open(Filename, [raw, write, binary, delayed_write]),
             Self = self(),
             Worker = fun() ->
                              %% Spend as little time on the vnode as possible,
@@ -158,7 +158,7 @@ handle_call({make_keylist, Partition, Filename}, From, State) ->
                                               gen_server2:cast(MPid, {keylist, Bin}),
                                               case Count of
                                                   100 ->
-                                                      ok = gen_server:call(MPid,
+                                                      ok = gen_server2:call(MPid,
                                                           keylist_ack),
                                                       {MPid, 0};
                                                   _ ->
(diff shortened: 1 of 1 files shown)


src/riak_repl_fullsync_helper.erl
```diff
@@ -154,13 +154,21 @@ handle_call({make_keylist, Partition, Filename}, From, State) ->
...
-                             Folder = fun(K, V, MPid) -> 
+                             Folder = fun(K, V, {MPid, Count}) ->
                                               H = hash_object(V),
                                               Bin = term_to_binary({pack_key(K), H}),
-                                              file:write(FP, <<(size(Bin)):32, Bin/binary>>),
-                                              MPid
+                                              gen_server2:cast(MPid, {keylist, Bin}),
+                                              case Count of
+                                                  100 ->
+                                                      ok = gen_server:call(MPid,
```
jonmeredith: should be gen_server2:call - works fine but shouldn't break encapsulations

src/riak_repl_fullsync_helper.erl
```diff
@@ -154,13 +154,21 @@ handle_call({make_keylist, Partition, Filename}, From, State) ->
             Worker = fun() ->
                              %% Spend as little time on the vnode as possible,
                              %% accept there could be a potentially huge message queue
-                             Folder = fun(K, V, MPid) -> 
+                             Folder = fun(K, V, {MPid, Count}) ->
```
jonmeredith: Can this folder be broken out to a separate fun so we don't have as many problems with passing funs around.

src/riak_repl_fullsync_helper.erl
```diff
@@ -265,6 +275,9 @@ handle_cast({merkle, K, H}, State) ->
         false ->
             {noreply, State#state{buf = NewBuf, size = NewSize}}
     end;
+handle_cast({keylist, Row}, State) ->
+    file:write(State#state.kl_fp, <<(size(Row)):32, Row/binary>>),
```
jonmeredith: if the file:write fails - should the process continue?  I think it should crash.

src/riak_repl_fullsync_helper.erl
Vagabond: Yes, I'm going to do that once we merge Ryan's branch.

src/riak_repl_fullsync_helper.erl
```diff
@@ -149,18 +149,26 @@ handle_call({make_keylist, Partition, Filename}, From, State) ->
...
-                             Folder = fun(K, V, MPid) -> 
+                             Folder = fun(K, V, {MPid, Count}) ->
                                               H = hash_object(V),
                                               Bin = term_to_binary({pack_key(K), H}),
-                                              file:write(FP, <<(size(Bin)):32, Bin/binary>>),
-                                              MPid
+                                              gen_server2:cast(MPid, {keylist, Bin}),
+                                              case Count of
+                                                  100 ->
+                                                      ok = gen_server2:call(MPid,
```
jonmeredith: is the default 5s timeout what we want here, given we've seen long stalls from level of > 5s
how about 60s just to keep it honest?

jonmeredith:
+1 merge

PR #42, opened 2012-02-09
Move lambda into proper function to avoid distributed lambda problems

rzezeski:
This function was purposefully created because if you use a lambda
then things will go wrong when the MD5 of the module changes. I.e. if
the lambda is shipped to another node with a different version of
`riak_repl_fullsync_helper`, even if the code inside the lambda is the
same, then a `badfun` error will occur since the MD5s of the modules
are not the same.

See http://www.javalimit.com/2010/05/passing-funs-to-other-erlang-nodes.html
## How I Tested
1. 6 node devrel w/ a modified version of repl to include this patch
2. 2 clusters: dev1-dev3, dev4-dev6
3. Add listener to each node in each cluster
4. Add site to each node that points to a node in the other cluster
5. Throw 100 keys in via curl, verify some of the keys have made it to all nodes
6. Modify the `riak_repl_fullsync_helper` module so that it's MD5 changes, just added a module that is exported but never used
7. Take down the repl leader on cluster 1 (dev3 for me)
8. Move the old repl beams out of the way and move the modified ones in
9. Start dev3
10. Stop dev1 and dev2 so that dev3 becomes leader again
11. Start dev1 and dev2
12. Perform a fullsync (you have to run this on leader cause repl is lame like that)
13. Verify no errors in dev1/2/3 error.logs
14. Attach to dev1 and dev3 and verify module versions are different
    
    ```
    (dev1@127.0.0.1)1> proplists:get_value(vsn,proplists:get_value(attributes, riak_repl_fullsync_helper:module_info())).
    [293939874887387123494073304616182840441]
    
    (dev3@127.0.0.1)7> proplists:get_value(vsn,proplists:get_value(attributes, riak_repl_fullsync_helper:module_info())).
    [224311110303547758476974554848261823912]
    ```

### DIFF
--- a/src/riak_repl_fullsync_helper.erl
+++ b/src/riak_repl_fullsync_helper.erl
@@ -22,6 +22,9 @@
 -export([init/1, handle_call/3, handle_cast/2, handle_info/2,
          terminate/2, code_change/3]).
 
+%% HOFs
+-export([merkle_fold/3]).
+
 -include("riak_repl.hrl").
 -include("couch_db.hrl").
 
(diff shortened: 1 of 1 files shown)


Vagabond:
+1 for merge.

PR #43, opened 2012-03-12
Add heartbeat to repl protocol

Vagabond:
The default connection timeout on linux works like this:

If no activity after net.ipv4.tcp_keepalive_time (default 7200 seconds, or 2 hours) the tcp stack will start sending keepalives every net.ipv4.tcp_keepalive_intvl (default 75 seconds) until net.ipv4.tcp_keepalive_probe (default 9) keepalives get retransmitted.

This boils down to the fact that interrupted connection will take 2 hours to timeout and close, unless the user modifies their TCP configuration globally on the server via sysctl.

Other operating systems have similar defaults, but the configuration is different:

http://www.gnugk.org/keepalive.html

A solution to this would be to add a heartbeat message to repl, so it can notice, independently, when the connection dies. Just a simple ping-pong would be sufficient.

### DIFF
--- a/src/riak_repl_tcp_client.erl
+++ b/src/riak_repl_tcp_client.erl
@@ -36,7 +36,8 @@
         work_dir,
         fullsync_worker,
         fullsync_strategy,
-        pool_pid
+        pool_pid,
+        keepalive_time
     }).
 
 
@@ -111,7 +112,7 @@ handle_info({connected, Socket}, #state{listener={_, IPAddr, Port}} = State) ->
         my_pi=proplists:get_value(my_pi, Props),
         partitions=proplists:get_value(partitions, Props)},
     send(Socket, {peerinfo, NewState#state.my_pi,
-            [bounded_queue, {fullsync_strategies,
+            [bounded_queue, keepalive, {fullsync_strategies,
                     app_helper:get_env(riak_repl, fullsync_strategies,
                         [?LEGACY_STRATEGY])}]}),
     inet:setopts(Socket, [{active, once}]),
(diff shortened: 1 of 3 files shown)


Vagabond:
This change simply has the client and the server send keepalive messages after 60 seconds of not having received any TCP messages. The keepalives aren't tracked, they're purely to trigger the TCP retransmit timeout to close the connection. In practice, this takes between 10 and 20 minutes.

Vagabond:
Also, see http://stackoverflow.com/questions/5227520/how-many-times-will-tcp-retransmit for linux and http://docs.oracle.com/cd/E19082-01/819-2724/chapter4-32/index.html for solaris details on this. Tested on both linux and solaris.

src/riak_repl_tcp_client.erl
```diff
@@ -135,26 +136,47 @@ handle_info({tcp, Socket, Data}, State=#state{socket=Socket}) ->
...
+        Time when is_integer(Time) ->
+            {noreply, NewState, Time};
+        _ ->
+            {noreply, NewState}
     end;
 handle_info(try_connect, State) ->
     NewState = do_async_connect(State),
     {noreply, NewState};
+handle_info(timeout, State) ->
+    case State#state.keepalive_time of
```
russelldb, line 172: Just for discussion (sorry if it is a dumb question) but why not match in the function head for Time? Like

```
handle_info(timeout, #state{keepalive_time=Time, socket=Socket}=State) when is_integer(Time) ->
    send(socket, keepalive),
    {noreply, State, Time};
```

Is it just a style/preference thing?

russelldb:
Tested with a trace to see keepalive/ack messages between clusters. +1

src/riak_repl_tcp_client.erl
Vagabond, line 172: I suppose the only benefit to this is that we prevent the timeout message hitting the fallback clause.

PR #45, opened 2012-03-15
Cannot add IP/ports to existing sites

Vagabond:
```
riak-repl add site <IP> <port> <sitename>
```

does nothing if the site already exists. Sites already support multiple IP/ports internally, but you can't add them from the console.

The consequence of this is that you can't configure a site with all the potential IPs of the other cluster, just one. If that particular node is down, replication will be unable to connect at all.


russelldb:
+1, tested by adding multiple IPs for site and checking the config through riak_repl_ring:get_repl_config(Ring), saw all IP, Port pairs in list.

PR #44, opened 2012-04-03
Don't run repl client on leader node

Vagabond:
Right now, repl puts all clients and servers on the same node, this really puts a lot of strain on that node.

However, there's little reason to force the client to be on the leader node, other than to make `riak-repl stats` consistent. However, if `riak-repl status` showed the _cluster wide_ repl status, this would be a non-issue.


src/riak_repl_client_sup.erl
```diff
@@ -20,6 +21,10 @@ stop_site(SiteName) ->
 running_site_procs() ->
     [{SiteName, Pid} || {SiteName, Pid, _, _} <- supervisor:which_children(?MODULE)].
 
+%% return the node along with the running sites for accounting
+running_site_procs_rpc() ->
+    {node(), catch(running_site_procs())}.
```
rzezeski, line 26: Why the catch?

src/riak_repl_leader.erl
```diff
@@ -275,12 +316,271 @@ leader_change(A, A) ->
...
+    %% we've become the leader, stop any local clients
     RunningSiteProcs = riak_repl_client_sup:running_site_procs(),
     [riak_repl_client_sup:stop_site(SiteName) || 
-        {SiteName, _Pid} <- RunningSiteProcs],
+        {SiteName, _Pid} <- RunningSiteProcs];
+leader_change(true, false) ->
+    %% we've lost the leadership, close any local listeners
     riak_repl_listener:close_all_connections().
+
+%% here be dragons
```
rzezeski: Nice comment, any possibility to get at short description of the purpose of this fun?

src/riak_repl_leader.erl
```diff
@@ -275,12 +316,271 @@ leader_change(A, A) ->
...
+    end.
+
+%% Effectively, this function maps M configured replication clients to N
+%% nodes, where N > 1 and M >= 0. As nodes are added or removed from the
+%% cluster clients will be shuffled around to try to balance the load on any
+%% individual node.
+balance_clients(CurrentConfig, ConfiguredSites) ->
+    %% currentconfig is a list of {node, [site]} tuples and configuredsites is
+    %% merely a list of sites that should be running
+
```
rzezeski, line 531: I'm not sure `clients_per_node` follow the logic this comment implies.  E.g.

```
 clients_per_node(5, 3) => {1, 2}

 clients_per_node(5, 6) => {0, 5}
```

rzezeski:
The sites are not balancing for me.

https://github.com/basho/riak_test/compare/master...rz-repl-test

rzezeski:
While in Boston, Andrew and I noticed failures with the leader EQC tests.  I'm kicking this back to him for further investigation.

rzezeski:
For posterity:

The `repl_leader_eqc:prop_main` property will **hang** when I run it from this branch.  If I run it from my branch with the code refactoring `rz-gh44-mod-balance-clients` then it will fail but not always in the same manner.  Basically, there seems to be a race condition either in that test or in the repl code itself.

rzezeski:
+1 to merge

Vagabond:
Closing as we need to merge this into master now, not 1.1

PR #49, opened 2012-04-12
Allow buckets to participate in fullsync but not realtime replication

Vagabond:
For buckets with {repl, false}, if you want them to be fullsynced, just set {repl_fullsync, true}.


Vagabond:
Against wrong branch.

PR #50, opened 2012-04-12
Allow buckets to be fullsynced, but not participate in realtime repl

Vagabond:
This is done via new values for the 'repl' bucket property:
- {repl, both} - both types of replication, same as {repl, true}
- {repl, fullsync} - only fullsync this bucket, no realtime
- {repl, realtime} - only realtime repl this bucket, no fullsync
  
  As noted above, {repl, true} is a synonym for {repl, both} and so old
  configurations will upgrade gracefully.

### DIFF
--- a/src/riak_repl.erl
+++ b/src/riak_repl.erl
@@ -26,7 +26,7 @@ install_hook() ->
 fixup(_Bucket, BucketProps) ->
     CleanPostcommit = strip_postcommit(BucketProps),
     case proplists:get_value(repl, BucketProps) of
-        true ->
+        Val when Val==true; Val==realtime; Val==both ->
             UpdPostcommit = CleanPostcommit ++ [?REPL_HOOK],
 
             {ok, lists:keystore(postcommit, 1, BucketProps, 
--- a/src/riak_repl_util.erl
+++ b/src/riak_repl_util.erl
@@ -131,7 +131,7 @@ repl_helper_recv([{App, Mod}|T], Object) ->
 repl_helper_send(Object, C) ->
     B = riak_object:bucket(Object),
     case proplists:get_value(repl, C:get_bucket(B)) of
-        true ->
+        Val when Val==true; Val==fullsync; Val==both ->
             case application:get_env(riak_core, repl_helper) of
                 undefined -> [];
                 {ok, Mods} ->


rzezeski:
I feel like this could be confusing.  The repl property can now take on two different semantics depending on it's value.  Either it refers realtime+fullsync if true or it refers to just realtime if false.

`repl=true + repl_fullsync=true` - second property is redundant
`repl=false + repl_fullsync=true` - don't do realtime but do fullsync, sorta confusing to see repl=false but in reality it is still occurring
`repl=true + repl_fullsync=false` - fullsync is ignored in this case
`repl=false + repl_fullsync=false` - don't run repl at all

Vagabond:
I agree, do you have a better solution that is less confusing AND backwards compatible?

Vagabond:
Perhaps {repl, fullsync_only} ? But that would be a more complicated patch.

rzezeski:
I'm not sure I would keep backwards compatibility if it leads to confusion.  I wonder if anyone would ever have a reason to do realtime but not fullsync?  Perhaps a caching service that does best attempt to replicate?  I wonder if rather than a binary value it should be a set of repl types, e.g. `{repl, [fullsync, realtime]}`.  Perhaps two booleans, `repl_fullsync` and `repl_realtime` and setting `repl` to true causes both to be true?

Vagabond:
I could live with {repl, (true|realtime|fullsync|)}, and maybe add 'both' as a synonym of 'true'.

rzezeski:
I think that is less confusing and I can't think of a reason not to do it that way.

Vagabond:
Alright, will reformulate patch.

Vagabond:
Re-pushed and updated the issue description to match the new behaviour.

jonmeredith:
works nicely.

+1 merge

test helper code...

```

-module(replh).
-compile([export_all]).

test_buckets() ->
    [{<<"replfalse">>, false},
     {<<"repltrue">>, true},
     {<<"replreal">>, realtime},
     {<<"replfull">>, fullsync},
     {<<"replboth">>, both}].


set_brops() ->
    [riak_core_bucket:set_bucket(B,[{repl, Mode}]) || {B,Mode} <- test_buckets()].


write() ->
    {ok, C} = riak:local_client(),
    [C:put(riak_object:new(B, <<"key">>, atom_to_binary(Mode, utf8))) || {B, Mode} <- test_buckets()].

read() ->
    {ok, C} = riak:local_client(),
    [case C:get(B, <<"key">>) of
         {error, Reason} ->
             {Mode, Reason};
         {ok, RObj} ->
             {Mode, riak_object:get_value(RObj)}
     end || {B, Mode} <- test_buckets()].

clear() ->
    {ok, C} = riak:local_client(),
    [C:delete(B, <<"key">>) || {B, _Mode} <- test_buckets()].
```

PR #51, opened 2012-04-12
Fix repl so it can be disabled in default_bucket_props

Vagabond:
The code to delete the repl property before adding it back to the
default bucket properties was a holdover from when it was deleting the
repl postcommit hook and then re-adding it, in case it had changed.

This is actually harmful now, because it means you CANNOT disable the
realtime repl hook globally, if you wish (and enable it only on specific
buckets). The code has therefore been removed.

### DIFF
--- a/src/riak_repl.erl
+++ b/src/riak_repl.erl
@@ -16,10 +16,6 @@ stop() ->
     application:stop(riak_repl).
 
 install_hook() ->
-    {ok, DefaultBucketProps} = application:get_env(riak_core, 
-                                                   default_bucket_props),
-    application:set_env(riak_core, default_bucket_props, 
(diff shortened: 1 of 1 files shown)


src/riak_repl.erl
```diff
@@ -18,8 +18,6 @@ stop() ->
 install_hook() ->
     {ok, DefaultBucketProps} = application:get_env(riak_core,
```
jonmeredith: This code doesn't need to be here any more either - and the compiler agrees with me

```
==> riak_repl (compile)
src/riak_repl.erl:19: Warning: variable 'DefaultBucketProps' is unused
```

jonmeredith:
+1 merge after removing L19.

PR #53, opened 2012-04-13
Re-read site IPs when the list to try runs out

Vagabond:
This is in-case more have been added by the user in the meantime.


Vagabond:
Damn, wrong target branch.

PR #47, opened 2012-04-16
SSL support for repl connections

Vagabond:
Doing repl over the public internet is not a wise thing to do, unless you use a VPN or SSL tunnel it. Repl should have native SSL support to mitigate this.


Vagabond:
Notes for testing:

I've prepared some certificates here: http://hijacked.us/~andrew/certs.tgz The README in the file covers what is what.

You can also generate your own CA and keys by using this guide: http://www.debian-administration.org/articles/618 Make sure you remove the password protection from the keys you generate.

Now, to actually test it, you need to set the 3 following properties in the app.config under repl:

```
{certfile, "/home/andrew/certs/selfsigned/site1-cert.pem"},
{keyfile, "/home/andrew/certs/selfsigned/site1-key.pem"},
{cacertdir, "/home/andrew/certs/selfsigned/ca"}
```

The cacertdir is a directory containing all the CA certificates needed to verify the CA chain back to the root.

You can also verify the peer's certificate common name by doing the following:

```
{common_name_acl, ["site3.basho.com"]}
```

You can provide multiple ACLs, and you can wildcard the leftmost part of the common name, so "*.basho.com" would match "site3.basho.com" but not "foo.bar.basho.com" or "basho.com". If the ACL is specified and not the special value "*", certificates not matching any of the rules will not be allowed to connect. If no ACLs are configured, no checks on the common name are done.

Also, due to a bug in R14B04 and R15, the certificate validation doesn't actually _work_, you must use R13B03 or R15B01.

Note that you WILL be able to connect site3 to site[12] despite them not having an identical CA chain (site[12] are signed by an intermediate CA, site3 is signed by the root). You can actually disallow that behaviour by setting the {ssl_depth, ...} parameter, see http://erlang.org/pipermail/erlang-questions/2012-April/065806.html for an explanation of how it works.

Finally, this has been tested to be backwards compatible with 1.0 and1.1 nodes, they will be denied connection and the node running the SSL support will log an error. Nodes running with this patch but no SSL configured should be able to connect to older nodes as well as throw an error if they connect to a node requiring SSL. SSL is negotiated after the connection is made so that sane errors can be thrown, rather than trying to detect SSL at connection time.

bookshelfdave:
It might be nice to have {use_ssl, true} in app.config to enable/disable SSL without having to delete/comment out the whole config.

Can the timeouts be constants?

Does it matter if clusters can connect to each other with the same set of cert info?

common_name_acl seems like it should be peer_common_name_acl, otherwise it seems like you are naming this node.

@chardan @vagabond

chardan:
+1!
Nice-to-have would be improvements to error reporting when the configuration is invalid (for example, passing a file when a directory was expected).

bookshelfdave:
Connecting to a listener with invalid CA info (bad path) generates the following message repeatedly (and never stops):

(on the listening server)
16:18:32.098 [error] CRASH REPORT Process <0.5240.0> with 0 neighbours exited with reason: no match of right hand value {error,enoent} in riak_repl_util:load_certs/1 line 304 in gen_server:terminate/6 line 747
16:18:32.099 [error] Supervisor ranch_conns_sup had child ranch_conns_sup started with {ranch_conns_sup,start_protocol,undefined} at <0.5240.0> exit with reason no match of right hand value {error,enoent} in riak_repl_util:load_certs/1 line 304 in context child_terminated
16:18:32.109 [error] gen_server <0.5242.0> terminated with reason: no match of right hand value {error,enoent} in riak_repl_util:load_certs/1 line 304
16:18:32.110 [error] CRASH REPORT Process <0.5242.0> with 0 neighbours exited with reason: no match of right hand value {error,enoent} in riak_repl_util:load_certs/1 line 304 in gen_server:terminate/6 line 747
16:18:32.111 [error] Supervisor ranch_conns_sup had child ranch_conns_sup started with {ranch_conns_sup,start_protocol,undefined} at <0.5242.0> exit with reason no match of right hand value {error,enoent} in riak_repl_util:load_certs/1 line 304 in context child_terminated
16:18:32.122 [error] gen_server <0.5244.0> terminated with reason: no match of right hand value {error,enoent} in riak_repl_util:load_certs/1 line 304

%% from app.config - I appended asd to the path.
{cacertdir, "/Users/dparfitt/basho/code_review/riak_ee/certs/selfsigned/caasd"},

Do modifications to the state record present any problems?

@chardan @Vagabond Otherwise, code looks fine and works good.

Vagabond:
Please review the commit notes and see if there's anything I've forgotten to address. Also, please re verify functionality.

bookshelfdave:
@Vagabond Tested all items specified in b1e3fb0. Works great!
+1

chardan:
+1, nice work!

PR #59, opened 2012-05-24
Use folsom for repl stats

russelldb:
depends on 

https://github.com/basho/folsom/pull/4
https://github.com/basho/folsom/pull/3
https://github.com/basho/folsom/pull/2

and

https://github.com/boundary/folsom/pull/22

### DIFF
--- a/src/riak_repl_app.erl
+++ b/src/riak_repl_app.erl
@@ -28,6 +28,8 @@ start(_Type, _StartArgs) ->
     %% Spin up supervisor
     case riak_repl_sup:start_link() of
         {ok, Pid} ->
+            %% register stats
+            riak_repl_stats:register_stats(),
             ok = riak_core_ring_events:add_guarded_handler(riak_repl_ring_handler, []),
             {ok, Pid};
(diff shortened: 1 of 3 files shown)


buddhisthead:
when I add a listener to the replication master, the node crashes due to a bad function call:

2012-06-05 12:36:25.400 [error] <0.2302.0> gen_server riak_repl_leader_gs terminated with reason: bad argument in call to ets:update_counter(folsom_counters, {riak_repl,elections_leader_changed}, 1) in folsom_metrics_counter:inc/2 line 45

If I am not mistaken, there is a typo in ./deps/riak_repl/src/riak_repl_stats.erl.
 "increment_counter(elections_leader_changed)." does not agree with "{elections_leader_change, counter}," by the very tiny single character "d"...change vs changed.

I changed the code to use elections_leader_changed consistently and that resolved the crash.

russelldb:
Added a commit and some much needed eunit tests.

buddhisthead:
+1 to merge. Good fix and unit tests. Thanks!

PR #60, opened 2012-05-29
Make Repl Client NAT-aware

buddhisthead:
Some thrashing at AT&T due to Network Address Translation presence sparked some interest in adding NAT-aware commands and behavior to replication. Dave P has already added the commands to a branch, and I'm adding the client code to riak_repl_tcp_client so that it won't try and connect to itself; and will use the remote site's public NAT'd addresses if they are present.

### DIFF
--- a/src/riak_repl_console.erl
+++ b/src/riak_repl_console.erl
@@ -10,13 +10,19 @@
 
 add_listener(Params) ->
     Ring = get_ring(),
-    {ok, NewRing} = add_listener_internal(Ring,Params),
-    ok = maybe_set_ring(Ring, NewRing).
+    case add_listener_internal(Ring,Params) of
+        {ok, NewRing} ->
+            ok = maybe_set_ring(Ring, NewRing);
+        error -> error
+    end.
 
 add_nat_listener(Params) ->
     Ring = get_ring(),
-    {ok, NewRing} = add_nat_listener_internal(Ring, Params),
-    ok = maybe_set_ring(Ring, NewRing).
+    case add_nat_listener_internal(Ring, Params) of
+        {ok, NewRing} ->
+            ok = maybe_set_ring(Ring, NewRing);
+        error -> error
+    end.
 
 add_listener_internal(Ring, [NodeName, IP, Port]) ->
     Listener = make_listener(NodeName, IP, Port),
@@ -30,15 +36,15 @@ add_listener_internal(Ring, [NodeName, IP, Port]) ->
                 false ->
                     io:format("~p is not a valid IP address for ~p\n",
                               [IP, Listener#repl_listener.nodename]),
-                    {error,error};
+                    error;
                 Error ->
                     io:format("Node ~p must be available to add listener: ~p\n",
                               [Listener#repl_listener.nodename, Error]),
-                    {error,error}
+                    error
             end;
         false ->
             io:format("~p is not a member of the cluster\n", [Listener#repl_listener.nodename]),
-            {error, error}
+            error
     end.
 
 add_nat_listener_internal(Ring, [NodeName, IP, Port, PublicIP, PublicPort]) ->
@@ -51,14 +57,11 @@ add_nat_listener_internal(Ring, [NodeName, IP, Port, PublicIP, PublicPort]) ->
                     {ok, NewRing2};
                 {error, IPParseError} ->
                     io:format("Invalid NAT IP address: ~p~n", [IPParseError]),
-                    {error, error};
-                {_,Error} ->
-                    io:format("Error adding NAT Listener: ~s~n",[Error]),
-                    {error, error}
+                    error
             end;
         {error,_} ->
             io:format("Error adding nat address. ~n"),
-            {error, error}
+            error
     end.
 
 del_listener([NodeName, IP, Port]) ->
--- a/src/riak_repl_ring.erl
+++ b/src/riak_repl_ring.erl
@@ -342,4 +342,111 @@ del_natlistener_publicip_test() ->
     %% in this test
     ?assertEqual(undefined, get_listener(Ring1, {ListenAddr, ListenPort})),
     ?assertEqual(undefined, get_nat_listener(Ring1, Listener)).
+
+get_all_listeners(Ring) ->
+    RC = riak_repl_ring:get_repl_config(Ring),
+    Listeners = dict:fetch(listeners, RC),
+    NatListeners = dict:fetch(natlisteners, RC),
+    {Listeners, NatListeners}.
+
+add_del_private_and_publicip_nat_test() ->
+    Ring0 = ensure_config_test(),
+    NodeName   = "test@test",
+    ListenAddr = "127.0.0.1",
+    ListenPort = 9010,
+    NatAddr    = "10.11.12.13",
+    NatPort    = 9011,
+    NatListener = #nat_listener{nodename=NodeName,
+                                listen_addr={ListenAddr, ListenPort},
+                                nat_addr={NatAddr, NatPort}
+                               },
+    %% you can only get a nat_listener by using a repl_listener
+    Listener = #repl_listener{nodename=NodeName,
+                              listen_addr={ListenAddr, ListenPort}},
+    %% similar to riak_repl_console:add_nat_listener
+    Ring1 = add_listener(Ring0, Listener),
+    Ring2 = add_nat_listener(Ring1, NatListener),
+    Ring3 = del_listener(Ring2, Listener),
+    Ring4 = del_nat_listener(Ring3, Listener),
+    {Listeners, NatListeners} = get_all_listeners(Ring4),
+    ?assertEqual(0,length(Listeners)),
+    ?assertEqual(0,length(NatListeners)),
+    ?assertEqual(undefined, get_listener(Ring4, {ListenAddr, ListenPort})),
+    ?assertEqual(undefined, get_nat_listener(Ring4, Listener)).
+
+%% similar to test above, however, just delete the nat address
+add_del_private_and_publicip_nat2_test() ->
+    Ring0 = ensure_config_test(),
+    NodeName   = "test@test",
+    ListenAddr = "127.0.0.1",
+    ListenPort = 9010,
+    NatAddr    = "10.11.12.13",
+    NatPort    = 9011,
+    NatListener = #nat_listener{nodename=NodeName,
+                                listen_addr={ListenAddr, ListenPort},
+                                nat_addr={NatAddr, NatPort}
+                               },
+    %% you can only get a nat_listener by using a repl_listener
+    Listener = #repl_listener{nodename=NodeName,
+                              listen_addr={ListenAddr, ListenPort}},
+    %% similar to riak_repl_console:add_nat_listener
+    Ring1 = add_listener(Ring0, Listener),
+    Ring2 = add_nat_listener(Ring1, NatListener),
+    Ring3 = del_listener(Ring2, Listener),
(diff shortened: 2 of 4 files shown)


src/riak_repl_tcp_server.erl
```diff
@@ -367,3 +395,55 @@ send_diffobj(Msg,State=#state{socket=Socket,pending=Pending}) ->
...
+    ListenAddr = "127.0.0.1",
+    ListenPort = 9010,
+    Listener = #repl_listener{nodename=NodeName,
+                              listen_addr={ListenAddr, ListenPort}},
+    Ring1 = riak_repl_ring:add_listener(Ring0, Listener),
+    {Ip, Port} = ip_and_port_for_node(NodeName, Ring1, ListenAddr),
+    ?assertEqual("127.0.0.1", Ip),
+    ?assertEqual(9010, Port).
+
+skip_nap_test() ->
```
Vagabond: I think this is a typo.

src/riak_repl_tcp_client.erl
```diff
@@ -125,16 +127,20 @@ handle_info({connected, Socket}, #state{listener={_, IPAddr, Port}} = State) ->
...
         my_pi=proplists:get_value(my_pi, Props),
         partitions=proplists:get_value(partitions, Props)},
     send(Socket, {peerinfo, NewState#state.my_pi,
-            [bounded_queue, keepalive, {fullsync_strategies,
+                  [bounded_queue, keepalive,
+                   {fullsync_strategies,
                     app_helper:get_env(riak_repl, fullsync_strategies,
-                        [?LEGACY_STRATEGY])}]}),
+                                       [?LEGACY_STRATEGY])},
+                   {connected_ip, TheirIP}
```
Vagabond: TheirIP which comes from inet:peername, returns an IP tuple like {127, 0, 0, 1} when we actually want a string. I was wrong to suggest using it instead of IPAddr.

Vagabond:
I can confirm that, with the fix to connected_ip mentioned above, I can both connect to an external address and be redirected to the correct external address and do the same test with internal addresses.

I can't test both scenarios in the same cluster due to limitations on virtualbox networking, but I have tested them independently.

src/riak_repl_console.erl
```diff
@@ -17,22 +26,46 @@ add_listener([NodeName, IP, Port]) ->
...
+        {ok,NewRing} ->
+            case inet_parse:address(PublicIP) of
+                {ok, _} ->
+                    NatListener = make_nat_listener(NodeName, IP, Port, PublicIP, PublicPort),
+                    NewRing2 = riak_repl_ring:add_nat_listener(NewRing, NatListener),
+                    {ok, NewRing2};
+                {error, IPParseError} ->
+                    io:format("Invalid NAT IP address: ~p~n", [IPParseError]),
+                    {error, error};
+                {_,Error} ->
```
Vagabond: Exactly what error are you expecting here, and what does the tuple start with if not {error, ...}?

src/riak_repl_console.erl
```diff
@@ -17,22 +26,46 @@ add_listener([NodeName, IP, Port]) ->
                                 riak_repl_util, valid_host_ip, [IP]) of
                 true ->
                     NewRing = riak_repl_ring:add_listener(Ring, Listener),
-                    ok = maybe_set_ring(Ring, NewRing);
+                    {ok,NewRing};
                 false ->
                     io:format("~p is not a valid IP address for ~p\n",
-                              [IP, Listener#repl_listener.nodename]);
+                              [IP, Listener#repl_listener.nodename]),
+                    {error,error};
```
Vagabond: The error here could be more useful.

src/riak_repl_console.erl
```diff
@@ -17,22 +26,46 @@ add_listener([NodeName, IP, Port]) ->
...
                 Error ->
                     io:format("Node ~p must be available to add listener: ~p\n",
-                              [Listener#repl_listener.nodename, Error])
+                              [Listener#repl_listener.nodename, Error]),
+                    {error,error}
```
Vagabond: Also here.

src/riak_repl_console.erl
```diff
@@ -17,22 +26,46 @@ add_listener([NodeName, IP, Port]) ->
...
+                    NewRing2 = riak_repl_ring:add_nat_listener(NewRing, NatListener),
+                    {ok, NewRing2};
+                {error, IPParseError} ->
+                    io:format("Invalid NAT IP address: ~p~n", [IPParseError]),
+                    {error, error};
+                {_,Error} ->
+                    io:format("Error adding NAT Listener: ~s~n",[Error]),
+                    {error, error}
+            end;
+        {error,_} ->
```
Vagabond, line 62: Why not log the error reason, assuming you make it something useful.

src/riak_repl_ring.erl
```diff
@@ -135,25 +138,64 @@ add_listener(Ring,Listener) ->
...
     end.
 
+-spec(add_nat_listener/2 :: (ring(), #nat_listener{}) -> ring()).
+%% @doc Add a replication NAT listener host/port to the Ring.
+add_nat_listener(Ring,NatListener) ->
+    RC = get_repl_config(Ring),
+    case dict:find(natlisteners,RC) of
+        {ok, NatListeners} ->
+            case lists:member(NatListener, NatListeners) of
+            false ->
```
Vagabond: Indent seems goofed here.

src/riak_repl_ring.erl
```diff
@@ -135,25 +138,64 @@ add_listener(Ring,Listener) ->
...
+                riak_core_ring:update_meta(
+                  ?MODULE,
+                  dict:store(natlisteners, NewListeners, RC),
+                  Ring);
+            true ->
+                Ring
+            end;
+        error ->
+            %% there are no natlisteners entries yet.
+                NewListeners = [NatListener],
```
Vagabond: More weird indenting.

src/riak_repl_ring.erl
```diff
@@ -135,25 +138,64 @@ add_listener(Ring,Listener) ->
...
+    end.
+
+-spec(del_nat_listener/2 :: (ring(),#repl_listener{}) -> ring()).
+%% @doc Delete a nat_listener from the list of nat_listeners
+del_nat_listener(Ring,Listener) ->
+    RC  = get_repl_config(Ring),
+    case get_nat_listener(Ring, Listener) of
+        undefined ->  Ring;
+        NatListener ->
+            NatListeners  = dict:fetch(natlisteners, RC),
```
Vagabond: You carefully used dict:find before in add_nat_listener to handle when the key wasn't in the dict (legacy repl config) but now you use fetch(), which will throw an error?

src/riak_repl_ring.erl
```diff
@@ -161,9 +203,29 @@ get_listener(Ring,{_IP,_Port}=ListenAddr) ->
         {value,Listener} -> Listener
     end.
 
+-spec(get_nat_listener/2 :: (ring(), #repl_listener{}) -> #nat_listener{}|undefined).
+%% @doc Fetch a replication nat host/port listener record from the Ring.
+get_nat_listener(Ring,Listener) ->
+    RC = get_repl_config(Ring),
+    NatListeners  = dict:fetch(natlisteners, RC),
```
Vagabond: dict:fetch() again. Maybe you need a wrapper function for getting the natlisteners safely.

src/riak_repl_ring.erl
```diff
@@ -206,19 +268,45 @@ del_site_test() ->
...
+                              listen_addr={ListenAddr, ListenPort},
+                              nat_addr={NatAddr, NatPort}
+                               },
+    Ring1 = add_nat_listener(Ring0, NatListener),
+    %% you can only get a nat_listener by using a repl_listener
+    Listener = #repl_listener{nodename=NodeName,
+                              listen_addr={ListenAddr, ListenPort}},
+    ?assertNot(undefined == get_nat_listener(Ring1, Listener)).
+
+del_natlistener_test() ->
```
Vagabond, line 317: There's no test for deleting a nat listener using the external ip/port combo. Interestingly, what that does is it removes only the nat listener, but not the underlying local listener.

src/riak_repl_tcp_client.erl
```diff
@@ -23,6 +23,8 @@
 -export([init/1, handle_call/3, handle_cast/2, handle_info/2,
          terminate/2, code_change/3]).
 
+-include_lib("eunit/include/eunit.hrl").
```
Vagabond, line 27: This needs to be inside an  -ifdef(TEST) block.

src/riak_repl_tcp_client.erl
```diff
@@ -401,35 +408,340 @@ handle_peerinfo(#state{sitename=SiteName, socket=Socket} = State, TheirPeerInfo,
...
+    NatListenAddrs = [R#nat_listener.listen_addr || R <- NatListeners],
+    ListenAddrs++NatAddrs++NatListenAddrs.
+
+-spec(rewrite_config_site_ips_pure/4 :: (repl_config(),ring(),repl_sitename(),ip_addr_str())
+                                        -> none|ring()).
+%% @doc Update replication configuration with corrected set of IP addrs for a RemoteSite.
+%%
+%% Given a "remote" server's replication configuration and our own ring configuration,
+%% update the list of IP addresses for the remote server. This function will ensure
+%% that we don't add our own lisenter IP addresses to that list, even if we are setup
```
Vagabond: Typo: lisenter

src/riak_repl_tcp_client.erl
```diff
@@ -401,35 +408,340 @@ handle_peerinfo(#state{sitename=SiteName, socket=Socket} = State, TheirPeerInfo,
...
+%%
+%% This also ensures that we remove our own IP addrs from the list, just in case they
+%% leaked into the remote site's list of listen IP addresses. Not likely, but game over
+%% if it happens.
+update_site_ips(TheirReplConfig, SiteName, ConnectedIP) ->
+    {ok, OurRing} = riak_core_ring_manager:get_my_ring(),
+    NeededConfigChanges = rewrite_config_site_ips_pure(TheirReplConfig, OurRing,
+                                                       SiteName, ConnectedIP),
+    case NeededConfigChanges of
+	none ->
```
Vagabond: Bad indenting. Use spaces, not tabs.

src/riak_repl_tcp_server.erl
```diff
@@ -28,6 +28,8 @@
 -export([start_fullsync/1, cancel_fullsync/1, pause_fullsync/1,
         resume_fullsync/1, handle_peerinfo/3, make_state/5]).
 
+-include_lib("eunit/include/eunit.hrl").
```
Vagabond, line 32: Ifdef this as mentioned above.

src/riak_repl_tcp_server.erl
bookshelfdave: An awesome typo though.

src/riak_repl_ring.erl
bookshelfdave, line 317: riak_repl_console:del_listener ties both riak_repl_ring:del_listener and riak_repl_ring:del_nat_listener together. 

I thought that code inside of riak_repl_ring should be more orthogonal than the console code.

Vagabond:
I am +1 on this. @buddhisthead is testing the NAT support as well so we're going to wait to merge until he confirms my test results.

buddhisthead:
+1 to merge. All my testing is complete. There is a problem with stats on the client and server not matching, but this must be a long-standing bug because none of the new code touches that.

Issue #65, opened 2012-06-23
Server Stats and Start/Stop/Pause FullSync Do Not Work


ghost:
Previously, `riak_repl_server_sup` handled `riak_repl_tcp_server`. When the fullsync and stats commands were run from `riak_repl_console`, the following was ran:

``` erlang
server_pids() ->
    [P || {_,P,_,_} <- supervisor:which_children(riak_repl_server_sup), P /= undefined].
```

This listed the `riak_repl_tcp_server` children, and then the commands were sent like so:

``` erlang
start_fullsync([]) ->
    [riak_repl_tcp_server:start_fullsync(Pid) || Pid <- server_pids()],
    ok.
```

However, this has changed. Instead of calling `riak_repl_server_sup:start_server/1`, it is called via `riak_repl_listener_sup:ensure_listeners/1`, which calls `riak_repl_listener_sup:start_listener/1`:

``` erlang
start_listener(Listener = #repl_listener{listen_addr={IP, Port}}) ->
    case riak_repl_util:valid_host_ip(IP) of
        true ->
            lager:info("Starting replication listener on ~s:~p",
                [IP, Port]),
            {ok, RawAddress} = inet_parse:address(IP),
            ranch:start_listener(Listener, 10, ranch_tcp,
                [{ip, RawAddress}, {port, Port}], riak_repl_tcp_server, []);
        _ ->
            lager:error("Cannot start replication listener "
                "on ~s:~p - invalid address.",
                [IP, Port])
    end.
```

You'll note that it called `ranch:start_listener/6`. This is a new dependency, managing socket connection pools. The source is located at https://github.com/extend/ranch. 

With this, calling `supervisor:which_children(riak_repl_server_sup).` results in nothing:

``` erlang
[]
```

This is because children are no longer spawned under `riak_repl_server_sup`. 

However, unlike listing supervisors under `riak_repl_server_sup`, which had `riak_repl_tcp_server` as a direct child, `ranch` operates differently.

Note the following:

``` erlang
supervisor:which_children(ranch_sup).               
[{{ranch_listener_sup,#repl_listener{nodename = 'dev1@127.0.0.1',
                                     listen_addr = {"127.0.0.1",9010}}},
  <0.31129.0>,supervisor,
  [ranch_listener_sup]}]
```

This has a `ranch_listener_sup` as a child. Digging further in:

``` erlang
supervisor:which_children(c:pid(0,31129,0)).
[...]

PR #67, opened 2012-06-25
Make ranch processes available to REPL.

bookshelfdave:
Fix for issue #65

Thanks to Dan Reverri for the list comprehension.

### DIFF
--- a/src/riak_repl_console.erl
+++ b/src/riak_repl_console.erl
@@ -266,4 +266,7 @@ server_stats(Pid) ->
     {Pid, erlang:process_info(Pid, message_queue_len), State}.
 
 server_pids() ->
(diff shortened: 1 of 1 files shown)


beerriot:
Is there any way that we could push this OTP tree walking down into a ranch branch? It seems a bit fragile to me, but announcing that we care about such a feature by exposing it in an API on a branch of their repo would improve the situation.

bookshelfdave:
@beerriot My original intent for this PR was to fix REPL as quickly as possible. I agree that these changes make more sense in Ranch. Andrew mentioned that this is a feature the Ranch dev's have already considered. I'll chat with them on IRC and see what the status is. Thanks for the input!

Vagabond:
+1 on this, if you fix the indenting and add a comment about why this is needed and to check for updates upstream.

PR #63, opened 2012-06-20
Cast writes to folsom, cache reads via riak core stat cache

russelldb:
Testing showed that concurrent reads and writes to folsom slowed through put enough to make async calls better. Ideally we'd have a process per-stat, and I'll introduce that next, but for expediency's sake have gone with back to gen_server cast.

Also, use the riak core stat cache for reads.

PR #68, opened 2012-06-26
Start riak_core_stat_cache in eunit tests

beerriot:
This is an addition to #63 to fix the tests complaining about `riak_core_stat_cache` not being started.


russelldb:
+1 thanks for that.

PR #70, opened 2012-06-28
Don't crash if the peerinfo from a 1.1 server doesn't include natlisters

Vagabond:
@metadave Please review and check that I didn't miss any other places you did this unsafely. get_all_listener_addrs should be safe since it is only called on the local repl config. Although this may be unsafe if the ring is one upgraded from 1.1.


bookshelfdave:
@Vagabond fixed. 
+1

PR #71, opened 2012-06-28
Fix repl tx/rx stats

russelldb:
Flatten the history list
Update the last_report time
Add unit test for report


Vagabond:
+1 for merge, and thanks for fixing this.

PR #72, opened 2012-07-06
Make client balancing work right in cluster with legacy nodes

Vagabond:


jtuple:
Still working on long-term automated test to verify repl, but I was able to determine that this change fixes the error I observed yesterday.

Specifically, on master you run into this error with a mixed-version cluster:

```
2012-07-06 14:08:58.521 [error] <0.1600.0> Supervisor riak_repl_sup had child riak_repl_leader started with riak_repl_leader:start_link() at <0.2428.0> exit with reason bad arithmetic expression in riak_repl_leader:clients_per_node/2 in context child_terminated
2012-07-06 14:08:58.523 [error] <0.2431.0> gen_server riak_repl_leader_gs terminated with reason: bad arithmetic expression in riak_repl_leader:clients_per_node/2
```

With this branch, the error is avoided. Also, repl appears to still work fine -- with repl between A->B, I verified writes to A showed up on B.

Looks good for now. The more involved basho_expect tests will exercise everything, and hopefully pass this time. +1 merge.

PR #73, opened 2012-07-11
Don't put {badrpc,nodedown} in status when there's no leader

Vagabond:


jonmeredith:
+1 merge, resolves issue for me.

PR #74, opened 2012-07-19
Register repl stat mod with riak_core on app start up

russelldb:
If the cache is restarted, automatically register the riak_repl stats mod.


jonmeredith:
+1 merge

PR #76, opened 2012-07-24
Set stage_start when leaving wait_for_partition

Vagabond:
Not setting stage_start means 'riak-repl status' can crash the fullsync
if called before the keylist helper is started.


bookshelfdave:
@Vagabond - this patch seems to work fine with EE master, no errors.
+1

PR #77, opened 2012-07-24
Redirect 1.0 and earlier nodes to the leader correctly

Vagabond:


jonmeredith:
+1 merge.

Resolves the problem for me.  Replicated by hand, two 1.2 nodes as server, single 1.0.3 node as client.

PR #78, opened 2012-07-25
Place riak_repl_stats under supervision of riak_core stat subsytem

russelldb:


russelldb:
Should have said, depends on https://github.com/basho/riak_core/pull/217

argv0:
+1

PR #80, opened 2012-07-31
Configurable timeout for "riak-repl status" statistics

Vagabond:
Right now, the stats query done by "riak-repl status" has a hardcoded timeout of 1000 milliseconds, which is too short if the server is under heavy load.  Make this timeout configurable, or hardcode 'infinity' instead.


slfritchie:
+1 to merge, many thanks!

jonmeredith:
Worked for me too.

PR #82, opened 2012-08-01
Realtime discards with large mailboxes

Vagabond:
There are 3 mailboxes that can grow large when realtime load is high and replication throughput is slower:
- Local gen_leader candidate mailbox (from postcommit hook)
- Elected gen_leader candidate (from gen_leader candidate)
- riak_repl_tcp_server (from elected gen_leader)

The strategy used to mitigate this is to inspect the mailbox size of the process we're sending the message to and to drop the message if the destination mailbox is over a certain size (20 kilomessages). Because polling this on every request involves getting a lock, which can cause the process asking for the data getting descheduled (when the target is running), we try to poll intermittently to avoid that. In addition, the cross-node call for the candidate->leader has 2 thresholds and will drop a % of messages in order to try to keep the mailbox below the limit, rather than switch purely between drop and send.

### DIFF
--- a/src/riak_repl_leader.erl
+++ b/src/riak_repl_leader.erl
@@ -188,6 +188,8 @@ handle_cast({set_candidates, CandidatesIn, WorkersIn}, State) ->
             {noreply, restart_helper(UpdState2)}
     end;
 handle_cast({repl, Msg}, State) when State#state.i_am_leader =:= true ->
+    %% To simulate a slow elected repl leader, uncomment and/or change amount
+    %% timer:sleep(1),
     case State#state.receivers of
         [] ->
             riak_repl_stats:objects_dropped_no_clients();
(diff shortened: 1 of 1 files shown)


buddhisthead:
I am +1 on this.

One question... does doing an rpc:call() across nodes every second seem like an issue? What is the rate of mailbox growth under load. Is one second fast enough? I guess it's really a Nyquist question. If it doesn't do any harm to occasionally exceed the 20K limit, then I guess we're ok. We should ideally sample at 2X the frequency of "how soon can we blow past the upper bound".

I tested without the fix (with timer:sleep(1) in place) to verify that mailboxes where growing like crazy. Yep.
Was unable to reproduce the problem by slowing down the network link; I added 600 msec of latency and it still didn't cause backup in the mailboxes, so I don't understand that. Ring size was 128. 20 workers in basho_bench.

Tested with the fix (with timer:sleep(1) in place) and verified the pain went away. The mailbox stayed pegged pretty much right at 19999 or 20000 once things got hot (with basho_bench running as above). Throughput was down to 5 Mbs, which was as expected.

Removed timer:sleep(1) and verified that the new code did no harm. Throughput remained at about 32 Mbs, compared to same test scenario on branch master, which also ran at about 32 Mbs.

PR #85, opened 2012-08-02
Expose repl stats over HTTP

jaredmorrow:


argv0:
making a new PR.  We don't need bolth.

PR #87, opened 2012-08-02
Add max queue size and percentage used to repl status

Vagabond:
This should wait to be merged after 1.2.0 is released.


Vagabond:
Obsolete.

PR #89, opened 2012-08-02
Cet fullsync latency

buddhisthead:
Full-sync replication, over a link with high latency, can perform sub-optimally due to large gaps in the flow of diff-objects from server to client. This was due to the server waiting for a handshake to return from the client before sending another batch of diffs.

This PR maintains the same protocol from the client's perspective, but keeps two half-sized batches in flight to maintain a full pipeline of diff-objects on the wire with no gaps. Gaps could occur if the client is unable to keep up and eventually can not acknowledge either of the two batches in flight from the server.

Compared to the previous implementation, this change was shown to improve performance over a link with 60 msec of latency from 39 Mb/sec to 50 Mb/sec, which equalled the maximum rate when there was zero latency. So, the result seems to show that we can achieve the same throughput as zero latency, even when the link has significant latency.

### DIFF
--- a/src/riak_repl_fullsync_helper.erl
+++ b/src/riak_repl_fullsync_helper.erl
@@ -376,16 +376,16 @@ itr_next(Size, File, Tag) ->
 
 diff_keys(R, L, #diff_state{replies=0, fsm=FSM, ref=Ref, count=Count} = DiffState) ->
     gen_fsm:send_event(FSM, {Ref, diff_paused}),
-    lager:info("waiting for stop or resume"),
+%%    lager:info("waiting for stop or resume"),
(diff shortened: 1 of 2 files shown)


buddhisthead:
I don't understand why this can't be merged automatically. I've rebased against master.

buddhisthead:
Ok, merge completed and we have a shinny green button now :-) I re-ran my tests so this should be g2g. Thanks @rzezeski , @metadave and @Vagabond for your help in understanding the rebase workflow.

Also, the credit for the performance measurements go to @Vagabond as well.

Just don't ask me to squash the commits :-) I know it's messy but I don't want to change this again.

PR #91, opened 2012-08-02
Fix fullsync crashes on stale message and delayed_write errors

Vagabond:
The client can get a stale kl_exchange message if it told the server to
skip a partition but the server had already built the keylist.

Also, in some cases, the sync() done after writing the keylist can fail,
because of delayed_write silliness, and that will cause the filesorter
to crash with file_error errors. Double sync/closing the file seems to
prevent this from happening.


jonmeredith:
Code looks good, trying to replicate for a little longer.

jonmeredith:
+1 merge.

PR #94, opened 2012-08-08
added snappy/gzip compression

bookshelfdave:
configure by adding:
{compression, snappy}
or
{compression, [snappy,gzip]}
to riak_repl section of app.config

NOTE: 
To build, you'll need to check these branches out:
https://github.com/basho/erlang_js/pull/29
https://github.com/basho/riak_ee/pull/85

cc @Vagabond


src/riak_repl_syncv1_server.erl
```diff
@@ -27,15 +27,15 @@
 -include_lib("kernel/include/file.hrl").
```
Vagabond: The syncv1 files are actually deprecated and we can remove them now instead of updating them.

src/riak_repl_tcp_client.erl
```diff
@@ -296,12 +304,24 @@ async_connect(Parent, IPAddr, Port) ->
...
     end.
 
-send(Transport, Socket, Data) when is_binary(Data) ->
+send(Transport, Socket, Data, undefined) when is_binary(Data) ->
+    R = Transport:send(Socket, Data),
+    riak_repl_stats:client_bytes_sent(size(Data)),
+    R;
+send(Transport, Socket, Data0, CompType) when is_binary(Data0) ->
+    lager:debug("Client send compressed"),
+    Data = riak_repl_util:compress(CompType, Data0),
```
Vagabond: Given that we ALWAYS compress after something is converted to a binary, it might be nice to have a separate stat for client_bytes_sent and client_bytes_sent_uncompressed, so you can compare them. Ditto for the server side, obviously.

src/riak_repl_util.erl
```diff
@@ -567,3 +572,31 @@ parse_vsn(Str) ->
...
+    end.
+
+compress(?GZIP_COMPRESSION, Data) ->
+    CompressedBin = zlib:gzip(Data),
+    << ?GZIP_COMPRESSION_HEADER, CompressedBin/binary>>;
+compress(?SNAPPY_COMPRESSION, Data) ->
+    {ok, CompressedBin} = snappy:compress(Data),
+    << ?SNAPPY_COMPRESSION_HEADER, CompressedBin/binary >>;
+compress(UnknownCompScheme, Data) ->
+    %% don't blow up, just pass the data on uncompressed
```
Vagabond, line 601: Y U NO blow up?

src/riak_repl_tcp_client.erl
```diff
@@ -142,16 +144,19 @@ handle_info({connected, Transport, Socket}, #state{listener={_, IPAddr, Port}} =
...
             %% we don't renegotiate SSL. This avoids leaking the ring and
             %% accidentally connecting to an insecure site
+            %% Don't send this data compressed. Compression hasn't been negotiated.
+
             send(Transport, Socket, {peerinfo,
-                    riak_repl_util:make_fake_peer_info(),
+                                     riak_repl_util:make_fake_peer_info(),
                                      [ssl_required,
-                                      {connected_ip, IPAddr}]})
+                                      {connected_ip, IPAddr}]}, [])
```
Vagabond: My mistake, passing Compression here is fine because it will be 'undefined' at this point and simply sent uncompressed. Sending [] makes warning messages appear.

src/riak_repl_tcp_server.erl
```diff
@@ -347,21 +366,24 @@ send_peerinfo(#state{transport=Transport, socket=Socket, sitename=SiteName} = St
...
                             {noreply, State#state{
                                     client=proplists:get_value(client, Props),
                                     election_timeout=undefined,
                                     my_pi=PI}};
                         {Config, tcp} ->
+                            %% Don't send this data compressed. Compression hasn't been negotiated.
                             send(Transport, Socket, {peerinfo,
                                     riak_repl_util:make_fake_peer_info(),
-                                    [ssl_required]}),
+                                    [ssl_required]}, []),
```
Vagabond: Another place you should just be sending Compression.

src/riak_repl_fullsync_worker.erl
```diff
@@ -10,6 +10,7 @@
 -export([start_link/1, do_put/3, do_get/6, do_get/7]).
 
 -record(state, {
+          compression
     }).
 
 start_link(_Args) ->
```
Vagabond: Args is ignored, and not passed through to start_link so 'Args' in init() is always [], so compression never works in the fullsync_worker, which means all the fullsync GETs are sent uncompressed.

src/riak_repl_fullsync_worker.erl
```diff
@@ -25,8 +26,10 @@ do_get(Pid, Bucket, Key, Transport, Socket, Pool, Partition) ->
     gen_server:call(Pid, {get, Bucket, Key, Transport, Socket, Pool, Partition}, infinity).
 
 
-init([]) ->
-    {ok, #state{}}.
+init(Args) ->
+    WorkerArgs = proplists:get_value(worker_args,Args,[]),
+    Compression = proplists:get_value(compression, WorkerArgs, []),
```
Vagabond: The default should be 'undefined' here, or just use get_value/2 which will return undefined if the value is missing.

src/riak_repl_tcp_server.erl
```diff
@@ -230,11 +233,18 @@ handle_msg({peerinfo, PI}, State) ->
     handle_msg({peerinfo, PI, Capability}, State);
 handle_msg({peerinfo, TheirPI, Capability}, State) ->
     handle_peerinfo(State, TheirPI, Capability);
+<<<<<<< HEAD
```
Vagabond, line 236: Unresolved conflict!

PR #97, opened 2012-08-15
Seperate the bounded_queue out into a seperate process

Vagabond:
Also, improves the queue info in 'riak-repl status'.


jonmeredith:
+1 merge - thanks for fixing reversed/leader receiver issues.

PR #99, opened 2012-08-16
Use a lock-free queue and try to size packets better

Vagabond:
DO NOT MERGE - this change does not work with rolling upgrades!

The rationale behind these changes is to try to size the packets on the wire close to the MTU (or above, for larger objects). Also, the queue is not drained when things are enqueued, only when the client requests us to do so (either by hitting the ack_freq or by a timer that fires if the ack_freq hasn't been hit recently).

Additionally, the bounded_queue has been replaced by erl_lfq, a lock-free single-producer single-consumer queue where the producer is the queue process and the consumer is the TCP server process. This avoid the problem of the message to flush the queue being buried in the mailbox of the queue itself, behind lots of elements to be enqueued.

### DIFF
--- a/ebin/riak_repl.app
+++ b/ebin/riak_repl.app
@@ -42,6 +42,7 @@
                   ssl,
                   riak_core,
                   riak_kv,
+                  erl_lfq,
                   ranch]},
   {registered,   [riak_repl_connector_sup,
                   riak_repl_leader,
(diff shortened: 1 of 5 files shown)


slfritchie:
Hrm, there's a part of me that would really love to see what happens without resorting to a new NIF.  I haven't had any time  to work on the `slf-multiregproc-leader` branch, though ... I have DTrace stuff for R15B02 to shepherd through the OTP team first.  Any chance that you or @jonmeredith can explore?

Vagabond:
Interesting experiment, but made redundant by the BNW stuff, closing.

PR #101, opened 2012-08-17
Cet bloom latency merge

buddhisthead:
This PR merges or completes the following features:
- Bloom fold
- reduced latency on full-sync using two windows of ACKs with the client
- uses the old keylist_folder on a cluster of mixed capabilities
- re-introduces client throttling via ACKs
- documents the message sequences for normal full-sync operation

Measured performance gains are 200+ % improvement in the time to exchange differences (zero latency link) using Bloom fold over the key-list method (1M keys, 100% differences, i.g. client was blank). There is a 70% reduction in exchange time performance when there are no differences between server and client. This is because the Bloom fold is folding over all keys again. However, the total time in that case is still dominated by the time to build the actual key list, so although 70% sound bad, it's the smallest component in the total time. An optimization switches to the old key-list method if the cluster has not been fully upgraded.

Total full-sync time was reduced by 97% for the blank client case and only worsened by 23% in the "no differences" case. This is a good tradeoff since the "no differences" case completes very quickly anyhow.

Performance improvements for the low-latency, multiple ACK window method, gain 20% when observed over a link with 60 msec latency (30 msec in each direction). This helps when MDCs are connected by crap network links, like we have observed in some very large and should-do-better telecomm companies.

Implementation wise, you will observe that I have cloned the diff_keylist sates to diff_bloom states. This was a good choice, I believe, because the bloom states perform push-back (ACKs) with the client at a different stage than the diff_keylist states and thus preserves the old code path as purely as possible.

Finally, to see the protocol differences between the old way (diff_keylist) and then the new way (diff_bloom), please look at the .png files in ./doc, which are Message Sequence Charts generated using Plantuml source repl.plu.

Thanks for reading and reviewing.
Cheers, Chris


src/riak_repl_keylist_server.erl
```diff
@@ -216,20 +276,42 @@ wait_keylist(kl_eof, #state{their_kl_fh=FH} = State) ->
...
-        State#state.kl_fn, State#state.their_kl_fn,
-        State#state.diff_batch_size),
-    {next_state, diff_keylist, State#state{diff_ref=Ref, diff_pid=Pid,
-            stage_start=now()}};
+                                                      State#state.kl_fn,
+                                                      State#state.their_kl_fn,
+                                                      DiffSize),
+    %% Use bloom fold unless the number of keys is small, which would just be slower
+    %% because we'd have to fold over a giant key space for a small number of things.
+    NextState = case NumKeys > ?KEY_LIST_THRESHOLD of
```
Vagabond: As discussed before, NumKeys is the size of the local keylist, not the number of differences. IIRC we decided to just always bloom fold and take the minor hit on the 0% differences case.

buddhisthead: I disagree. I have changed the variable to show that I know that it represents the number of keys. When the number of keys is small, then so will be the number of diffs. This is a crude optimization until we can measure the number of differences. I would prefer to keep the switch in place, not to always do the fold.

Vagabond: Then maybe it makes sense to also check the capability for bloom_fold there rather than when we start fullsync so we can still do fullsyncs in a mixed cluster by forcing the older keylist diff mode.

buddhisthead: Ah, I see what you mean. So, you think maybe it's better to go slow rather than not at all, in the case of mixed clusters? You have a good point. But I'm reluctant to mess with that code path at this point because of our other pending work. It will take the rest of the day to make and test those changes.

Vagabond: True. I may take a stab at it myself and let you take a look rather than derail you too much.

src/riak_repl_fullsync_helper.erl
```diff
@@ -468,14 +473,16 @@ merkle_fold(K, V, Pid) ->
 %% @private
 %%
 %% @doc Visting function for building keylists. Similar to merkle_fold.
-keylist_fold(K, V, {MPid, Count}) ->
```
Vagabond, line 471: Changing the signature of the accumulator value breaks when the leader is running old code but the node doing the fold isn't. We should add a second clause for this function that uses the old 2-tuple accumulator for that case.

src/riak_repl_keylist_server.erl
```diff
@@ -250,26 +333,174 @@ diff_keylist({Ref, {merkle_diff, {{B, K}, _VClock}}}, #state{
...
+diff_bloom({Ref, diff_done}, #state{diff_ref=Ref, partition=Partition, bloom=Bloom} = State) ->
+    lager:info("Full-sync with site ~p; fullsync difference generator for ~p complete (completed in ~p secs)",
+               [State#state.sitename, State#state.partition,
+                riak_repl_util:elapsed_secs(State#state.partition_start)]),
+    {ok, Ring} = riak_core_ring_manager:get_my_ring(),
+    OwnerNode = riak_core_ring:index_owner(Ring, Partition),
+
+    Self = self(),
+    DiffSize = State#state.diff_batch_size,
+    Worker = fun() ->
```
Vagabond: If ebloom:elements() is 0 here, we can skip the fold completely and save like 30 seconds (for a 1 million key vnode).

Vagabond:
I made the discussed changed, plus a couple more I found along the way. I'm +1 on this on my end, but you should review my changes. Also, please don't merge this to master, we're going to merge to 1.2 instead.

Vagabond:
Also, the cet-bloom-latency-1.2 branch, is this same branch rebased off of 1.2.

buddhisthead:
+1 to Andrew's latest changes. I am not merging to master. @Vagabond  the merge is in your hands.

Vagabond:
Manually merged.

Issue #69, opened 2012-06-26
Clients do not handle gen_tcp:send/2 errors


dreverri:
The `riak_repl_tcp_client` does not handle the return from `gen_tcp:send/2` which may return `{error, Reason}`. This can cause replication clients to hang around indefinitely without replication actually occurring.

A module has been provided below to help observe current client behavior. To observe the issue you will need to setup 2 Riak nodes and connect them via riak-repl. Once that is setup, you can use the below module as follows:

``` erlang
riak_repl_client_test:test_client_ports().                                 
[{<0.8179.0>,ok},{<0.8182.0>,ok}]

riak_repl_client_test:close_client_ports().                                  
[{<0.8179.0>,true},{<0.8182.0>,true}]

riak_repl_client_test:test_client_ports(). 
[{<0.8179.0>,{error,closed}},{<0.8182.0>,{error,closed}}]

riak_repl_client_test:test_client_ports_through_client().
[{<0.8179.0>,
  {status,[{node,'node1@127.0.0.1'},
           {site,"node2-to-node1"},
           {strategy,riak_repl_keylist_client},
           {fullsync_worker,<0.8184.0>},
           {put_pool_size,5},
           {connected,"127.0.0.1",9012},
           {state,wait_for_fullsync}]}},
 {<0.8182.0>,
  {status,[{node,'node1@127.0.0.1'},
           {site,"node3-to-node1"},
           {strategy,riak_repl_keylist_client},
           {fullsync_worker,<0.8186.0>},
           {put_pool_size,5},
           {connected,"127.0.0.1",9013},
           {state,wait_for_fullsync}]}}]
```

The last command, `riak_repl_client_test:test_client_ports_through_client/0`, forces the client to send a `keapalive_ack` over the closed port. The expectation is that this would cause the client to fail and re-connect. The client does not fail. The client ignores the `{error, closed}` returned from `gen_tcp:send/2`.

``` erlang
-module(riak_repl_client_test).

-compile(export_all).

test_client_ports_through_client() ->
    Pids = get_client_pids(),
    [{Pid, send_keepalive_through_client(Pid)} || Pid <- Pids].

test_client_ports() ->
    Ports = get_client_ports(),
    [{Pid, send_keepalive(Port)} || {Pid, Port} <- Ports].

close_client_ports() ->
    Ports = get_client_ports(),
    [{Pid, close_port(Port)} || {Pid, Port} <- Ports].

get_client_ports() ->
    Pids = get_client_pids(),
[...]

Issue #56, opened 2012-04-18
Zombie server processes


Vagabond:
The send_timeout_close option seems to generate zombie server processes when exposed to bad network conditions.

Vagabond:
Closed by #102

Issue #55, opened 2012-04-17
Slow connections can cause the creation of many repl server processes


dreverri:
## Description

It appears that slow WAN connections can cause communication between two data centers to fail but the server process is never killed. The client data center opens a new connection which creates a new server process. The server processes consume more and more memory over time until the node runs out of memory.
## Steps to reproduce
- Start two servers
- Install Riak EE
- Setup MDC replication between the servers
- Add several large objects (1MB)
- Use `netem` to add artificial latency to the connection (6000ms)
- Wait for processes to build up
## Vagrant example

The following Vagrantfile and provisioning script will setup the above scenario. You have to replace RIAK_EE_URL in the provisioning script with the appropriate URL for the amd64 deb package.
### Vagranfile

``` ruby
# -*- mode: ruby -*-
# vi: set ft=ruby :

Vagrant::Config.run do |config|
        [{:name => :one, :ip => "192.168.33.11", :other_ip => "192.168.33.12"},
         {:name => :two, :ip => "192.168.33.12", :other_ip => "192.168.33.11"}]
        .each do |m|
                config.vm.define m[:name] do |c|
                        c.vm.box = "lucid64"
                        c.vm.box_url = "http://files.vagrantup.com/lucid64.box"
                        c.vm.network :hostonly, m[:ip]
                        c.vm.provision :shell do |s|
                                s.path = "setup-riak-ee.sh"
                                s.args = "#{m[:ip]} #{m[:other_ip]} &"
                        end
                end
         end
end
```
### setup-riak-ee.sh

``` bash
#!/bin/bash

IP=$1
OTHER_IP=$2

sudo apt-get -y install curl

echo "Download riak-ee"
curl --silent RIAK_EE_URL -o riak-ee_1.1.1-1_amd64.deb

echo "Install riak-ee"
sudo dpkg -i riak-ee_1.1.1-1_amd64.deb

echo "Start riak-ee"
riak start

echo "Wait for riak-ee"
riak-admin wait-for-service riak_kv riak@127.0.0.1

echo "Add data"
dd if=/dev/urandom count=1024 bs=1024 of=1MB_file
for i in {1..10}
do
        curl --silent http://127.0.0.1:8098/riak/b/$i -XPUT -d @1MB_file \
                -H 'content-type:text/plain'
done

echo "Add delay"
tc qdisc add dev eth1 root netem delay 6000ms

echo "Add listener"
riak-repl add-listener riak@127.0.0.1 $IP 9010

[...]

PR #102, opened 2012-08-21
Manual error handling for send timeouts, bump send timeout to 5m

Vagabond:
See also #69, #55 and #56.


bookshelfdave:
Confirmed that issue #69 if fixed, @dreverri confirmed that issue #55 fixed.

+1

PR #103, opened 2012-08-22
Better cleanup of open file handles on fullsync exit

Vagabond:


bookshelfdave:
+1
Observed files being closed, stopped fullsync worker after client kill -9'd.

PR #104, opened 2012-08-27
Do all ring changes inside ring_trans

Vagabond:
Helps avoid problems when the ring changes between get_ring and applying the change.


Vagabond:
I don't think we care anymore.

PR #105, opened 2012-08-28
Avoid dropping realtime messages right after election

Vagabond:
The default mailbox size of the elected leader is 'undefined' until a
mailbox check message comes in. This means that all messages received
before the first mailbox check will be dropped. This patch changes the
behaviour to reset the elected leader's known mailbox size to 0 on every
election.


slfritchie:
+1 to merge, sorry I wrote that bug.

PR #106, opened 2012-09-08
Change erlang:now() -> os:timestamp() when it is safe

Vagabond:


src/riak_repl_keylist_client.erl
```diff
@@ -125,7 +125,7 @@ request_partition(continue, #state{partitions=[P|T], work_dir=WorkDir, socket=So
                                                                  KeyListFn),
     {next_state, request_partition, State#state{kl_fn=KeyListFn,
             our_kl_ready=false, their_kl_ready=false,
-            partition_start=now(), stage_start=now(), skipping=false,
+            partition_start=os:timestamp(), stage_start=os:timestamp(), skipping=false,
```
jonmeredith, line 128: Replacement code is fine - but why double-call?  Wonder if this would work

```
partition_start=os:timestamp()=Start, stage_start=Start, skipping=false,
```

Low enough frequency event it doesn't matter.

jonmeredith:
+1 merge

PR #109, opened 2012-10-13
when using inverse_connection stops sites on loss of leadership

joecaswell:


bookshelfdave:
Once riak_ee works in riak_test, I would love to see a test for this.

bookshelfdave:
+1

PR #112, opened 2012-10-25
new_leader state correction

joecaswell:
should check Leader against State0 not State so leader change processes


Vagabond:
Superseded by #113.

PR #113, opened 2012-10-30
Adt new leader

Vagabond:
Fix leader changing that was broken in 1.2.0.


Vagabond:
Against wrong branch, needs to be against 1.2.

Vagabond:
See #114.

PR #114, opened 2012-10-30
Adt new leader

Vagabond:
Against 1.2 instead.


joedevivo:
+1 verified 10 times with riak_test

PR #130, opened 2012-11-09
Show fullsync/realtime enabled/running in riak-repl status and JSON stats

bookshelfdave:
riak_repl2_fscoordinator:is_running doesn't seem to be working correctly.

### DIFF
--- a/src/riak_repl2_fscoordinator.erl
+++ b/src/riak_repl2_fscoordinator.erl
@@ -51,7 +51,7 @@
 %% ------------------------------------------------------------------
 
 -export([start_link/1, start_fullsync/1, stop_fullsync/1,
-    status/0, status/1, status/2]).
+    status/0, status/1, status/2, is_running/1]).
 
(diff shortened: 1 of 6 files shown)


src/riak_repl2_fscoordinator.erl
```diff
@@ -151,6 +156,27 @@ handle_call(status, _From, State = #state{socket=Socket}) ->
     ],
     {reply, SelfStats, State};
 
+handle_call(is_running, _From, State) ->
+    PartitionEntry = lists:keytake(self(), 1, State#state.running_sources),
+    case PartitionEntry of
+        false ->
+            {reply, false, State};
```
lordnull, line 167: Partition entry will always be false.  The fscoodinator never puts itself in the running sources list.  The lists:keytake and resulting case clause can be removed.

lordnull:
+1.  Fullsync and realtime correctly report they are enabled/running to the appropriate clusters, even when there is more than one.

PR #138, opened 2012-11-29
Mw limit running sinks

buddhisthead:
ok, this time with the right base branch.
This is a PR on Micah's behalf.
Testing is done and beams are shipped to Anya already.

This PR adds the ability to limit the number of in-bound connections to a sink node.


buddhisthead:
Tested and baked.
Nice work, Micah!

+1

PR #139, opened 2012-12-03
Cet fs reservations

buddhisthead:
We experienced a problem at Anya where we had too many syncs running against the sink node. This caused latencies in the 90th percentile that were unacceptable. The issue is that the sink node was overloaded by too many syncs in progress that caused regular put/get slow down.

Technically, the problem was caused by a an over-commitment on the sink node. The protocol sends "whereis" requests from source to sink (via the full sync coordinator). Those return the IP address of the node that owns a partition too quickly and we've already scheduled up too much work for the sink node that hasn't actually started any work yet.

Our solution is to put in place a reservation system where the source side gets back a location_busy if the sink node has already got too many reservations placed against it. The sink side manages the reservations.

### DIFF
--- a/ebin/riak_repl.app
+++ b/ebin/riak_repl.app
@@ -28,6 +28,7 @@
                   'riak_repl2_fscoordinator_sup',
                   'riak_repl2_fscoordinator_serv',
                   'riak_repl2_fscoordinator_serv_sup',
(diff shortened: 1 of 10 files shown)


buddhisthead:
+1

PR #140, opened 2012-12-05
Improve whereis lookup, maxing out fullsync, and avoiding deadlocks due to race conditions

lordnull:


buddhisthead:
+1 for me. Thanks for the commentary in the files. Performance questions can wait until we've tied it at customer site.

PR #145, opened 2012-12-13
add an indicator that shows rt src/sink errors

bookshelfdave:
used to indicate that a fullsync may be needed (at the users discretion). They'll need to monitor both source and sink for errors.

Any thoughts?


bookshelfdave:
I'll have to clean this up in BNW.

PR #147, opened 2012-12-14
Merge Brave New World in preparation for 1.3

bookshelfdave:


rebar.config
```diff
@@ -5,10 +5,10 @@
 {erl_first_files, ["src/gen_leader.erl"]}.
 {lib_dirs, [".."]}.
 {deps, [
-        {riak_kv, ".*", {git, "git://github.com/basho/riak_kv", {branch, "master"}}},
-        {riak_search, ".*", {git, "git://github.com/basho/riak_search",
-                                 {branch, "master"}}},
-        {meck, ".*", {git, "git://github.com/eproxus/meck.git", "HEAD"}},
+        {riak_kv, "1.2.*", {git, "git://github.com/basho/riak_kv", {tag, "1.2.1p3"}}},
```
jaredmorrow, line 8: These should remain pointing at .\* of branch master.  If riak_kv has what it needs in 'master' this shouldn't be a problem.

buddhisthead:
I don't understand why we see this many commits in the merge. I merged BNW to master only a few weeks ago. This looks like it contains lots of commits prior to my merge. Is that just the way it happens when you re-merge a branch again?

jaredmorrow:
No, it should not have this many commits if it has been merged already.  The problem is that this is coming form another branch and not the actual bravenewworld branch.

I think it would be better just to merge by hand.  'git pull' on master, make sure your bravenewworld branch is up-to-date and then

`git merge --no-commit --no-ff bravenewworld` then review the conflicts and commit / push.

jaredmorrow:
Actually it looks pretty terrible in github, here's the compare view of commits https://github.com/basho/riak_repl/compare/bravenewworld

On the command line

```
➜  riak_repl git:(master) ✗ git log origin/master..bravenewworld --oneline | wc -l
     274
```

So yeah, need to merge to master more often me thinks.

buddhisthead:
I think something got botched since my last merge, which was super clean.
There are some ancient changes here, that appear to have gotten re-merged at some point.
But looking at the comparison, I think it will work out alright.

Vagabond:
The reason you see all these old changes is that BNW was never merged into 1.2 (or master).

The comments are all comments on specific _commits_ included in this branch.

Vagabond:
Or, more accurately, the BNW merge to master was done all in one commit, somehow:

https://github.com/basho/riak_repl/commit/8b219e5c2769ca5d8f81a631b2ee848a9d4c0f98

Vagabond:
If we knew what commit that merge to master represented, we could construct a branch containing only commits _since_ that merge, rather than merging all 270+ commits.

bookshelfdave:
manually merged by Andrew.

PR #149, opened 2012-12-18
Improve realtime migration when riak_repl is stopping

Vagabond:
These changes are intended to avoid dropping realtime messages that
happen or are pending during riak_repl stopping. The key changes are
that we try to send realtime messages to the rtq_proxy first, which
avoids a race with the process dictionary. We also explicitly stop() the
repl2_rtq so that we can gurantee terminate() completes. To do that, we
needed to make it a transient child of its supervisor, so it wouldn't
immediately restart.

### DIFF
--- a/src/riak_repl2_rt.erl
+++ b/src/riak_repl2_rt.erl
@@ -140,12 +140,14 @@ postcommit(RObj) ->
     case riak_repl_util:repl_helper_send_realtime(RObj, riak_client:new(node(), undefined))++[RObj] of
         Objects when is_list(Objects) ->
             BinObjs = term_to_binary(Objects),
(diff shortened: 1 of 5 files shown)


Vagabond:
See also:

basho/riak_kv#454
basho/webmachine#106
basho/riak_test#128

bookshelfdave:
looks good, riak_test passes. A +1 depends on any pending changes to webmachine.

bookshelfdave:
replication2 test passes with latest webmachine tweaks.
+1

seancribbs:
@metadave @Vagabond What are this and basho/riak_test#128 waiting on?

PR #150, opened 2012-12-19
randomize remote ips

lordnull:
Randomizes the list of IP's a remote cluster gives to a requesting cluster; this should minimize slamming a node on the remote side if multiple re-connects need to be made.


bookshelfdave:
Can we get riak_test/replication2 updated with this case?

buddhisthead:
Unit test passes. Tested branch with full and realtime sync. Everything works as expected, including uniform distribution of realtime connections from source to sink.

+1 to merge

PR #151, opened 2012-12-31
Fixed IP match in cases of subnet mismatch

lordnull:
if
    \* On the sink, if the cluster_mgr is set to listen on {0.0.0.0}
    \* The source has an external IP on a given subnet mask
    \* The sink has an external IP using the same subnet mask
    \* The source and sink IP's are on different subnets
an ip would fail to match.
This patch checks to see if the two IP's are both external or both
internal, and in the case of a match, uses the IP unless a better one
(ie, subnet mask match) is found.


bookshelfdave:
a manual test of this seem to pass, +1

the fun in get_matching_address/2 is getting huge.

lordnull:
Yeah, I'll take that function and break it up a bit, as well as add tests. For now, merging.

PR #152, opened 2013-01-02
added stats to indicate rt errors

bookshelfdave:
If a source node detects a RT error, increment the rt_source_error or rt_sink_error stat. Any increment to rt_source_error or rt_sink_error will automatically incremement rt_dirty. The rt_dirty value is written to a file to persist across restarts. When any of these stats are incremented and their value == 1, the node is registered with the FS coordinator (if it's running). Also, if a node is restarted, it will again try to register itself with the coordinator if its rt_dirty value > 0. When a fullsync is complete, the coordinator will clear rt_dirty (and remove the rt_dirty file from the filesystem) from all registered dirty nodes. If any nodes become dirty during fullsync, keep them rt_dirty even if the fullsync completes. The next pass at fullsync will clear this status. 

Note: if there is some type of disconnect between the source node and the coordinator, restarting the source node will re-register with the coordinator if it's still dirty.

misc:
you can test json stats with this:
curl -q http://localhost:8091/riak-repl/stats | python -mjson.tool

### DIFF
--- a/src/riak_repl2_fscoordinator.erl
+++ b/src/riak_repl2_fscoordinator.erl
@@ -689,9 +689,10 @@ notify_rt_dirty_nodes(State = #state{dirty_nodes = DirtyNodes,
     State1 = case ordsets:size(DirtyNodes) > 0 of
         true ->
             lager:info("Notifying dirty nodes after fullsync"),
-            % if there are nodes that are in BOTH sets, then don't notify
-            % those nodes below
-            NodesToNotify = ordsets:subtract(DirtyNodes, DirtyNodesDuringFS),
+            % notify all nodes in case some weren't registered with the coord
+            AllNodesList = riak_core_node_watcher:nodes(riak_repl),
+            NodesToNotify = lists:subtract(AllNodesList,
+                                           ordsets:to_list(DirtyNodesDuringFS)),
             lager:info("Notifying nodes ~p", [ NodesToNotify]),
             rpc:multicall(NodesToNotify, riak_repl_stats, clear_rt_dirty, []),
             % TODO: Check return values, only clear those that passed
--- a/src/riak_repl2_fscoordinator.erl
+++ b/src/riak_repl2_fscoordinator.erl
@@ -234,9 +234,11 @@ handle_call({node_dirty, Node}, _From,
         end,
     {reply, ok, NewState};
 
-handle_call({node_clean, Node}, _From, State) ->
+handle_call({node_clean, Node}, _From, State = #state{dirty_nodes=DirtyNodes}) ->
     lager:debug("Marking ~p clean after fullsync", [Node]),
-   {reply, ok, State};
+    NewDirtyNodes = ordsets:del_element(Node, DirtyNodes),
+    NewState = State#state{dirty_nodes=NewDirtyNodes},
+    {reply, ok, NewState};
 
 handle_call(_Request, _From, State) ->
     lager:info("ignoring ~p", [_Request]),
--- a/src/riak_repl2_fscoordinator.erl
+++ b/src/riak_repl2_fscoordinator.erl
@@ -43,7 +43,10 @@
     running_sources = [],
     successful_exits = 0,
     error_exits = 0,
-    pending_fullsync = false
+    pending_fullsync = false,
+    dirty_nodes = ordsets:new(),          % these nodes should run fullsync
+    dirty_nodes_during_fs = ordsets:new() % these nodes reported realtime errors
+                                          % during an already running fullsync
 }).
 
 %% ------------------------------------------------------------------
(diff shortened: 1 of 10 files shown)


src/riak_repl2_fscoordinator.erl
```diff
@@ -628,3 +672,60 @@ is_fullsync_in_progress(State) ->
         _ ->
             true
     end.
+
+is_running_priv(State) ->
```
lordnull: There should be a function at the bottom called is_fullsync_in_progress, which takes the state.

bookshelfdave:
Looking into the case where realtime is enabled and running, but fullsync is not enabled.

src/riak_repl_stats.erl
```diff
@@ -98,6 +104,34 @@ elections_elected() ->
...
+
+rt_dirty() ->
+    touch_rt_dirty_file(),
+    increment_counter(rt_dirty),
+    Stat = lookup_stat(rt_dirty) + 1,  % we know it's at least 1
+    % increment counter is a cast, so if the number is relatively small
+    % then notify the server. otherwise, we don't want to spam the
+    % coordinator. Once this value is reset to 0, we'll notify the
+    % coordinator again.
+    case Stat > 0 andalso Stat < 5 of
```
bookshelfdave, line 127: hacky, I know

bookshelfdave:
riak_test code for this located at:
https://github.com/basho/riak_test/pull/141

lordnull:
eunit tests failed. If the rt dirty file reports the node is dirty, the stats set up is dependant on fscoordinator:node_dirty/1, which calls risk_core_cluster_mgr:get_leader/0 and fscooridinator_sup:started/1.

bookshelfdave:
@lordnull do you have the latest commit (db47f06) that "fixes" eunit issues? All 86 tests pass for me, although I have seen some of the connection mgr tests fail in the past.

The call to fscoordinator:node_dirty/1 won't do anything if the coordinator isn't running. I'm thinking of putting in some code after a fullsync completes that rpc:multicalls every node in the cluster (minus nodes that are currently marked as dirty during a fullsync) to clear the rt_dirty state. The good thing is that the node will stay marked dirty.

Edit: this change makes registering an rt_dirty node somewhat redundant because the coord will clear all nodes anyways. But I think to leave it in is fine, as rt_dirty nodes will show up in coordinator stats across the cluster.

src/riak_repl2_fscoordinator.erl
```diff
@@ -628,3 +672,47 @@ is_fullsync_in_progress(State) ->
...
+                                     DirtyNodesDuringFS}) ->
+    State1 = case ordsets:size(DirtyNodes) > 0 of
+        true ->
+            lager:info("Notifying dirty nodes after fullsync"),
+            % if there are nodes that are in BOTH sets, then don't notify
+            % those nodes below
+            NodesToNotify = ordsets:subtract(DirtyNodes, DirtyNodesDuringFS),
+            lager:info("Notifying nodes ~p", [ NodesToNotify]),
+            rpc:multicall(NodesToNotify, riak_repl_stats, clear_rt_dirty, []),
+            % TODO: Check return values, only clear those that passed
```
buddhisthead: Is this TODO important to implement before we ship this next build? I presume that if the RPC call fails, then the node won't have it's dirty flag cleared and it will re-post it's status of dirty. So, it's ok if it fails?

bookshelfdave: @buddhisthead no, it's not important for the release and should be removed. I think it's important that if the call fails, the node will still stay "dirty".

In the latest commit that I made, I actually rpc:multicall to all nodes (minus those in the DirtyNodesDuringFS set) to clear dirty status in case the coord wasn't running and the dirty status wasn't registered.

buddhisthead: Yeah, I like that broad spectrum approach. Simplifies things and reduces code paths.

buddhisthead:
Are you planning on removing all these lager:info statements or at least changing them to lager:debug? I think that normal operation shouldn't produce so much output, even though it's handy for debugging purposes. At least we shouldn't print anything to the console when there are no dirty nodes. Silence means it's OK.

buddhisthead:
Works as advertised on a running cluster. I tested by starting up basho_bench and shutting down half of the six sink nodes in the other cluster. I was surprised that only two of the source nodes out of six reported errors, but I guess they were not connected uniformly to the sinks. After doing a fullsync, all dirty flags were cleared.

Nice.

If you remove/change the lager:info statements, I would happily give this a +1.

src/riak_repl2_fscoordinator.erl
```diff
@@ -177,30 +207,43 @@ handle_call(status, _From, State = #state{socket=Socket}) ->
...
+          true -> lager:debug("Node dirty during fullsync from ~p ", [Node]),
+                  NewDirty = ordsets:add_element(Node, DirtyDuringFS),
+                  State#state{dirty_nodes_during_fs = NewDirty};
+          false -> lager:debug("Node dirty from ~p ", [Node]),
+                   NewDirty = ordsets:add_element(Node, DirtyNodes),
+                   State#state{dirty_nodes = NewDirty}
+        end,
+    {reply, ok, NewState};
+
+handle_call({node_clean, Node}, _From, State = #state{dirty_nodes=DirtyNodes}) ->
```
bookshelfdave, line 237: This may be useful if something goes wrong at runtime.

bookshelfdave:
@buddhisthead @lordnull changes pushed, thanks for the review!

PR #153, opened 2013-01-03
Fixes fullsync poor capacity by setting application env values on all nodes

buddhisthead:
The reason that fullsync was not maxing out it's connections was that the console commands to change the maximums (fssource, fssink, cluster) was only changing the application env on the node where the command was issued.

The solution used here was to make an rpc multi-call (using a riak_core_utility function) to set the app env value on all nodes of the riak cluster. Then, running a fullsync, I observed that the coordinator would correctly schedule multiple syncs per node, according to the maximums set.

While in there, I also added a couple of info and debug statements to make diagnosis of future problems easier. It might also be nice to print these maximums in the repl-status, but that's for another day.

One potential problem I see here is when the maximums are altered via the console, and then a node joins the system; the new node will not have the new value. This is a problem because new nodes often become the coordinator. It seems unreasonable for customers to have to reset the maximums after every node bounce. I don't have a good answer for this yet.

### DIFF
--- a/src/riak_repl_console.erl
+++ b/src/riak_repl_console.erl
@@ -26,8 +26,7 @@
 -export([modes/1, set_modes/1, get_modes/0,
          max_fssource_node/1,
          max_fssource_cluster/1,
-         max_fssink_node/1,
-         set_fullsync_param/2]).
+         max_fssink_node/1]).
 
 add_listener(Params) ->
     Ring = get_ring(),
@@ -437,11 +436,6 @@ modes(NewModes) ->
 %% becomes the fullsync coordinator will have the correct values. TODO: what happens when a
 %% machine bounces and becomes leader? It won't know the new value. Seems like we need a central
 %% place to hold these configuration values.
-set_fullsync_param(Param, Value) ->
-    Ring = get_ring(),
-    NewRing = riak_repl_ring:set_simple_param(Ring,Param,Value),
-    maybe_set_ring(Ring, NewRing).
-
 max_fssource_node([]) ->
     %% show the default so as not to confuse the user
     io:format("max_fssource_node value = ~p~n",
@@ -450,7 +444,6 @@ max_fssource_node([]) ->
 max_fssource_node([FSSourceNode]) ->
     NewVal = erlang:list_to_integer(FSSourceNode),
     riak_core_util:rpc_every_member(?MODULE, max_fssource_node, [NewVal], ?CONSOLE_RPC_TIMEOUT),
-    set_fullsync_param(max_fssource_node, NewVal),
     max_fssource_node([]),
     ok;
 max_fssource_node(NewVal) ->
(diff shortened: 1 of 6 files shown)


buddhisthead:
Notes:
Store the parameters in the repl part of the ring.
When FS coordinators start, check the ring. if param is undefined, write my app env value to the ring.
                  Otherwise, use ring's value and update my env value.
When console commands update the value, update the ring.

buddhisthead:
Ok, made the changes to store the fullsync parameters in the replication configuration in the ring.
We only write to the ring in the cases where the ring's values are undefined (like first boot of cluster) and when they are changed by a console command.

However, after more thought, I am going to back this out. If the cluster is shutdown completely and the app config vars are changed, then they would be ignored since the ring persists. So, sadly the only answer that works is to make sure the app config files are updated to the desired values before reboot.. If they are not the same on all nodes, then whatever node happens to get assigned as coordinator is going to trump any others that don't match it's config.

buddhisthead:
Ready for review.

src/riak_repl_console.erl
```diff
@@ -431,6 +431,11 @@ modes(NewModes) ->
     set_modes(Modes),
     modes([]).
 
+%% For each of these "max" parameter changes, we need to make an rpc multi-call to every node
+%% so that all nodes have the new value in their application environment. That way, whoever
+%% becomes the fullsync coordinator will have the correct values. TODO: what happens when a
+%% machine bounces and becomes leader? It won't know the new value. Seems like we need a central
+%% place to hold these configuration values.
 max_fssource_node([]) ->
     %% show the default so as not to confuse the user
```
lordnull, line 440: ~~Seems more confusing to always show the default as this is called after setting the value; shouldn't it show the last set value, or what acutally will be used?~~

Displays set value, I prolly should have looked up the macro.

src/riak_repl_console.erl
```diff
@@ -438,9 +443,11 @@ max_fssource_node([]) ->
                                   ?DEFAULT_SOURCE_PER_NODE)]);
 max_fssource_node([FSSourceNode]) ->
     NewVal = erlang:list_to_integer(FSSourceNode),
-    application:set_env(riak_repl, max_fssource_node, NewVal),
+    riak_core_util:rpc_every_member(?MODULE, max_fssource_node, [NewVal], ?CONSOLE_RPC_TIMEOUT),
     max_fssource_node([]),
-    ok.
+    ok;
+max_fssource_node(NewVal) ->
```
lordnull, line 449: Not sure what purpose this clause (and the others like it) serve.

lordnull:
My comment is a nit-pick, the solutions works (i.e.: the settings get shared).

src/riak_repl_console.erl
buddhisthead, line 449: That is the line that called via the RPC multi-call. Without it, the pattern match would fail during the RPC.

lordnull:
:+1:

PR #154, opened 2013-01-03
Refactor IP code out into its own module, with tests

Vagabond:


lordnull:
Tests pass, and manually running a fullsync still works, so :+1: .

PR #155, opened 2013-01-04
Make bloom folding avoid giant message queues

Vagabond:
Additionally, provide a way to disable bloom folding via an app env which
will short-circuit the capabilities negotiation.

Tested with riak_test master with replication2 test with both {bloom_fold, true} and {bloom_fold, false} in app env.


buddhisthead:
+1

PR #156, opened 2013-01-04
return ok instead of whatever I felt like that day

bookshelfdave:


lordnull:
Passing tests give me the warm fuzzes :+1: .

PR #157, opened 2013-01-04
Gracefully handle partition conflicts during fullsync

Vagabond:
The fssource_sup uses the partition in the childspec as the ID, so 2
fssources can't run for a single partition. However we weren't catching
a fssource startup failure, which was crashing the coordinator.


Vagabond:
I tested this with a 3 2-node clusters (from a devrel). Set the max_fssource_node, max_fssource_cluster and max_fssink_node values to 30, configure fullsync from A->B and A->C and do a 'riak-repl fullsync start'.

Without this patch, you'll get crashes about {error,{already_started,Pid}} badmatches, which crash the fscoordinator.

PR #158, opened 2013-01-10
add per-cluster fullsync scheduling

bookshelfdave:
example:
{riak_repl, [
  {data_root, "./data/riak_repl/"},
  % clusters foo + bar have difference intervals (in minutes)
  {fullsync_interval, [{"bar", 1}, {"foo", 2}]}
]}

### DIFF
--- a/src/riak_repl_util.erl
+++ b/src/riak_repl_util.erl
@@ -620,10 +620,9 @@ schedule_cluster_fullsync(Cluster, Pid) ->
                     start_fullsync_timer(Pid, FullsyncIvalMins),
                     ok
             end;
-        {ok, Tuple} when is_tuple(Tuple) ->
-            {_Cluster, FullsyncIvalMins} = Tuple,
-            start_fullsync_timer(Pid, FullsyncIvalMins),
-            ok;
+        {ok, {Cluster, FullsyncIvalMins}} ->
+            start_fullsync_timer(Pid, FullsyncIvalMins);
+        {ok, Tuple} when is_tuple(Tuple) -> ok;
         {ok, FullsyncIvalMins} ->
             %% this will affect ALL clusters that have fullsync enabled
             start_fullsync_timer(Pid, FullsyncIvalMins)
--- a/src/riak_repl_util.erl
+++ b/src/riak_repl_util.erl
@@ -34,6 +34,8 @@
          repl_helper_send_realtime/2,
          schedule_fullsync/0,
          schedule_fullsync/1,
+         schedule_cluster_fullsync/1,
+         schedule_cluster_fullsync/2,
          elapsed_secs/1,
          shuffle_partitions/2,
          proxy_get_active/0,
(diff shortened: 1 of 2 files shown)


src/riak_repl_util.erl
```diff
@@ -562,15 +564,51 @@ configure_socket(Transport, Socket) ->
...
+            ok;
+        {ok, [{_,_} | _] = List} ->
+            case proplists:lookup(Cluster, List) of
+                none -> ok;
+                {_, FullsyncIvalMins} ->
+                    start_fullsync_timer(Pid, FullsyncIvalMins),
+                    ok
+            end;
+        {ok, Tuple} when is_tuple(Tuple) ->
+            {_Cluster, FullsyncIvalMins} = Tuple,
```
bookshelfdave: I should check to see if the first value in the tuple == Cluster.

Vagabond: You could just have 2 clauses; {ok, {Cluster, FullsyncIvalMins}} -> ...; and then an {ok, Tuple} when is_tuple() -> ok;

src/riak_repl_util.erl
```diff
@@ -562,15 +564,51 @@ configure_socket(Transport, Socket) ->
...
+
+schedule_cluster_fullsync(Cluster, Pid) ->
+    case application:get_env(riak_repl, fullsync_interval) of
+        {ok, disabled} ->
+            ok;
+        {ok, [{_,_} | _] = List} ->
+            case proplists:lookup(Cluster, List) of
+                none -> ok;
+                {_, FullsyncIvalMins} ->
+                    start_fullsync_timer(Pid, FullsyncIvalMins),
```
Vagabond: Why not use the erlang:send_after like the olf schedule_fullsync does?

bookshelfdave: The call to riak_repl2_fscoordinator is a cast, not a call. In the older repl, start_fullsync was a call. I felt like there were less changes to the existing code by just using gen_server:cast.

src/riak_repl_util.erl
```diff
@@ -562,15 +564,50 @@ configure_socket(Transport, Socket) ->
...
+                    ok
+            end;
+        {ok, {Cluster, FullsyncIvalMins}} ->
+            start_fullsync_timer(Pid, FullsyncIvalMins);
```
Vagabond: It might be nice to log "Fullsync for Cluster scheduled in 2 hours" or something...

bookshelfdave: Ok. Unless you have a strong opinion regarding hours, I'd rather show minutes. We don't have to have any different cases for intervals smaller than 1 hour.

buddhisthead:
Dave,

This is cool. What do you think about having full sync run "automatically" when we detect the rt dirty flag? It could be some minutes later to avoid slamming a cluster. Maybe some kind of maximum frequency. But it could be cool if it ran when it needed to. Just a thought.

bookshelfdave:
Chris -
  I could see that as being an app.config param:
...
{fullsync_on_connect, false},
{fullsync_on_rtdirty, [params...]}
...

I would add it as a feature for 2.0 to make sure we meet our 1.4 deadline.

buddhisthead:
Yep. I concur about delaying it for 2.0.
Just wanted to toss the idea out there. Thanks for discussing.

bookshelfdave:
@buddhisthead I created a 2.0 issue for that feature: https://github.com/basho/riak_repl/issues/196

bookshelfdave:
TODO:
- [x] remove riak_repl_stats:server_fullsyncs() when fullsync finishes in BNW
- [x] add per cluster server_fullsync stat
- [x] a riak_test that uses per cluster server_fullsync stat

bookshelfdave:
see https://github.com/basho/riak_test/pull/221 for a riak_test.

buddhisthead:
Be aware that with the latest changes to dependencies of protocol buffer APIs, this branch needs to be merged into the latest master so that it will compile. Don't just checkout the branch :-)

buddhisthead:
Besides these comments, everything looks great. Tests Pass.

bookshelfdave:
I rebased and pushed this morning so the deps should be fine now.

buddhisthead:
+1 to merge. Thanks, Dave. Nice work.

PR #169, opened 2013-01-18
Quickcheck test for RTQ

bookshelfdave:
This is @Vagabond's QC test for the repl RT queue. 

The only real addition is setting the max # of bytes in the queue before starting to exercise q trimming. 

Initially, I started down the path of keeping a shadow ETS table in the QC model, but the model ended up using most of the trimming code from the rtq to shadow the size of the queue. This did not seem ideal as the test was using the same code that it was testing. Now I only assert that the queue size is always <= max_bytes which is set upon test startup. 

What else should this test include?

### DIFF
--- a/src/riak_repl2_rtq.erl
+++ b/src/riak_repl2_rtq.erl
@@ -20,6 +20,7 @@
 -behaviour(gen_server).
 %% API
 -export([start_link/0,
+         start_test/0,
          register/1,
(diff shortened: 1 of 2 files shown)


Vagabond:
If trimming happens, if you don't apply the trim to the model, how can the expectation as to what the next item to be popped from the queue hold?

PR #177, opened 2013-01-24
rt_dirty + stats startup bug

bookshelfdave:
Wait 30 seconds to notify the coordinator of an rt_dirty node after startup. If the coordinator isn't up yet, just log a warning. A fullsync will clear the rt_dirty flag either way.


buddhisthead:
Tested (without the change to riak_test ... pull 178) and it passed! Yay!

+1

Issue #179, opened 2013-01-24
repl "no partition" info messages should be reworded


bookshelfdave:

bookshelfdave:
fixed in #186.

PR #186, opened 2013-01-30
Reworded the 'no partition' log message

lordnull:
Many more words, but states exactly what the situation is.

Addresses #179.


src/riak_repl2_fscoordinator.erl
```diff
@@ -564,7 +564,7 @@ determine_best_partition(State) ->
     determine_best_partition(SeedPart, Busies, Owners, Waiting, queue:new()).
 
 determine_best_partition({empty, _Q}, _Business, _Owners, _Waiting, AccQ) ->
-    lager:info("no partition"),
+    lager:info("No partition in the queue that will not exceede a limit; will try again later."),
```
bookshelfdave: exceede typo

PR #187, opened 2013-01-30
Restore the 'server_stats' and 'client_stats' for legacy repl

Vagabond:


src/riak_repl_console.erl
```diff
@@ -621,8 +617,9 @@ server_stats() ->
         undefined ->
             [{sources, RT2}];
         _ ->
-            [{sources, rpc:call(LeaderNode, ?MODULE, server_stats_rpc,
-                                     [])++RT2}]
```
bookshelfdave, line 625: won't this break for customers that are already scraping BNW repl stats?

bookshelfdave:
Both client and server can be testing with:
curl -q http://localhost:port/riak-repl/stats

src/riak_repl_wm_stats.erl
```diff
@@ -143,4 +165,44 @@ jsonify_stats_test_() ->
       ?assertEqual(Expected, jsonify_stats(Stats, []))
     end}].
 
+jsonify_stats_test_2() ->
+    [{"legacy server_stats and client_stats", fun() ->
```
Vagabond: The tuple here should just be added to the list inside jsonify_stats_test_()

Vagabond:
Closed in favor of PR against 1.3 branch.

PR #191, opened 2013-02-01
additional json stats testing

bookshelfdave:
This PR will melt your eyes. No, really. Wear a welding mask or something similar.


src/riak_repl_wm_stats.erl
```diff
@@ -155,7 +170,722 @@ jsonify_stats([{K,V}|T], Acc) ->
...
+                          {starting,0},
+                          {successful_exits,0},
+                          {error_exits,0},
+                          {busy_nodes,0},
+                          {running_stats,<<>>},
+                          {socket,<<>>},
+                          {fullsync_suggested,<<"dev1@127.0.0.1">>},
+                          {fullsync_suggested_during_fs,<<>>}]}]}],
+              ?assertEqual(Expected, jsonify_stats(Actual, [])),
+              _Result = mochijson2:encode({struct, Expected})
```
lordnull: Is the purpose of these lines (_Result = mochijson2:encode/2) to ensure the expected can actually be encoded? If so, maybe a note somewhere to say as much. I agree w/ the reasoning (if it can't be encoded the test will explode) it's just not obvious on first read.

bookshelfdave:
I should probably explain myself here. I generated a ton of input to be passed to jsonify_stats by kicking off repl 1.2/1.3 repl. This includes several cases that blew up before I added tests. Each term that was generated was the result of the calls in the first part of get_stats(). 
https://github.com/basho/riak_repl/blob/6b1bb8aef92185b809e273a49a0115966ccdd5d1/src/riak_repl_wm_stats.erl#L92

Ultimately, they are wrapped up in a monster list at:
https://github.com/basho/riak_repl/blob/6b1bb8aef92185b809e273a49a0115966ccdd5d1/src/riak_repl_wm_stats.erl#L104

The last 2 lines of most tests take the terms and try to jsonify them. Previously, this function would miss PID's, 3 and 4-tuples etc causing mochijson:encode to blow up. 

```
?assertEqual(Expected, jsonify_stats(Actual, [])),
_Result = mochijson2:encode({struct, Expected})
```

If the stats are jsonified and mochijson:encode doesn't blow up, then we are all good. Excessive, but these will catch more than we've had in the past.

src/riak_repl_wm_stats.erl
bookshelfdave: Yes, I'll add a comment to each mochijson2:encode.

bookshelfdave:
thanks @lordnull

PR #197, opened 2013-02-13
Replication using binary riak object formats

buddhisthead:
The new binary object support has been added to riak_kv via the branch jrw-pdx-binary-robj-rebased2.

To summarize, it changes how we store objects in KV such that we now store a binary form that no longer includes the bucket, key, or a bloated dictionary. This new format can also be used as a wire format in handoff, etc in core.

Replication can take advantage of the new format as well, by passing the binary object over socket connections. To do this, we needed to add a wrapper that includes the bucket and key since they no longer appear in the object. Thus, I've added a "wire" format. This format is denoted either w0 or w1. w0 is the old way and is binary compatible with older versions of replication (pre 1.4). w1 can be used when both connected nodes (source and sink) exchange a negotiated protocol of 1.1 or greater. I think Micah is currently upping our protocol to 2.0. This could be a problem, but I'll get to that separately.

Of note is that we now write the new w1 format into the RT queue because it's only the RT sources that know what protocol is supported with their connected sinks. If the sink is older, then the source will down-convert the binary format to w0 before sending.

We take care to maintain the same hashing method for the keylist differences. That will not be needed when we move to AAE full sync (except for connecting to old clusters). We also take care to fold over binary objects now. So, there has been a lot of plumbing changes to support passing K,B around as needed.

That's about it. All riak_tests for replication pass, but you'll need that riak_kv branch mentioned above.

### DIFF
--- a/src/riak_repl2_fssource.erl
+++ b/src/riak_repl2_fssource.erl
@@ -18,7 +18,8 @@
         cluster,
         connection_ref,
         fullsync_worker,
-        work_dir
+        work_dir,
+        ver
     }).
 
 start_link(Partition, IP) ->
@@ -53,7 +54,7 @@ init([Partition, IP]) ->
                   {nodelay, true},
                   {packet, 4},
                   {active, false}],
-    ClientSpec = {{fullsync,[{1,0}]}, {TcpOptions, ?MODULE, self()}},
+    ClientSpec = {{fullsync,[{1,1}]}, {TcpOptions, ?MODULE, self()}},
 
     %% TODO: check for bad remote name
     lager:info("connecting to remote ~p", [IP]),
@@ -66,8 +67,20 @@ init([Partition, IP]) ->
             {stop, Reason}
     end.
 
-handle_call({connected, Socket, Transport, _Endpoint, _Proto, Props}, _From,
+deduce_wire_version_from_proto({_Proto,{CommonMajor,CMinor},{CommonMajor,HMinor}}) ->
+    %% if common protocols are both >= 1.1, then we know the new binary wire protocol
+    case CommonMajor >= 1 andalso CMinor >= 1 andalso HMinor >= 1 of
+        true ->
+            %% new sink. yay! new wire protocol supported.
+            w1;
+        _False ->
+            %% old sink mandates old wire protocol only.
+            w0
+    end.
+
+handle_call({connected, Socket, Transport, _Endpoint, Proto, Props}, _From,
         State=#state{ip=IP, partition=Partition}) ->
+    Ver = deduce_wire_version_from_proto(Proto),
     Cluster = proplists:get_value(clustername, Props),
     lager:info("fullsync connection to ~p for ~p",[IP, Partition]),
 
@@ -85,7 +98,7 @@ handle_call({connected, Socket, Transport, _Endpoint, _Proto, Props}, _From,
     riak_repl_keylist_server:start_fullsync(FullsyncWorker, [Partition]),
     {reply, ok, State#state{transport=Transport, socket=Socket,
             cluster=Cluster,
-            fullsync_worker=FullsyncWorker, work_dir=WorkDir}};
+            fullsync_worker=FullsyncWorker, work_dir=WorkDir, ver=Ver}};
 handle_call(start_fullsync, _From, State=#state{fullsync_worker=FSW}) ->
     riak_repl_keylist_server:start_fullsync(FSW),
     {reply, ok, State};
--- a/src/riak_repl2_fssource.erl
+++ b/src/riak_repl2_fssource.erl
@@ -67,20 +67,10 @@ init([Partition, IP]) ->
             {stop, Reason}
     end.
 
-deduce_wire_version_from_proto({_Proto,{CommonMajor,CMinor},{CommonMajor,HMinor}}) ->
-    %% if common protocols are both >= 1.1, then we know the new binary wire protocol
-    case CommonMajor >= 1 andalso CMinor >= 1 andalso HMinor >= 1 of
-        true ->
-            %% new sink. yay! new wire protocol supported.
-            w1;
-        _False ->
-            %% old sink mandates old wire protocol only.
-            w0
-    end.
-
 handle_call({connected, Socket, Transport, _Endpoint, Proto, Props}, _From,
         State=#state{ip=IP, partition=Partition}) ->
-    Ver = deduce_wire_version_from_proto(Proto),
+    Ver = riak_repl_util:deduce_wire_version_from_proto(Proto),
+    lager:debug("Negotiated ~p wire format", [Ver]),
     Cluster = proplists:get_value(clustername, Props),
     lager:info("fullsync connection to ~p for ~p",[IP, Partition]),
 
--- a/src/riak_repl2_rt.erl
+++ b/src/riak_repl2_rt.erl
@@ -144,10 +144,11 @@ postcommit(RObj) ->
             %% map riak objects to their wire format, according to the highest
             %% commonly supported object format between the two clusters.
             Ver = v0,
-            BinObjs = case Ver of
-                          v0 -> riak_repl_util:to_wire(w0, Objects);
-                          _V -> riak_repl_util:to_wire(w1, Objects)
-                      end,
+            W = case Ver of
+                    v0 -> w0;
+                    _V -> w1
+                end,
+            BinObjs = riak_repl_util:to_wire(W, Objects),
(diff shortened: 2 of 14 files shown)


src/riak_repl2_fssource.erl
```diff
@@ -66,8 +67,20 @@ init([Partition, IP]) ->
             {stop, Reason}
     end.
 
-handle_call({connected, Socket, Transport, _Endpoint, _Proto, Props}, _From,
+deduce_wire_version_from_proto({_Proto,{CommonMajor,CMinor},{CommonMajor,HMinor}}) ->
```
bookshelfdave: This code is copied in several areas. Could it be shared in riak_repl_utils?

src/riak_repl2_rt.erl
```diff
@@ -138,8 +138,11 @@ get_sink_pids() ->
 %% Realtime replication post-commit hook
 postcommit(RObj) ->
     case riak_repl_util:repl_helper_send_realtime(RObj, riak_client:new(node(), undefined))++[RObj] of
```
bookshelfdave, line 140: @angrycub may be interested in these changes

src/riak_repl2_rtsource_conn.erl
```diff
@@ -151,20 +153,37 @@ handle_call({connected, Socket, Transport, EndPoint, Proto}, _From,
     %% before turning it active (e.g. handoff of riak_core_service_mgr to handler
     case Transport:send(Socket, <<>>) of
         ok ->
-            {ok, HelperPid} = riak_repl2_rtsource_helper:start_link(Remote, Transport, Socket),
+            Ver = deduce_wire_version_from_proto(Proto),
+            lager:info("Negotiated ~p wire format", [Ver]),
```
bookshelfdave: should this be a lager:debug instead?

src/riak_repl_keylist_server.erl
```diff
@@ -495,6 +497,15 @@ diff_bloom({Ref,diff_exchanged},  #state{diff_ref=Ref} = State) ->
 %% end of bloom states
 %% --------------------------------------------------------------------------
 
+encode_obj_msg(V, {fs_diff_obj, RObj}) ->
```
bookshelfdave: same thing here, can this be put into riak_repl_utils?

src/riak_repl2_rtsource_conn.erl
buddhisthead: Yes, absolutely.
efailuretocleanupdebugcode

src/riak_repl2_fssource.erl
buddhisthead: It's possible, but if you look closely, you will see that one "encode" deals with diff_obj and the other deals with fs_diff_obj. If we refactor the code to handle both, then possibly it can move to riak_repl_util. But I don't expect us to ever use that code again.

src/riak_repl2_rt.erl
buddhisthead, line 140: Who is that?

src/riak_repl2_fssource.erl
buddhisthead: Oops. Wrong line! Yep, I took your advice here and moved the "deduce..." function to util and all other files call it. Thanks.

buddhisthead:
@dp Thanks for the comments. I've moved the two functions you mentioned into riak_repl_util.erl and made calls to them from the common access points. If you have time, please also try and run the tests to confirm goodness before we merge. Thanks again!

buddhisthead:
@lordnull Micah, I know that I said the 2.0 protocol would be ok, but then I was thinking about how the connection negotiates protocols between client and service. It will pick the highest common Major number and then whatever minor numbers are on each side will get reported back.

That means that a connection between a 1.1 client and a 2.0 service will negotiate on 1.0 (I think it's .0). So, I think a change we need to make in the connection negotiation is to have the 2.0 dominate the 1.1 and report back 1.1 as the common protocol. Because surely the 2.0 side can speak 1.anything. The other route we could go is to bump the protocol up to 2.0 for binary objects as well.

While we are on it, I'd like to build caps-exchange into the protocol negotiation. That would be worthy of a 2.0 sticker!

src/riak_repl2_rt.erl
bookshelfdave, line 140: Charlie, he was looking into the post-commit hooks recently.

bookshelfdave:
Is the intent to merge to master, or would you prefer the integration branch?

src/riak_repl2_rtsource_helper.erl
```diff
@@ -38,22 +39,37 @@ status(Pid) ->
...
+                   socket = Socket, deliver_fun = Deliver, ver = Ver},
     async_pull(State),
     {ok, State}.
 
+%% @doc BinObjs are in new riak binary object format. If the remote sink
+%%      is storing older non-binary objects, then we need to downconvert
+%%      the objects before sending. V is the format expected by the sink.
+maybe_downconvert_binary_objs(BinObjs, V) ->
+    case V of
+        w1 ->
```
bookshelfdave, line 56: Would it make sense to add w0 and w1 constants + descriptions in riak_repl.hrl?

src/riak_repl_util.erl
```diff
@@ -688,3 +711,172 @@ sockname(Socket, Transport) ->
...
+      BLen:32/integer, B:BLen/binary,
+      KLen:32/integer, K:KLen/binary, BinObj/binary>>.
+
+%% @doc Return the wire format version of the given wire blob
+wire_version(<<131, _Rest/binary>>) ->
+    w0;
+wire_version(<<?MAGIC:8/integer, ?W1_VER:8/integer, _Rest/binary>>) ->
+    w1;
+wire_version(<<?MAGIC:8/integer, N:8/integer, _Rest/binary>>) ->
+    %% TODO: convert to an atom like atom_of('w' + N)
```
bookshelfdave: is this TODO still relevant?

src/riak_repl_util.erl
```diff
@@ -688,3 +711,172 @@ sockname(Socket, Transport) ->
...
+    term_to_binary(Objects);
+to_wire(w0, Object) ->
+    term_to_binary(Object);
+to_wire(w1, Objects) when is_list(Objects) ->
+    BObjs = [to_wire(w1,O) || O <- Objects],
+    term_to_binary(BObjs);
+to_wire(w1, Object) when not is_binary(Object) ->
+    B = riak_object:bucket(Object),
+    K = riak_object:key(Object),
+    to_wire(w1, B, K, Object).
```
bookshelfdave, line 763: Should you have

```
to_wire(w1, Object) when is_binary(Object) ->
  Object.
```

just in case?

bookshelfdave:
Code seems fine, eunit tests pass. 
There are a couple issues with the riak_tests failing, but I don't think they are related:
https://gist.github.com/metadave/87c05efd478b5141ed94
https://gist.github.com/metadave/34f3ec691d862743140c

src/riak_repl_util.erl
buddhisthead, line 763: Hmm. I don't think so. The code would have to check the binary version of the object and if it was in old binary format, it would have to convert to new binary format. Also, w1 is a wire format, not the object format. to_wire() is a function for packaging non-binary objects. I think it's best to leave it that way. Otherwise, it will have to handle conversions.

src/riak_repl2_rtsource_helper.erl
buddhisthead, line 56: Agreed. I moved the type def for w0 and w1 into the include file with some description.

src/riak_repl_util.erl
buddhisthead: Done. Thanks for pointing this out.

buddhisthead:
I am not sure about merging. I probably should have branched off the integration branch, but at this point I don't want to try and rebase. That will be painful and I'll screw something up. Let's just merge to master once all is happy. However, we can't merge until the change to KV has happened here:

https://github.com/basho/riak_kv/pull/479

since the build will break without those binary object changes.

buddhisthead:
In going through the commit notes of basho/riak_kv#479, it occurred to me that I don't look at the capabilities of the cluster to see what version the cluster wants. I assume that the put_fsm will eventually write the correct format to the disk. The two connected clusters will negotiate their protocol version based on the protocol numbers of the fullsync or RT sync pairs.

buddhisthead:
Dave, are you good for me to merge this PR? I've re-run tests after re-basing against master.

bookshelfdave:
:+1: merge away

PR #199, opened 2013-02-15
Fix incorrect use of riak_core_vnode:monitor

Vagabond:
riak_core_vnode_monitor does not take a raw pid as an argument, it
function_clauses instead.


kellymclaughlin:
+1 to merge. Nice catch!

PR #201, opened 2013-02-18
don't report an rt_sink_error w/ backpressure

bookshelfdave:


bookshelfdave:
see https://github.com/basho/riak_repl/pull/202

Issue #213, opened 2013-03-13
Running riak-repl status on node startup can cause an exception


bookshelfdave:
Attempting to restart script through sudo -H -u riak
RPC to 'riak@10.1.35.184' failed: {'EXIT',
                                   {function_clause,
                                    [{lists,sort,
                                      [{error,{not_registered,riak_repl}}],
                                      [{file,"lists.erl"},{line,414}]},
                                     {riak_repl_console,status2,1,
                                      [{file,"src/riak_repl_console.erl"},
                                       {line,122}]},
                                     {rpc,'-handle_call_call/6-fun-0-',5,
                                      [{file,"rpc.erl"},{line,203}]}]}}

russelldb:
Is it coming from here https://github.com/basho/riak_core/blob/master/src/riak_core_stat_cache.erl#L89?

Are you using the cache for stats, but not registering with the cache, or maybe calling get_stats(riak_repl) before registering? Might be better for the cache to return an empty list, or maybe call the produce stats fun.

bookshelfdave:
New PR: https://github.com/basho/riak_repl/pull/215

russelldb:
All I was thinking is it might be better for the stat cache to return the empty list rather than an error…saves every app from coding for it.

bookshelfdave:
fixed for 1.4

PR #214, opened 2013-03-14
safely get stats during startup

bookshelfdave:
See: https://github.com/basho/riak_repl/issues/213

Also: hey, I didn't really get any sleep last night, so you know, this might be totally off base.


bookshelfdave:
Actually, forgot the JSON stats.

PR #215, opened 2013-03-14
safely get stats during startup

bookshelfdave:


lordnull:
I tested this on a rebased version of your branch (just to avoid annoying rebar dependancy issues) and it works as advertised.  +1.

PR #219, opened 2013-03-15
Fix replication of binary objects for RT proxy

buddhisthead:
The initial PR: https://github.com/basho/riak_repl/pull/197 implemented replication of binary objects. However, after writing a replication2_upgrade test, it was discovered that the realtime queue shutdown was broken for binary objects.

The problem was two-fold:
- handoff of RT objects to a peer node (in the same cluster) did not account for the expected wire version of the receiving node. If the handoff node was 1.3, it would get a new binary object and crash on the put. For 1.4 -> 1.3 handoff, the objects are now downgraded before being pushed onto the input queue of the 1.3 handoff node.
- Objects can also be proxy'd to a remote node by the node that took over for a node that is shutting down. The node doing the proxy will send RT objects to a node on the remote (sink) cluster. The sender was not considering the wire version of the remote node.

See also https://github.com/basho/riak_test/pull/235

The replication2 test now passes. However, the replication2_upgrade test fails more often than it succeeds. This is due to a failure of the cluster manager to reconnect after it's connected peer is upgraded. I don't think it has anything to do with the upgrade, really. The test hangs while waiting for the clusters to be "connected", which never happens. This usually occurs when the second of the three remote "B" nodes is upgraded. This appears unrelated to binary objects and is almost certainly a cluster manager issue, which I will address after we ship 1.3.1.


bookshelfdave:
@buddhisthead the old PR shows that it's already been merged: https://github.com/basho/riak_repl/pull/197
Is that going to be a problem?

buddhisthead:
@metadave  Good question. I'll need to think about the best way to extract the updated part into a new branch, or something. Did you experience a problem with checking out this branch?

bookshelfdave:
I tried rebasing it so the deps that we've been fighting with would be correct:

[0]R15B01:[cet-binary-robj]slag:~/basho/riak_ee/deps/riak_repl$ git rebase master
First, rewinding head to replay your work on top of it...
Applying: First pass at making calls to new riak object accessors
Using index info to reconstruct a base tree...
Falling back to patching base and 3-way merge...
Auto-merging src/riak_repl_tcp_client.erl
CONFLICT (content): Merge conflict in src/riak_repl_tcp_client.erl
Auto-merging src/riak_repl_keylist_server.erl
CONFLICT (content): Merge conflict in src/riak_repl_keylist_server.erl
Auto-merging src/riak_repl_fullsync_worker.erl
CONFLICT (content): Merge conflict in src/riak_repl_fullsync_worker.erl
Auto-merging src/riak_repl_fullsync_helper.erl
CONFLICT (content): Merge conflict in src/riak_repl_fullsync_helper.erl
Auto-merging src/riak_repl2_rt.erl
CONFLICT (content): Merge conflict in src/riak_repl2_rt.erl
Failed to merge in the changes.
Patch failed at 0001 First pass at making calls to new riak object accessors

When you have resolved this problem run "git rebase --continue".
If you would prefer to skip this patch, instead run "git rebase --skip".
To check out the original branch and stop rebasing run "git rebase --abort".

buddhisthead:
@metadave Dave, I hope this last commit makes things easier to merge. It shows green for me. The test fails as expected with a failure of the cluster manager to connect. But that's after 2/3 nodes have been upgraded (with the replication2_upgrade test on it's own branch). So, I'd like to merge these changes once your have reviewed the code. 

At that point, I'll open a new PR for the cluster manager changes on a new branch.

src/riak_repl_migration.erl
```diff
@@ -78,26 +84,31 @@ drain_queue(false, Peer) ->
...
                 end,
              ok end),
-    drain_queue(riak_repl2_rtq:is_empty(qm), Peer);
+    drain_queue(riak_repl2_rtq:is_empty(qm), Peer, PeerWireVer);
 
-drain_queue(true, _Peer) ->
+drain_queue(true, _Peer, _Ver) ->
    done.
 
+%% filter_peers_for_capability(Peers, Cap) ->
```
bookshelfdave: is this code dead, or for future use?

bookshelfdave:
@buddhisthead I did a thorough reading and it looks good. 

Things I like: other nodes brought up during a shutdown won't be contacted (as the node list is cached in the proxy), and a failure to rpc to a node defaults to the previous object version. 

+1

buddhisthead:
Thanks, Dave.

PR #225, opened 2013-03-27
Mw 1.2.5 1.3.0 upg fix

lordnull:
Fixes Error when upgrading a cluster from 1.2.5 to 1.3 which would break ip matching due to function/module refactoring.

### DIFF
--- a/src/riak_repl_app.erl
+++ b/src/riak_repl_app.erl
@@ -4,6 +4,9 @@
 -author('Andy Gross <andy@basho.com>').
 -behaviour(application).
 -export([start/2,prep_stop/1,stop/1]).
+% deprecated: used only when on a mixed version cluster where some
+% versions < 1.3
+-export([get_matching_address/2]).
 
 -include("riak_core_connection.hrl").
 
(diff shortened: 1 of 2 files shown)


src/riak_repl_app.erl
```diff
@@ -145,11 +148,25 @@ cluster_mgr_member_fun({IP, Port}) ->
             %AddressMask = mask_address(NormIP, MyMask),
             %?TRACE(lager:notice("address mask is ~p", [AddressMask])),
             Nodes = riak_core_node_watcher:nodes(riak_kv),
-            {Results, _BadNodes} = rpc:multicall(Nodes, riak_repl2_ip,
+            {Results, BadNodes} = rpc:multicall(Nodes, riak_repl2_ip,
                 get_matching_address, [NormIP, CIDR]),
+            % when this code was written, a multicall will list the results
+            % in the same order as the nodes where tried.
+            maybe_retry_ip_rpc(Results, Nodes, BadNodes, [NormIP, CIDR]),
```
Vagabond: The result of this expression is discarded, instead of being passed to lists_shuffle(). This means that the bad data is still sent to the other cluster.

Vagabond:
replication2_upgrade 1.2.7-anya -> 1.3.1 passes, with this patch.

I've run it both in forward upgrade order and reverse upgrade order. As far as I can tell, the problem is solved, we test both new leader + old candidate(s) and old leader + new candidate(s).

+1, after squashing.

src/riak_repl_app.erl
```diff
@@ -145,11 +148,25 @@ cluster_mgr_member_fun({IP, Port}) ->
             %AddressMask = mask_address(NormIP, MyMask),
             %?TRACE(lager:notice("address mask is ~p", [AddressMask])),
             Nodes = riak_core_node_watcher:nodes(riak_kv),
-            {Results, _BadNodes} = rpc:multicall(Nodes, riak_repl2_ip,
+            {Results, BadNodes} = rpc:multicall(Nodes, riak_repl2_ip,
                 get_matching_address, [NormIP, CIDR]),
-            lists_shuffle(Results)
+            % when this code was written, a multicall will list the results
+            % in the same order as the nodes where tried.
```
Vagabond, line 154: Jon and I just read this code, and this looks like a bad assumption.

PR #226, opened 2013-03-28
Proxy get for EE 1.4

bookshelfdave:
There are other incoming PR's. I'll update the PR when complete.


bookshelfdave:
oops, need to squash

PR #227, opened 2013-03-28
Mw ensure cluster mgr leadered

lordnull:
A bandage (better than a band-aid) to the cluster manager losing it's leader. The root cause seemed to be the cluster manager being restarted due to a crash. This patch has the cluster_manager check if it's leader is defined later on in it's life, and attempt to rectify that.

### DIFF
--- a/src/riak_core_cluster_mgr.erl
+++ b/src/riak_core_cluster_mgr.erl
@@ -204,9 +204,20 @@ init([]) ->
     %% schedule a timer to poll remote clusters occasionaly
     erlang:send_after(?CLUSTER_POLLING_INTERVAL, self(), poll_clusters_timer),
     BalancerFun = fun(Addr) -> round_robin_balancer(Addr) end,
-    {ok, #state{is_leader=false,
-                balancer_fun=BalancerFun
-               }}.
+    MeNode = node(),
+    State = #state{is_leader = false, balancer_fun = BalancerFun},
+    case riak_repl2_leader:leader_node() of
+        undefined ->
+            % there's an election in progress, so we can just hang on until
+            % that finishes
+            {ok, State};
+        MeNode ->
+            State2 = become_leader(State#state{leader_node = MeNode}, MeNode),
+            {ok, State2};
+        NotMeNode ->
+            State2 = become_proxy(State#state{leader_node = NotMeNode}, NotMeNode),
+            {ok, State2}
+    end.
 
 handle_call(get_is_leader, _From, State) ->
     {reply, State#state.is_leader, State};
--- a/test/riak_core_cluster_mgr_sup_tests.erl
+++ b/test/riak_core_cluster_mgr_sup_tests.erl
@@ -1,0 +1,105 @@
+-module(riak_core_cluster_mgr_sup_tests).
+-compile(export_all).
+-include("riak_core_connection.hrl").
+-include_lib("eunit/include/eunit.hrl").
+
+murdering_test_() ->
+    {setup, fun() ->
+    % if the cluster_mgr is restarted after the leader process is running,
+    % it should ask about who the current leader is.
+        Kids = [
+            {riak_repl_leader, {riak_repl_leader, start_link, []},
+                  permanent, 5000, worker, [riak_repl_leader]},
+            {riak_repl2_leader, {riak_repl2_leader, start_link, []},
+                  permanent, 5000, worker, [riak_repl2_leader]},
+            {riak_core_cluster_mgr_sup, {riak_core_cluster_mgr_sup, start_link, []},
+                  permanent, infinity, supervisor, [riak_cluster_mgr_sup]}
+        ],
+        meck:new(riak_repl_sup, [passthrough]),
+        meck:expect(riak_repl_sup, init, fun(_Args) ->
+            {ok, {{one_for_one, 9, 10}, Kids}}
+        end),
+        meck:new(riak_core_node_watcher_events),
+        meck:expect(riak_core_node_watcher_events, add_sup_callback, fun(_fn) ->
+            ok
+        end),
+        meck:new(riak_core_node_watcher),
+        meck:expect(riak_core_node_watcher, nodes, fun(_) ->
+            [node()]
+        end),
+        application:start(ranch),
+        application:set_env(riak_repl, data_root, "."),
+        {ok, _Eventer} = riak_core_ring_events:start_link(),
+        {ok, _RingMgr} = riak_core_ring_manager:start_link(test),
+        {ok, _ClientSup} = riak_repl_client_sup:start_link(),
+        {ok, TopSup} = riak_repl_sup:start_link(),
+        riak_repl2_leader:register_notify_fun(
+            fun riak_core_cluster_mgr:set_leader/2),
+        riak_repl_leader:set_candidates([node()], []),
+        riak_repl2_leader:set_candidates([node()], []),
+        wait_for_leader(),
+        TopSup
+    end,
+    fun(_) ->
+        Mecks = [riak_repl_sup, riak_core_node_watcher_events, riak_core_node_watcher],
+        [meck:unload(Meck) || Meck <- Mecks],
+        riak_core_ring_manager:stop()
+    end,
+    fun(_) -> [
+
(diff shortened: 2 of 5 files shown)


src/riak_core_cluster_mgr.erl
```diff
@@ -389,6 +395,24 @@ handle_info(connect_to_clusters, State) ->
...
+handle_info(check_defined_leader, #state{leader_node = undefined} = State) ->
+    % if you want something done right...
+    Pid = self(),
+    case riak_repl2_leader:leader_node() of
+        undefined ->
+            lager:warning("Even repl2 leader doesn't know who the leader is"),
+            erlang:send_after(?LEADER_GOT_TIMEOUT, Pid, check_defined_leader);
+        Node ->
+            set_leader(Node, Pid), % we just discard the pid anyway.
+            erlang:send_after(?LEADER_GOT_TIMEOUT, Pid, check_defined_leader),
```
Vagabond: I propose we delete this line, it serves no apparent purpose.

src/riak_core_cluster_mgr.erl
```diff
@@ -389,6 +395,24 @@ handle_info(connect_to_clusters, State) ->
     end,
     {noreply, State};
 
+handle_info(check_defined_leader, #state{leader_node = undefined} = State) ->
+    % if you want something done right...
+    Pid = self(),
+    case riak_repl2_leader:leader_node() of
+        undefined ->
+            lager:warning("Even repl2 leader doesn't know who the leader is"),
```
Vagabond: Scary warning message that might well happen on startup is scary.

src/riak_core_cluster_mgr.erl
```diff
@@ -203,6 +204,11 @@ init([]) ->
     riak_core_service_mgr:register_service(ServiceSpec, {round_robin,?MAX_CONS}),
     %% schedule a timer to poll remote clusters occasionaly
     erlang:send_after(?CLUSTER_POLLING_INTERVAL, self(), poll_clusters_timer),
+    %% we don't know if we're started due to application start, or if we
+    %% crashed. In the case of we crashed, the leader is still going, and may
+    %% not have an election/node change for a long time. So, in that case
+    %% we need to be a bit pro-active.
+    erlang:send_after(?LEADER_GOT_TIMEOUT, self(), check_defined_leader),
```
Vagabond: You may want to make the initial message immediate, no sense in an initial 3 second delay if we really did crash.

jonmeredith:
Didn't run any code/check correctness.

1) Users are going to see the message when things go wrong.  Instead of "Even repl2 leader doesn't know who the leader is",  perhaps something like  "cluster manager node not selected, repl2 leader election in progress"

2) The current check only guards against undefined leader.  We're not certain why the leader was set undefined, if it was from a crash or a problem with the election code.  If you ended up with an incorrect election or out of date one where two nodes both thought the other was the master, I think you'd end up in an endless loop.  I would set the poll checking time to come from an appenv defaulting to 60s.

3) I was stupid in what I said about firing the check immediately on startup, or even 3s in, Andrew is right that it could well be electing still.  I think something that just polls every X seconds is the safest.

Vagabond:
The tests pass with the patch, fail without.

We should only poll for the leader on startup, once we get a leader, we should monitor() the pid, in my opinion.

lordnull:
Poll leader on startup is the prefered way, but would work best if the order of startup was reversed: ie: the leader was running before the cluster_mgr. Becuase it is not, blinding polling in the init will cause a crash, thus the send_after in init. 3 Seconds my be too short or too long:

Too short: elections haven't finished yet, so we have no leader. A warning is thrown into the logs are we poll later (a pointless exercise since the supervisor has registered us to get election changes anyway).

Too long: we've crashed and an election is a long time off. We're waiting around for 3 seconds when an immediate poll would work out for the best.

If we swap the start up order, the scary warning becomes an info: "election in progress" and we never need poll again as well will have subscribed/registered to leader election events; the other case becomes "setting leader".
Furthermore, we can do the check immediately, possibly even inside the init.

In the case of not being leader, and having a stale/invalid leader, we're kinda boned anyway. In the case described (Node1 proxies to Node2 which proxies to Node1) the calls will timeout (currently coded at 30 seconds). A more pathological case of Node1 -> Node2 -> Node3 is just as likely. I'd rather have a seperate call that indicates it's been proxies and can reply with {error, not_leader}, keeping the pathological case minimized, and alerting the caller than they have a stale leader and need to update, or wait for an election to finish.

jonmeredith:
I'm not sure swapping the startup order would provide any better guarantees.  gen_leader does not complete the election before returning the pid to the supervisor.

I think you can do 'election in progress' even without swapping the starting order, there's just a risk of an extra info message if you get the initial wait time too low.  Waiting for 3s is ok, traffic will be being buffered by the RTQ.  I'd rather that than never having a cluster manager.

Agree I'm wrong on the proxy loop.  I'd like it if it could respond {error, not_leader} but that significantly broadens the scope of the pull request, so I would say we punt on it.

Vagabond:
But sending a message to ourselves in init() is harmless, it'll just be handled immediately after init() completes, so we can initially poll as soon as init() completes.

lordnull:
The idea of swapping startup order is not to guarantee an election has completed, but to guarantee the riak_repl2_leader has started, so if we query for a leader we don't crash due to noproc. As is, when cluster_mgr is starting, we have no idea if a process we depend on running is actually up.

By having the leader started first, then having cluster mgr ask for a node we change the possible states from this:
- crashed, and leader is known: ask for leader
- crashed, and election in progress: don't ask, be told
- starting, leader hasn't started, don't ask, be told

To:
- starting, ask the leader if there's a node. If not, just wait for election to finish.

test/riak_repl_sup_tests.erl
```diff
@@ -0,0 +1,41 @@
...
+        meck:expect(riak_core_node_watcher_events, add_sup_callback, fun(_fn) ->
+            ok
+        end),
+        meck:new(riak_core_node_watcher),
+        meck:expect(riak_core_node_watcher, nodes, fun(_) ->
+            [node()]
+        end),
+
+        % needed by repl itself
+        application:start(ranch),
```
Vagabond, line 24: You don't stop ranch here...

test/riak_core_cluster_mgr_sup_tests.erl
```diff
@@ -0,0 +1,105 @@
...
+        end),
+        meck:new(riak_core_node_watcher_events),
+        meck:expect(riak_core_node_watcher_events, add_sup_callback, fun(_fn) ->
+            ok
+        end),
+        meck:new(riak_core_node_watcher),
+        meck:expect(riak_core_node_watcher, nodes, fun(_) ->
+            [node()]
+        end),
+        application:start(ranch),
```
Vagabond, line 30: Ranch also started but not stopped here...

Vagabond:
I know this isn't necessarily the fault of this pull request, but the repl test suite is broken. There's various places ranch is started without being stopped, but some places that do an ok = application:start(ranch),, There's also a typo in meck:unload:

https://github.com/basho/riak_repl/blob/b064d55388bf74e5b0ea06cf29a740ee1b36bf5e/test/riak_core_cluster_mgr_tests.erl#L368

I had to change the net_kernel:start() to use shortnames, and even with all that, one of the test suites hangs forever.

That said, I think the new tests you added are passing, sort of, but I can't tell if any old tests are breaking.

The code looks fine, and I'll grudgingly +1 it, but we really, really need to do something about our tests.

lordnull:
Seeing as one of the old tests that wasn't breaking is now breaking (riak_core_cluster_mgr_tests, correcting the typo causes a hang), I'll investigate that and try to fix it before merging.

lordnull:
Fixed the tests. While running them I failed to notice the line "some tests were cancelled" which would have clued me into the issues you saw. All tests that consistently passed before now consistently pass again, and inconsistent tests are still inconsistent (but pass often enough to be somewhat useful). There are two tests that do not pass, but are outside the scope of this patch.

lordnull:
Taking advantage of the earlier :+1: as advised by @jonmeredith .

PR #235, opened 2013-03-30
Fix repl deps for 1.3.1 release

buddhisthead:
Fix repl so that it points to riak_kv version 1.3.1
and update version in app file.


bookshelfdave:
looks good to me, +1

PR #237, opened 2013-04-02
Replace fullsync complete stat on server.

buddhisthead:
- Replace the line of code that increments the global fullsync completed stat on the server.
- Needed by the riak_test/src/repl_util.erl to detect when fullsyncs have completed.
- Maybe needed for other customers who may watch this stat.


Vagabond:
+1, although I did not actually run this code, I'm certain that it is correct.

buddhisthead:
Yeah, I ran the tests. I doesn't pass unless this code is in.

PR #163, opened 2013-04-11
NAT support for BNW

Vagabond:


Vagabond:
https://gist.github.com/Vagabond/6c9c1acaf199bd146a35

src/riak_repl2_ip.erl
```diff
@@ -221,15 +216,91 @@ find_best_ip(MyIPs, MyIP, Port, MyCIDR, SearchDepth) ->
...
+            case inet_res:gethostnames(Host) of
+                {ok, {hostent, _Domain, _, inet, 4, Addresses}} ->
+                    lager:debug("remotely resolved ~p to ~p", [Host, Addresses]),
+                    [{External, Addr} || Addr <- Addresses] ++ Acc;
+                Res2 ->
+                    lager:warning("Failed to resolve ~p locally: ~p", [Host, Res]),
+                    lager:warning("Failed to resolve ~p remotely: ~p", [Host, Res2]),
+                    Acc
+            end
+    end;
```
lordnull, line 300: These two clauses duplicate each other quite a bit, seems a good place to abstract some code out to another function.

Vagabond, line 300: In so far as they call similarly named functions with the same return signatures and then do a list comprehension on the result?

lordnull, line 300: The two clauses only differ on the list comprehension, so yes.

lordnull:
Based on the gist, I got a cluster behind nat up and running, and a cluster outside the nat able to talk to it. Correct IP/Ports were retrieved when I configured it right, so this gets a +1 from me. This also tested https://github.com/basho/riak_ee/pull/132 , giving that a +1 as well.

bookshelfdave:
rebased in https://github.com/basho/riak_repl/pull/282

PR #248, opened 2013-04-17
Handle node down correctly.

lordnull:
Remove the node entry, not the world, from state.


jonmeredith:
Turns out nodeup was broken too.  Fixed and corrected formatting.

bookshelfdave:
This file has been moved to riak_core: https://github.com/basho/riak_core/blob/master/src/riak_core_tcp_mon.erl, I'm assuming we'll have to merge the change in both places.

bookshelfdave:
Can we get an eunit test for this?

buddhisthead:
This PR has been subsumed by https://github.com/basho/riak_repl/pull/273

PR #250, opened 2013-04-18
Correcting order of clauses.

lordnull:
tcp_error and ssl_error now handled properly.


bookshelfdave:
+1, passes replication2 test

Issue #255, opened 2013-04-23
riak_core_cluster_mgr state corrupted when restarted by supervisor


Vagabond:
riak_core_cluster_mgr_sup is a rest_for_one supervisor, which means if riak_core_service_mgr or the riak_core_cluster_mgr itself crash, the riak_core_cluster_mgr will be restarted.

This is all fine and well, except that when the riak_core_cluster_mgr starts. it installs default stub resolution functions. It expects that the riak_repl application start process will register real resolution functions later. However, for a supervisor restart, this does not happen, so the stub functions remain.

This means that if you can crash the service or cluster manager, that node will no longer behave correctly. The most obvious symptom is that, when asked for a list of IPs in the cluster, it will return the empty list.

When this happens, riak-repl connect IP:Port will initially connect, but when it tries to pull the list of IPs from the other cluster to do the second phase connect (where it connects by cluster name), it will try to connect to a cluster with no known IPs.

This will then lead to 'riak-repl connections' showing no active connections, but attempts to reconnect will fail with 'already connected' since in the background we're trying to connect to this IP-less cluster.

There are various approaches to fix this, including:
- Hardcode the funs, make them pluggable later
- Add another child process to the riak_core_cluster_mgr_sup that registers the funs as part of its startup, so when the rest_for_one triggers, it is restarted as well
- Store the funs in an ETS table owned by some other process that isn't likely to restart, the riak_core_cluster_mgr_sup itself is a good candidate for this

buddhisthead:
Resolved by PR #257

PR #257, opened 2013-04-24
Add default registration functions to cluster manager supervisor

buddhisthead:
See issue https://github.com/basho/riak_repl/issues/255

This solution exports the member, save, and restore functions from riak_repl_app.erl and uses them as defaults in the cluster manager supervisor. They are also registered, as before, by riak_repl_app. If the cluster manager is terminated, the supervisor will not start it up with defaults that will allow connections to already known clusters. This solution allows the cluster manager to remain pluggable with a different supervision strategy or no supervisor.

An additional change was made to cause reconnections to be attempted after various intervals up to one minute in case the ring takes longer than 5 seconds to converge.

Tested by connecting two clusters and sequentially terminating all cluster managers on the source side, twice each, to ensure that restart is idempotent.

### DIFF
--- a/src/riak_core_cluster_mgr.erl
+++ b/src/riak_core_cluster_mgr.erl
@@ -81,6 +81,7 @@
                }).
 
 -export([start_link/0,
+         start_link/3,
          set_leader/2,
          get_leader/0,
          get_is_leader/0,
@@ -108,11 +109,13 @@
 %%%===================================================================
 %%% API
 %%%===================================================================
-
 %% @doc start the Cluster Manager
 -spec(start_link() -> ok).
 start_link() ->
-    Args = [],
+    gen_server:start_link({local, ?SERVER}, ?MODULE, [], []).
+
+start_link(DefaultLocator, DefaultSave, DefaultRestore) ->
+    Args = [DefaultLocator, DefaultSave, DefaultRestore],
     Options = [],
     gen_server:start_link({local, ?SERVER}, ?MODULE, Args, Options).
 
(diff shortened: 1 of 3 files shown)


src/riak_core_cluster_mgr.erl
```diff
@@ -193,19 +196,26 @@ set_gc_interval(Interval) ->
...
-
-init([]) ->
+init(Defaults) ->
     lager:debug("Cluster Manager: starting"),
-    register_cluster_locator(),
-    %% start our cluster_mgr service
-    ServiceProto = {?CLUSTER_PROTO_ID, [{1,0}]},
-    ServiceSpec = {ServiceProto, {?CTRL_OPTIONS, ?MODULE, ctrlService, []}},
-    riak_core_service_mgr:register_service(ServiceSpec, {round_robin,?MAX_CONS}),
+    %% start our cluster_mgr service if not already started.
```
Vagabond, line 201: Is there an issue with simply re-registering here?

src/riak_core_cluster_mgr.erl
```diff
@@ -414,6 +424,27 @@ code_change(_OldVsn, State, _Extra) ->
 %%% Private
 %%%===================================================================
 
+%% Cause ourself to try and reconnect to known clusters at various intervals.
+%% If a connection is already established, it won't create a new one.
+schedule_cluster_connections() ->
```
Vagabond, line 429: Why do we try to reconnect so many times? what happens if the network is down for 90 seconds after the crash?

buddhisthead, line 429: This is not about connections being down. This is about the ring being stable. Once a connection is started, it won't get restarted. Once it is started, the connection manager will do retries, even if the network connection is down for days.

src/riak_core_cluster_mgr.erl
buddhisthead, line 201: Yes. We don't want to re-register if the protocol id is already registered, because that will cause an error from the ranch layer. So, I think this guard is required.

Vagabond:
A handy way to test this is:

```
(riak@127.0.0.1)1> erlang:fun_info(element(5, element(2, hd(element(2, lists:nth(3, lists:nth(5, element(4, sys:get_status(whereis(riak_core_cluster_manager))))))))), module).
{module,riak_repl_app}
(riak@127.0.0.1)2> exit(whereis(riak_core_cluster_manager), kill).                                                                                                             true
(riak@127.0.0.1)3> erlang:fun_info(element(5, element(2, hd(element(2, lists:nth(3, lists:nth(5, element(4, sys:get_status(whereis(riak_core_cluster_manager))))))))), module).
{module,riak_repl_app}
```

In any case, this seems to resolve the issue. +1 to merge.

buddhisthead:
BTW, Thanks Andrew for the nice testing script, which I used to test the port to the 1.3 branch.

PR #260, opened 2013-04-29
removed ranch, it's brough in by riak_core

bookshelfdave:


buddhisthead:
+1... riak_ee builds fine with that dep removed from riak_repl

Issue #258, opened 2013-04-25
riak_repl2_rtq (possibly) consumes large memory


shino:
When site is down or not configured, riak_repl2_rtq (possibly) consumes large memory.

There is setting item for max bytes [1], but only ETS table size is checked [2].
Because Riak CS uses 1MB chunk size, riak_repl2_rtq may hold more than 100 GB memory.

[1] https://github.com/basho/riak_repl/blob/1.3/src/riak_repl2_rtq.erl#L114
[2] https://github.com/basho/riak_repl/blob/1.3/src/riak_repl2_rtq.erl#L386

Environment:
- riak_ee.git branch 1.3 (ad47b04d14cb30adcf221d285b32b0346cbc67b9)
- Ubuntu 12.10, amd64
- Erlang/OTP R15B03-1

steps to reporoduce:

```
$ make stagedev1
$ dev/dev1/bin/riak start
$ dev/dev1/bin/riak-repl add-listener dev1@127.0.0.1 127.0.0.1 11016
$ dd if=/dev/zero of=zero.1MB bs=1M count=1

$ for i in {1..300}; do
    curl -X PUT --data-binary @zero.1MB http://127.0.0.1:10018/buckets/test/keys/${i};
  done

$ dev/dev1/bin/riak attach
> [erlang:garbage_collect(P)|| P <- erlang:processes()].
> erlang:memory(binary).
314835224  %% ~ 300MB
> [erlang:external_size(M) || M <- riak_repl2_rtq:dumpq()].
[1049120,1049122,1049122,1049122,1049122,1049122,1049120,
 1049122,1049122,1049123,1049123,1049123,1049123,1049123,
 1049123,1049123,1049123,1049123,1049123,1049123,1049123,
 1049123,1049123,1049121,1049123,1049121,1049123,1049123,
 1049121|...]

> lists:sum([erlang:external_size(M) || M <- riak_repl2_rtq:dumpq()]).
314737151  %% ~ 300 MB

> ets:i().
 id              name              type  size   mem      owner
 ----------------------------------------------------------------------------
 [snip]
 983210          riak_repl2_rtq    ordered_set 300    4590     riak_repl2_rtq
 [snip]

> 4590*erlang:system_info(wordsize)/300.
122.4   %% bytes per riak_repl2_rtq ETS record
```

jonmeredith:
Thank you for the excellent demonstration Shino.  We need to address this.

bookshelfdave:
Fixed in #261. 

[...]

PR #261, opened 2013-04-29
accurately calculate RTQ size.

bookshelfdave:
See: https://github.com/basho/riak_repl/issues/258

https://gist.github.com/metadave/951fc4a5c1254e3ad144

### DIFF
--- a/src/riak_repl2_rtq.erl
+++ b/src/riak_repl2_rtq.erl
@@ -46,7 +46,11 @@
                 qseq = 0,  % Last sequence number handed out
                 max_bytes = undefined, % maximum ETS table memory usage in bytes
                 cs = [],
-                shutting_down=false}).  % Consumers
+                shutting_down=false,
+                qsize_bytes = 0
+               }).
+
+% Consumers
 -record(c, {name,      % consumer name
             aseq = 0,  % last sequence acked
             cseq = 0,  % last sequence sent
@@ -124,7 +128,7 @@ handle_call(status, _From, State = #state{qtab = QTab, max_bytes = MaxBytes,
          || #c{name = Name, aseq = ASeq, cseq = CSeq, 
                drops = Drops, errs = Errs} <- Cs],
     Status =
-        [{bytes, qbytes(QTab)},
+        [{bytes, qbytes(QTab, State)},
          {max_bytes, MaxBytes},
          {consumers, Consumers}],
     {reply, Status, State};
@@ -210,10 +214,8 @@ ack_seq(Name, Seq, State = #state{qtab = QTab, qseq = QSeq, cs = Cs}) ->
                                 end
                         end, {[], QSeq}, Cs),
     %% Remove any entries from the ETS table before MinSeq
-    cleanup(QTab, MinSeq),
-    State#state{cs = UpdCs}.
-
-
+    NewState = cleanup(QTab, MinSeq, State),
+    NewState#state{cs = UpdCs}.
 
 handle_info(_Msg, State) ->
     {noreply, State}.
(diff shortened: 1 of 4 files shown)


Vagabond:
The major problem with this approach is that it will double-count things. The original problem is that large, off-heap binaries were stored by reference in ETS, so their size is not reflected in the table size.

Doing an erlang:external_size() means that the tuple and the 2 integer values (sequence and item count) will be double-counted, as they're not stored by reference in ETS.

I think the right thing to do is to check the size of the binary, if it is larger than 64 bytes (this may always be the case, because our object overhead?), is add to the queue size the byte_size(Bin) - (6 \* wordsize). The 6 \* wordsize junk is the size of an off-heap binary reference, as determined by erts_debug:flat_size() (may not be accurate?). The other elements of the tuple, and the tuple itself, since they are stored by value in ETS, need not have their size accounted for again.

src/riak_repl2_rtq.erl
```diff
@@ -373,19 +393,18 @@ trim_q_entries(QTab, MaxBytes, Cs) ->
...
+                    trim_q_entries(QTab, MaxBytes, Cs2, NewState);
                 _ ->
-                    Cs2
+                    {Cs2, NewState}
             end
     end.
 
-qbytes(QTab) ->
+qbytes(QTab, #state{qsize_bytes = QSizeBytes}) ->
     WordSize = erlang:system_info(wordsize),
```
Vagabond: Unrelated to this PR, but why the heck do we pull this value out of system info all the time, it can not change at runtime at all. Would be better to stash in the state or something.

Vagabond:
There's also additional overhead for the term_to_binary encoding that erlang:external_size is accounting for (although it seems to be minimal):

``` erlang
18> erts_debug:flat_size([1, 1, 1]).
6
19> erlang:external_size([1, 1, 1]).
7
```

Vagabond:
More details:

``` erlang
27> Tid = ets:new(foo, []).
24592
28> ets:info(Tid, memory).
300
29> ets:insert(Tid, {foo, [1, 1, 1]}).
true
30> ets:info(Tid, memory).
313
31> erts_debug:flat_size({foo, [1, 1, 1]}).
9
32> ets:delete(Tid).
true
33> f(Tid).
ok
34> Tid = ets:new(foo, []).
28688
35> ets:insert(Tid, {foo, Bin}).
true
36> ets:info(Tid, memory).
313
37> erts_debug:flat_size({foo, Bin}).
9
38> byte_size(Bin).
9714
```

src/riak_repl2_rtq.erl
```diff
@@ -353,16 +390,19 @@ trim_q(State = #state{qtab = QTab, qseq = QSeq, max_bytes = MaxBytes}) ->
...
 
-trim_q_entries(QTab, MaxBytes, Cs) ->
+trim_q_entries(QTab, MaxBytes, Cs, State) ->
     case ets:first(QTab) of
         '$end_of_table' ->
-            Cs;
+            {Cs, State};
         TrimSeq ->
+            [QEntry] = ets:lookup(QTab, TrimSeq),
+            ObjSize = erlang:external_size(QEntry),
```
Vagabond: You forgot one external_size here.

bookshelfdave:
I accidentally pushed a version of this branch with eqc + ranch dep branches merged into it. I removed the ranch dep, but I'll leave the eqc changes.

Vagabond:
I think this needs some eunit tests or something, since I think the EQC test is currently a little too complicated to fix to properly test trimming.

Vagabond:
I'm +1 on this, if you're ok with the eunit test. The EQC test is pretty broken still, I have some fixes to the EQC test here: https://github.com/basho/riak_repl/tree/adt-rtq-eqc-fix but they are incomplete.

bookshelfdave:
eunit test works fine, I'll merge and continue to work on EQC if time permits.

PR #268, opened 2013-05-08
Added send and recv sum kpbs stats for rt and fs

lordnull:
fullsync_send_kpbs, fullsync_recv_kbps, realtime_send_kbps, and
realtime_recv_kbps are at the top level of both console and web json
stats.

fullsync stats are only useful while a fullsync is in progress; due to the
connect/disconnect nature of fullsync, it will vary widely.

### DIFF
--- a/src/riak_repl_console.erl
+++ b/src/riak_repl_console.erl
@@ -8,6 +8,7 @@
 -export([status/1, start_fullsync/1, cancel_fullsync/1,
          pause_fullsync/1, resume_fullsync/1]).
 -export([client_stats_rpc/0, server_stats_rpc/0]).
+-export([extract_rt_fs_send_recv_kbps/1]).
 
 -export([clustername/1, clusters/1,clusterstats/1,
          connect/1, disconnect/1, connections/1,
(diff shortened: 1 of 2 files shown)


bookshelfdave:
I rebased this branch against 1.3 in mw-anya-stat-patch-rebased

bookshelfdave:
with the rebased branch `mw-anya-stat-patch-rebased`, the stats work fine. Since these numbers can change quickly, it might make sense to (also?) keep a histogram, although I'm not sure if we should try to cram that into 1.3.2. 

@lordnull if you open a PR with `mw-anya-stat-patch-rebased`, I'll +1 and merge it.

jonmeredith:
The intent for this was to graph on a stats system, so the trend can go there.  Post-1.4 I think we need to reconsider how TCP mon reports this info - we're going to have more connections going through it once we move handoff/aae and possibly bulk vnodes on to it.

We'll probably need to split it into two parts - one that keeps track of open connections and octets/packets sent.  Another that worries about keeping around mini-throughput rates for commandline tools.  The heavy lifting for big clusters is going to get done by the monitoring systems, but when it all goes to crap, having that built in is very useful.

lordnull:
Look to https://github.com/basho/riak_repl/pull/274 for the merge, as it has been rebased (by @dave) and can do so cleanly. I'm going to close this / reference it in the other.

PR #269, opened 2013-05-08
Add heartbeat to the realtime replication protocol on 1.3

buddhisthead:
This change bumps the RT proto to {1, 1}. We may need to resolve
this with the binary object work and cascaded realtime work.

If both sides are using the updated protocol, a heartbeat message is
sent from the RT source every {riak_repl, rt_heartbeat_interval} which
default to 15s.  If a response is not received in
{riak_repl, rt_heartbeat_timeout}, also default to 15s then the source
connection exits and will be re-established
by the supervisor.
1. On startup/interval timer - rtsource_conn casts to rtsource_helper
   to send over the socket.  If TCP buffer is full or rtsource_helper
   is otherwise hung the rtsource_conn process will still continue.
   rtsource_conn sets up a heartbeat timeout.
2. At rtsink, on receipt of a heartbeat message it sends back
   a heartbeat message and stores the timestamp it last received one.
   The rtsink does not worry about detecting broken connections
   as new ones can be established harmlessly.  Keep it simple.
3. If rtsource receives the heartbeat back, it cancels the timer
   and updates the hearbeat round trip time (hb_rtt) then sets
   a new heartbeat_interval timer.
   
   If the heartbeat_timeout fires, the rtsource connection terminates.
   The rtsource_helper:stop call is now wrapped in a timeout in
   case it is hung so we don't get nasty messages about rtsource_conn
   crashing when it's the helper that is causing the problems.

### DIFF
--- a/src/riak_repl2_rtframe.erl
+++ b/src/riak_repl2_rtframe.erl
@@ -2,8 +2,10 @@
 -module(riak_repl2_rtframe).
 -export([encode/2, decode/1]).
 
--define(MSG_OBJECTS, 16#10). %% List of objects to write
--define(MSG_ACK,     16#20). %% Ack
+-define(MSG_HEARTBEAT, 16#00). %% Heartbeat message
+-define(MSG_OBJECTS,   16#10). %% List of objects to write
+-define(MSG_ACK,       16#20). %% Ack
+
 
 %% Build an IOlist suitable for sending over a socket
 encode(Type, Payload) ->
@@ -17,8 +19,10 @@ encode_payload(objects, {Seq, BinObjs}) when is_binary(BinObjs) ->
      BinObjs];
 encode_payload(ack, Seq) ->
     [?MSG_ACK,
-     <<Seq:64/unsigned-big-integer>>].
-    
+     <<Seq:64/unsigned-big-integer>>];
+encode_payload(heartbeat, undefined) ->
+    [?MSG_HEARTBEAT].
+
 decode(<<Size:32/unsigned-big-integer, 
          Msg:Size/binary, % MsgCode is included in size calc
          Rest/binary>>) ->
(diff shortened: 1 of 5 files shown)


buddhisthead:
From @jonmeredith...
Protocol versioning needs attention interacting with the cascading replication code, and apologies for the lack of r_t.

```
2013-05-07 11:28:47.788 [warning] <0.8967.0>@riak_repl2_rtsource_conn:handle_info:262 Realtime connection 127.0.0.1:10026 to "c2" heartbeat timeout after 15000 milliseconds
```

You can trigger the behavior by sys:suspend() the remote rtsink process.

buddhisthead:
Because this PR is adding a new protocol version "back in time", we need to jiggle some of the assignments.
Before this PR, we have...
RT Protocols:
1.0 BNW intitial (1.2.5 release)
1.1 binary objects (1.4 release)
2.0 cascading writes (1.4 release)

After this PR, we have...
RT Protocols with Heartbeats (added to release 1.3.x with plans to port to 1.4 as well)
1.0 BNW intitial (1.2.5 release)
1.1 Heartbeat (1.3 and 1.4 releases)
1.5 binary objects (1.4 release)
2.0 cascading writes (1.4 release)

So, as part of this PR, I'll add a commit that revs the binary object support to use RT protocol 1.5 instead of 1.1. This will give us additional room (versions 1.2..1.4) for any emergency protocol additions to the 1.3.x code base.

jonmeredith:
Thanks for opening - submitted the other PR against the wrong base.

lordnull:
Eunit tests that passed before still do. riak_test replication2 passes. Manual testing passes. Only thing lacking is a test to ensure a stalled heartbeat really does stop the replication (mentioned in the pr for 1.4).

buddhisthead:
Tested with older versions and still works :-)

+1 to merge, knowing that we still need a riak_test for it, which will have to wait until after code freeze today.

Issue #243, opened 2013-04-10
Don't allow bad IP's in the ring


bookshelfdave:
```
        rp(riak_repl_ring:get_clusters(element(2, riak_core_ring_manager:get_my_ring()))).
[{"nk11-staging",[]},
 {"reporting",
  [{"17.173.26.138",9085},
   {"17.173.26.141",9085},
   {"17.173.26.149",9085},
   {"17.173.26.145",9085},
   {"17.173.26.153",9085},
   {"17.173.26.148",9085},
   {"17.173.26.155",9085},
   {"17.173.26.140",9085},
   {"17.173.26.157",9085},
   {"17.173.26.150",9085},
   {"17.173.26.151",9085},
   {"17.173.26.156",9085},
   {"17.173.26.142",9085},
   {"17.173.26.143",9085},
   {"17.173.26.144",9085},
   {"17.173.26.154",9085},
   {"17.173.26.152",9085},
   {"17.173.26.147",9085},
   {"17.173.26.139",9085},
   {"17.173.26.146",9085}]},
 {"nk11-prod",
  [{badrpc,{'EXIT',{undef,[{riak_repl_app,get_matching_address,
                                          [{17,154,234,192},288442],
                                          []},
                           {rpc,'-handle_call_call/6-fun-0-',5,
                                [{file,"rpc.erl"},{line,203}]}]}}},
   {"17.154.234.192",9085},
   {"17.154.234.206",9085},
   {"17.154.234.204",9085},
   {"17.154.234.203",9085},
   {"17.154.234.200",9085},
   {"17.154.234.207",9085},
   {"17.154.234.194",9085},
   {"17.154.234.205",9085},
   {"17.154.234.201",9085},
   {"17.154.234.202",9085},
   {"17.154.234.197",9085},
   {"17.154.234.181",9085},
   {"17.154.234.176",9085},
   {"17.154.234.175",9085},
   {"17.154.234.187",9085},
   {"17.154.234.173",9085},
   {"17.154.234.199",9085},
   {"17.154.234.198",9085},
   {"17.154.234.196",9085},
   {"17.154.234.164",9085},
   {"17.154.234.163",9085},
   {"17.154.234.191",9085},
   {"17.154.234.195",9085},
   {"17.154.234.193",9085},
   {"17.154.234.171",9085},
   {"17.154.234.169",9085},
   {"17.154.234.180",9085},
   {"17.154.234.179",9085},
   {"17.154.234.185",9085},
[...]

PR #271, opened 2013-05-09
Fix issue 243 - Keep bad IPs out of ring.

buddhisthead:
Resolves issue https://github.com/basho/riak_repl/issues/243

Filters IP addresses on the receiving side of a cluster manager connection, ensuring that only data values that are parsable IP addresses are accepted and stored into the ring.

This is difficult to test, so I exported the filter function riak_core_cluster_mgr:ensure_valid_ip_addresses/1, which can then be tested from an attached console.


bookshelfdave:
+1, riak_tested w/ replication2

Issue #222, opened 2013-03-20
Fix cluster manager re-connection bug


buddhisthead:
Replication2_upgrade riak_test fails because the remote node is killed during it's upgrade. This appears to cause the cluster manager to be unable to reconnect to a different node in the remote cluster.

buddhisthead:
Fixed by @lordnull merge of 1.3.2 -> master

PR #272, opened 2013-05-09
Fixes issue 222, cluster manager can't reconnect after disconnect

buddhisthead:
This PR is against 1.3 and will later be applied to 1.4

Resolves issue 222: https://github.com/basho/riak_repl/issues/222

Previously, the cluster manager wasn't removing a connection when disconnect was called. It was dropping the connection, but it wasn't removing the entry from the ring; rather, it was writing the connection name with an empty list of IP addresses. It would never drop the name from the list of connections and it could never re-connect to a new IP addresses. Removing the name from the dictionary fixes the problem and finally makes the list of connections correct after a disconnect! Yay.


bookshelfdave:
works fine, riak_tested with replication2
+1

buddhisthead:
Thanks, Dave!

PR #273, opened 2013-05-09
Handle node down correctly

buddhisthead:
Original PR was https://github.com/basho/riak_repl/pull/248

This is a cherry pick against 1.3 because rebasing was so horrible.


buddhisthead:
Changed the unit test to do a send via ! to the tcpmon so that it would trigger the handle_info(nodedown/nodeup) clauses. Previously, it was a cast and that wouldn't exercise the new code. Confirmed with (now removed) logging in the handle_info clauses and inspection of Stat1, Stat2, and Stat3.

+1 to merge with this commit

PR #274, opened 2013-05-09
Added send and recv sum kpbs stats for rt and fs

lordnull:
fullsync_send_kpbs, fullsync_recv_kbps, realtime_send_kbps, and
realtime_recv_kbps are at the top level of both console and web json
stats.

fullsync stats are only useful while a fullsync is in progress; due to the
connect/disconnect nature of fullsync, it will vary widely.

A rebase of mw-anya-stat-patch


lordnull:
As this is a rebase, it is taking over for https://github.com/basho/riak_repl/pull/268

Look there for relevant/extended commentary.

bookshelfdave:
late +1

PR #279, opened 2013-05-14
Fix RTQ EQC test to model trimming correctly

Vagabond:
Fixed some silliness I was doing so that the test actually works right.

cc @metadave

### DIFF
--- a/test/riak_repl2_rtq_eqc.erl
+++ b/test/riak_repl2_rtq_eqc.erl
@@ -14,6 +14,7 @@
 -include_lib("eunit/include/eunit.hrl").
 -endif.
 
+-define(BINARIED_OBJ_SIZE, 39). % 39 = byte_size(term_to_binary([make_ref()])).
 
 -record(state, {rtq, %% pid of queue process
                 qseq=0, %% Queue seq number
@@ -44,7 +45,7 @@ rtq_test_() ->
 
 
 max_bytes() ->
-    ?LET(MaxBytes, nat(), {size, (MaxBytes+1) * 39}).
+    ?LET(MaxBytes, nat(), {size, (MaxBytes+1) * ?BINARIED_OBJ_SIZE}).
 
 cleanup() ->
     catch(meck:unload(riak_repl_stats)),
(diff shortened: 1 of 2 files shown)


test/riak_repl2_rtq_eqc.erl
```diff
@@ -40,10 +40,9 @@ rtq_test_() ->
 
 
 max_bytes() ->
-    ?LET(MaxBytes, nat(), (MaxBytes+1) * 512).
+    ?LET(MaxBytes, nat(), (MaxBytes+1) * 39).
```
bookshelfdave: Can you comment on why 39 was chosen?

Vagabond: It is the size of the objects we are inserting into the queue (sans ETS overhead).

Vagabond:
Ok, this is actually ready to be reviewed now.

test/riak_repl2_rtq_eqc.erl
```diff
@@ -30,33 +31,84 @@
...
 -ifdef(TEST).
 rtq_test_() ->
-    {timeout, 30,
+    {timeout, 60,
      fun() ->
                 ?assert(eqc:quickcheck(eqc:testing_time(25,
-                                                        ?MODULE:prop_main())))
+                                                        ?MODULE:prop_main()))),
+                ?assert(eqc:quickcheck(eqc:testing_time(25,
+                                                        ?MODULE:prop_parallel())))
```
lordnull: Unloading the riak_repl_stats meck at this point prevents a large output of data due to cover being unable to find the riak_repl_stats_meck source file.

test/riak_repl2_rtq_eqc.erl
```diff
@@ -30,33 +31,84 @@
...
         end
     }.
 -endif.
 
 
 max_bytes() ->
-    ?LET(MaxBytes, nat(), (MaxBytes+1) * 512).
+    ?LET(MaxBytes, nat(), {size, (MaxBytes+1) * 39}).
+
```
lordnull, line 49: Why 39?

test/riak_repl2_rtq_eqc.erl
```diff
@@ -227,43 +312,44 @@ pull(#tc{name=Name}, Q) ->
...
-        0 ->
-            0;
-        N ->
-            choose(1, N)
-    end.
+    ?LET(E, elements(C#tc.trec), element(1, E)).
+
+trim(Q, #state{max_bytes=Max}) ->
+    {_Size, NewQ} = lists:foldl(fun({Seq, NumItems, Bin}, {Size, Acc}) ->
+                case (39 + Size) > Max of
```
lordnull: second time I've seen 39, so perhaps putting it in a descriptive macro?

lordnull:
Has my +1, but because it has some changes from me, I'll merge it once a 3rd party has given it a quick eyeball.

bookshelfdave:
looks good, we'll need to get this in 1.3
+1

PR #283, opened 2013-05-16
Add AAE fullsync strategy (technology release)

buddhisthead:
This is a squash of branch pdx-aae-fullsync that owes much debt to @jtuple . Thanks, Joe!

The fullsync coordinator is unchanged, while fssource and fssink  get a new capability.
- Source and Sink exchange caps if protocol is >= 2.0.
- If awe is enabled on both sides, they will negotiate the use of AAE strategy
- aae strategy now guarded on existence of cluster aae enabled.
- strategy dynamically determined as each partition is sync'd, so a long sync's strategy can be changed while it's running. If only we had a nice command to do that. For now, you can only change the strategy by changing the app env value in riak_repl of {fullsync_strategy, aee | keylist}.

If a partition is locked by the anti entropy manager, it will be retried periodically until the lock is released, which should only be a few seconds. The 'whereis', 'busy', and reservation mechanisms all still work the same.

For the most part, the new riak_repl_aae_sink and source work identical in their usage to the keylist workers (client and server, respectively).

A PR from riak_kv is required for this PR to complle.
https://github.com/basho/riak_kv/pull/555

And a riak test...
https://github.com/basho/riak_test/pull/282

### DIFF
--- a/src/riak_repl2_fscoordinator.erl
+++ b/src/riak_repl2_fscoordinator.erl
@@ -209,7 +209,7 @@ handle_call(status, _From, State = #state{socket=Socket}) ->
     StartTime =
         case State#state.fullsync_start_time of
             undefined -> undefined;
-            N -> calendar:gregorian_seconds_to_datetime(State#state.fullsync_start_time)
+            _N -> calendar:gregorian_seconds_to_datetime(State#state.fullsync_start_time)
         end,
     SelfStats = [
         {cluster, State#state.other_cluster},
@@ -336,7 +336,7 @@ handle_cast(_Msg, State) ->
 
 %% @hidden
 handle_info({'EXIT', Pid, Cause}, State) when Cause =:= normal; Cause =:= shutdown ->
-    lager:info("fssource ~p exited normally", [Pid]),
+    lager:debug("fssource ~p exited normally", [Pid]),
     PartitionEntry = lists:keytake(Pid, 1, State#state.running_sources),
     case PartitionEntry of
         false ->
@@ -384,7 +384,7 @@ handle_info({'EXIT', Pid, Cause}, State) when Cause =:= normal; Cause =:= shutdo
     end;
 
 handle_info({'EXIT', Pid, _Cause}, State) ->
-    lager:info("fssource ~p exited abnormally", [Pid]),
+    lager:warning("fssource ~p exited abnormally", [Pid]),
     PartitionEntry = lists:keytake(Pid, 1, State#state.running_sources),
     case PartitionEntry of
         false ->
--- a/src/riak_repl2_fscoordinator.erl
+++ b/src/riak_repl2_fscoordinator.erl
@@ -336,6 +336,7 @@ handle_cast(_Msg, State) ->
 
 %% @hidden
 handle_info({'EXIT', Pid, Cause}, State) when Cause =:= normal; Cause =:= shutdown ->
+    lager:info("fssource ~p exited normally", [Pid]),
     PartitionEntry = lists:keytake(Pid, 1, State#state.running_sources),
     case PartitionEntry of
         false ->
@@ -383,7 +384,7 @@ handle_info({'EXIT', Pid, Cause}, State) when Cause =:= normal; Cause =:= shutdo
     end;
 
 handle_info({'EXIT', Pid, _Cause}, State) ->
-    lager:warning("fssource ~p exited abnormally", [Pid]),
+    lager:info("fssource ~p exited abnormally", [Pid]),
     PartitionEntry = lists:keytake(Pid, 1, State#state.running_sources),
     case PartitionEntry of
         false ->
--- a/src/riak_repl2_fssink.erl
+++ b/src/riak_repl2_fssink.erl
@@ -1,9 +1,41 @@
 -module(riak_repl2_fssink).
 -include("riak_repl.hrl").
 
+%% @doc fssink
+%%
+%% This module is responsible, at a high level, for accomplishing the full sync
+%% of a single partition. The partition to synchronize is transmitted from the
+%% connected fssource on the source cluster. The strategy for fullsync is determined
+%% dynamically as well:
+%%   keylist - used if the protocol version is < 2.0
+%%   aae     - used if the protocol version is >= 2.0 and AAE is enabled and repl_strategy=aae
+%%
+%% Protocol Support Summary:
+%% -------------------------
+%% 1,0 and up supports keylist strategy
+%% 1,1 and up supports binary object
+%% 2,0 and up supports AAE strategy
+%%
+%% For keylist, the "old" tcp_server is used. That module is capable of handling multiple
+%% sequential partitions, but we only use it to sync a single partition and then we stop it.
+%% For aae, the repl_aae_sink is desinged to process a single partition and be stopped.
+%%
+%% If the protocol version >= 2.0, and the cluster has aae enabled and the replication
+%% configuration stanza enables the aae strategy (the default is keylist), then aae strategy
+%% will be selected. It is negotiated as follows: if the version is >= 2.0, then upon
+%% accepting a connection from the soruce, we send our capabilities to the the source. The
+%% source then sends it's caps to us. This is all done by the fssource and fssink modules
+%% in order to decide which strategy will be used. Once caps are exchanged, both sides can
+%% mutually agree on the strategy (aae if both sides announced it in their caps). If the
+%% node does not have aae enabled in riak_kv or aae strategy is not enabled in riak_repl,
+%% then the caps will not include the "aae" strategy.
+%%
+%% Once the strategy is settled, the sockets are handled to the appropriate spawned
+%% fullssync worker processes.
+
 -behaviour(gen_server).
 %% API
--export([start_link/4, register_service/0, start_service/5, legacy_status/2]).
+-export([start_link/4, register_service/0, start_service/5, legacy_status/2, fullsync_complete/1]).
 
 %% gen_server callbacks
 -export([init/1, handle_call/3, handle_cast/2, handle_info/2,
@@ -14,22 +46,30 @@
         socket,
         cluster,
         fullsync_worker,
-        work_dir,
+        work_dir = undefined,
+        strategy :: keylist | aae,
+        proto,
         ver              % highest common wire protocol in common with fs source
     }).
 
 start_link(Socket, Transport, Proto, Props) ->
     gen_server:start_link(?MODULE, [Socket, Transport, Proto, Props], []).
 
+fullsync_complete(Pid) ->
+    %% cast to avoid deadlock in terminate
+    gen_server:cast(Pid, fullsync_complete).
+
 %% Register with service manager
 register_service() ->
-    %% use 1,1 proto for new binary object
-    ProtoPrefs = {fullsync,[{1,1}]},
+    %% 1,0 and up supports keylist strategy
+    %% 1,1 and up supports binary object
+    %% 2,0 and up supports AAE strategy
+    ProtoPrefs = {fullsync,[{1,1}, {2,0}]},
     TcpOptions = [{keepalive, true}, % find out if connection is dead, this end doesn't send
                   {packet, 4},
                   {active, false},
                   {nodelay, true}],
-    HostSpec = {ProtoPrefs, {TcpOptions, ?MODULE, start_service, undefined}},
+    HostSpec     = {ProtoPrefs, {TcpOptions, ?MODULE, start_service, undefined}},
     riak_core_service_mgr:register_service(HostSpec, {round_robin, undefined}).
 
 %% Callback from service manager
@@ -50,25 +90,27 @@ init([Socket, Transport, OKProto, Props]) ->
     {ok, Proto} = OKProto,
     Ver = riak_repl_util:deduce_wire_version_from_proto(Proto),
     SocketTag = riak_repl_util:generate_socket_tag("fs_sink", Transport, Socket),
+    lager:info("Negotiated ~p with ver ~p", [Proto, Ver]),
     lager:debug("Keeping stats for " ++ SocketTag),
     riak_core_tcp_mon:monitor(Socket, {?TCP_MON_FULLSYNC_APP, sink,
                                        SocketTag}, Transport),
 
     Cluster = proplists:get_value(clustername, Props),
-    lager:info("fullsync connection"),
-    {ok, WorkDir} = riak_repl_fsm_common:work_dir(Transport, Socket, Cluster),
-    %% strategy is hardcoded
-    {ok, FullsyncWorker} = riak_repl_keylist_client:start_link(Cluster,
-        Transport, Socket, WorkDir),
-    {ok, #state{cluster=Cluster, transport=Transport, socket=Socket,
-            fullsync_worker=FullsyncWorker, work_dir=WorkDir, ver=Ver}}.
+    lager:info("fullsync connection (ver ~p) from cluster ~p", [Ver, Cluster]),
+    {ok, #state{proto=Proto, socket=Socket, transport=Transport, cluster=Cluster, ver=Ver}}.
 
 handle_call(legacy_status, _From, State=#state{fullsync_worker=FSW,
-                                               socket=Socket}) ->
-    Res = case is_pid(FSW) of
-        true -> gen_fsm:sync_send_all_state_event(FSW, status, infinity);
-        false -> []
-    end,
+                                               socket=Socket, strategy=Strategy}) ->
+    Res = case is_pid(FSW) andalso is_process_alive(FSW) of
+              true ->
+                  case Strategy of
+                      keylist ->
+                          gen_fsm:sync_send_all_state_event(FSW, status);
+                      aae ->
+                          gen_server:call(FSW, status, 5000)
+                  end;
+              false -> []
+          end,
     SocketStats = riak_core_tcp_mon:socket_status(Socket),
     Desc =
         [
@@ -83,6 +125,10 @@ handle_call(legacy_status, _From, State=#state{fullsync_worker=FSW,
 handle_call(_Msg, _From, State) ->
     {reply, ok, State}.
 
+handle_cast(fullsync_complete, State) ->
+    %% sent from AAE fullsync worker
+    lager:info("Fullsync of partition complete."),
+    {stop, normal, State};
 handle_cast(_Msg, State) ->
     {noreply, State}.
 
@@ -97,6 +143,7 @@ handle_info({Error, _Socket, Reason}, State)
     {stop, normal, State};
 handle_info({Proto, Socket, Data},
         State=#state{socket=Socket,transport=Transport}) when Proto==tcp; Proto==ssl ->
+    %% aae strategy will not receive messages here
     Transport:setopts(Socket, [{active, once}]),
     case decode_obj_msg(Data) of
         {fs_diff_obj, RObj} ->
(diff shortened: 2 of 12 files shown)


src/riak_repl2_fscoordinator.erl
```diff
@@ -336,6 +336,7 @@ handle_cast(_Msg, State) ->
 
 %% @hidden
 handle_info({'EXIT', Pid, Cause}, State) when Cause =:= normal; Cause =:= shutdown ->
+    lager:info("fssource ~p exited normally", [Pid]),
```
lordnull: Shouldn't this be a debug? It's not exactly useful in production.

src/riak_repl2_fscoordinator.erl
```diff
@@ -383,7 +384,7 @@ handle_info({'EXIT', Pid, Cause}, State) when Cause =:= normal; Cause =:= shutdo
     end;
 
 handle_info({'EXIT', Pid, _Cause}, State) ->
-    lager:warning("fssource ~p exited abnormally", [Pid]),
+    lager:info("fssource ~p exited abnormally", [Pid]),
```
lordnull: Why was this down-graded from a warning?

src/riak_repl2_fssink.erl
```diff
@@ -123,16 +196,65 @@ decode_obj_msg(Data) ->
...
+terminate(_Reason, #state{fullsync_worker=FSW, work_dir=WorkDir, strategy=Strategy}) ->
+    case is_pid(FSW) andalso is_process_alive(FSW) of
         true ->
-            gen_fsm:sync_send_all_state_event(FSW, stop);
+            case Strategy of
+                keylist ->
+                    gen_fsm:sync_send_all_state_event(FSW, stop);
+                aae ->
+                    gen_server:call(FSW, stop)
+            end;
```
lordnull, line 208: At some point we should define a fullsync worker behavior so can just call a function rather than case off the strategy and do raw gen_server/gen_fsm messaging.

src/riak_repl2_fscoordinator.erl
buddhisthead: You are correct. I should be a warning. Not sure why I changed it. Probably in a flurry of debugging changes.

src/riak_repl2_fscoordinator.erl
buddhisthead: Agreed. Left over from debugging.

src/riak_repl2_fssink.erl
buddhisthead, line 208: Agreed. I considered that and just ran out of time. Calling a function would be better because the implementation (FSM or Gen Server) is exposed here and must change if the implementation of the worker changes. Gross. Sorry.

src/riak_repl2_fssource.erl
```diff
@@ -54,23 +59,41 @@ init([Partition, IP]) ->
                   {nodelay, true},
                   {packet, 4},
                   {active, false}],
-    ClientSpec = {{fullsync,[{1,1}]}, {TcpOptions, ?MODULE, self()}},
+
+    DefaultStrategy = keylist,
```
lordnull: Might make more sense for this to be a macro.

src/riak_repl2_fssource.erl
```diff
@@ -79,25 +102,64 @@ handle_call({connected, Socket, Transport, _Endpoint, Proto, Props}, _From,
...
+    %% Strategy still depends on what the sink is capable of.
+    {_Proto,{CommonMajor,_CMinor},{CommonMajor,_HMinor}} = Proto,
+
+    OurCaps = decide_our_caps(DefaultStrategy),
+    TheirCaps = maybe_exchange_caps(CommonMajor, OurCaps, Socket, Transport),
+    lager:info("Got caps: ~p", [TheirCaps]),
+    Strategy = decide_common_strategy(OurCaps, TheirCaps),
+    lager:info("Common strategy: ~p", [Strategy]),
+
     Transport:setopts(Socket, [{active, once}]),
```
lordnull: This might be better put inside the "case Strategy of" keylist clause. The socket is given to the aae worker, which means if a message came in between the above {active, once} and the give away, it will come to this process rather than the aae worker.

src/riak_repl2_fssource.erl
```diff
@@ -79,25 +102,64 @@ handle_call({connected, Socket, Transport, _Endpoint, Proto, Props}, _From,
...
+            {ok, Client} = riak:local_client(),
+            {ok, FullsyncWorker} = riak_repl_aae_source:start_link(Cluster, Client,
+                                                                   Transport, Socket,
+                                                                   Partition,
+                                                                   self()),
+            ok = Transport:controlling_process(Socket, FullsyncWorker),
+            riak_repl_aae_source:start_exchange(FullsyncWorker),
+            {reply, ok,
+             State#state{transport=Transport, socket=Socket, cluster=Cluster,
+                         fullsync_worker=FullsyncWorker, work_dir="/dev/null",
```
lordnull: Seeing /dev/null is making me cringe. A comment or something on why a flag/atom wouldn't work here?

src/riak_repl_aae_sink.erl
```diff
@@ -0,0 +1,181 @@
...
+process_msg(?MSG_GET_AAE_SEGMENT, {SegmentNum,IndexN}, State=#state{tree_pid=TreePid}) ->
+    ResponseMsg = riak_kv_index_hashtree:exchange_segment(IndexN, SegmentNum, TreePid),
+    send_reply(ResponseMsg, State);
+
+%% no reply
+process_msg(?MSG_PUT_OBJ, {fs_diff_obj, BObj}, State) ->
+    RObj = riak_repl_util:from_wire(BObj),
+    %% B = riak_object:bucket(RObj),
+    %% K = riak_object:key(RObj),
+    %% lager:info("PUT ~p:~p", [B, K]),
```
lordnull: cruft

src/riak_repl_aae_sink.erl
```diff
@@ -0,0 +1,181 @@
...
+
+%% no reply
+process_msg(?MSG_COMPLETE, State=#state{owner=Owner}) ->
+    lager:info("got complete"),
+    riak_repl2_fssink:fullsync_complete(Owner),
+    {stop, normal, State}.
+
+trace(_Msg) ->
+%%    lager:info("Sink <-------- ~p", [_Msg]),
+    ok.
```
lordnull: Bascially a no-op, making it crufty.

src/riak_repl_aae_source.erl
```diff
@@ -0,0 +1,657 @@
...
+
+    lager:info("AAE fullsync source worker started for partition ~p", [Partition]),
+
+    Timeout = app_helper:get_env(riak_kv,
+                                 anti_entropy_timeout,
+                                 ?DEFAULT_ACTION_TIMEOUT),
+
+    {ok, TreePid} = riak_kv_vnode:hashtree_pid(Partition),
+
+    %% monitor(process, Manager),
```
lordnull: cruft

src/riak_repl_aae_source.erl
```diff
@@ -0,0 +1,657 @@
...
+                   client=Client,
+                   transport=Transport,
+                   socket=Socket,
+                   index=Partition,
+                   indexns=IndexNs,
+                   tree_pid=TreePid,
+                   timeout=Timeout,
+                   built=0,
+                   owner=OwnerPid,
+                   wire_ver=w1}, %% can't be w0 because they don't do AAE
```
lordnull: A note saying this is for future proofing would be nicer. My first reaction was "if it's always the same, then there's no need to put it in the state at all".

src/riak_repl_aae_source.erl
```diff
@@ -0,0 +1,657 @@
...
+
+handle_sync_event(_Event, _From, StateName, State) ->
+    {reply, ok, StateName, State}.
+
+handle_info(Error={'DOWN', _, _, _, _}, _StateName, State) ->
+    %% Either the entropy manager, local hashtree, or remote hashtree has
+    %% exited. Stop exchange.
+    lager:info("Something went down ~p", [Error]),
+    send_complete(State),
+    {stop, something_went_down, State};
```
lordnull, line 131: Store the reference returned from erlang:monitor in the state, pattern match on that, and tell the user what broke.

src/riak_repl_aae_source.erl
```diff
@@ -0,0 +1,657 @@
...
+update_trees(cancel_fullsync, State) ->
+    larger:info("AAE fullsync source cancelled for partition ~p", [State#state.index]),
+    send_complete(State),
+    {stop, normal, State};
+update_trees(start_exchange, State=#state{indexns=IndexN, owner=Owner}) when IndexN == [] ->
+    send_complete(State),
+    lager:info("AAE fullsync source completed partition ~p. Stopping.",
+               [State#state.index]),
+    riak_repl2_fssource:fullsync_complete(Owner),
+%%    {stop, normal, State};
```
lordnull: cruft

src/riak_repl_aae_source.erl
```diff
@@ -0,0 +1,657 @@
...
+    %% that the process needing to send/receive on that socket has ownership
+    %% of it.
+    Remote = fun(init, _) ->
+                     %% cause control of the socket to be given to AAE so that
+                     %% the get_bucket and key_hashes can send messages via the
+                     %% socket (with correct ownership). We'll send a 'ready'
+                     %% back here once the socket ownership is transfered and
+                     %% we are ready to proceed with the compare.
+                     trace("Init worker_pid", [Partition]),
+                     SourcePid ! {'$aae_src', worker_pid, self()},
```
lordnull, line 269: ~~This message never seems to be handled.~~

Handled in compare_loop/1 at the end of the function. Why blocking the fsm, and not handling this in handle_info/3?

src/riak_repl_aae_source.erl
```diff
@@ -0,0 +1,657 @@
...
+                       Acc = riak_kv_index_hashtree:compare(IndexN, Remote, AccFun, TreePid),
+                       %% Maybe you wish you could move this send up to the compare function,
+                       %% but we need it to send the Accumulated results back to our SourcePid.
+                       Count = case Acc of
+                                   [] -> 0;
+                                   [N] -> N
+                               end,
+                       lager:info("Full-sync with site ~p; fullsync difference generator for ~p complete (completed in ~p secs)",
+                                  [State#state.cluster, Partition, riak_repl_util:elapsed_secs(StageStart)]),
+                       SourcePid ! {'$aae_src', done, Acc}
```
lordnull, line 315: ~~This message never seems to be handled.~~

See earlier comment.

src/riak_repl_aae_source.erl
```diff
@@ -0,0 +1,657 @@
...
+    %%     [] ->
+    %%         %% exchange_complete(LocalVN, RemoteVN, IndexN, 0),
+    %%         lager:info("Repl'd 0 keys"),
+    %%         ok;
+    %%     [Count] ->
+    %%         %% exchange_complete(LocalVN, RemoteVN, IndexN, Count),
+    %%         lager:info("Found ~b differences during AAE exchange on ~p of ~p/~p ",
+    %%                    [Count, Cluster, Partition, IndexN])
+    %% end,
+
```
lordnull: cruft

src/riak_repl_aae_source.erl
```diff
@@ -0,0 +1,657 @@
...
+accumulate_diff(KeyDiff, Bloom, [Count], #state{index=Partition}) ->
+    NumObjects =
+        case KeyDiff of
+            {remote_missing, Bin} ->
+                %% send object and related objects to remote
+                {Bucket,Key} = binary_to_term(Bin),
+                lager:debug("Keydiff: remote partition ~p remote missing: ~p:~p",
+                            [Partition, Bucket, Key]),
+                ebloom:insert(Bloom, <<Bucket/binary, Key/binary>>),
+                %%ebloom:insert(Bloom, Bin),
```
lordnull: cruft

src/riak_repl_aae_source.erl
```diff
@@ -0,0 +1,657 @@
...
+                1;
+            {different, Bin} ->
+                %% send object and related objects to remote
+                {Bucket,Key} = binary_to_term(Bin),
+                lager:debug("Keydiff: remote partition ~p different: ~p:~p",
+                            [Partition, Bucket, Key]),
+                %% ebloom:insert(Bloom, Bin),
```
lordnull: cruft

src/riak_repl_aae_source.erl
```diff
@@ -0,0 +1,657 @@
...
+    ok.
+
+%% @private
+%% do_timeout(State=#state{local=LocalVN,
+%%                         remote=RemoteVN,
+%%                         index_n=IndexN}) ->
+%%     lager:info("Timeout during exchange between (local) ~p and (remote) ~p, "
+%%                "(preflist) ~p", [LocalVN, RemoteVN, IndexN]),
+%%     send_exchange_status({timeout, RemoteVN, IndexN}, State),
+%%     {stop, normal, State}.
```
lordnull: cruft

src/riak_repl_aae_source.erl
```diff
@@ -0,0 +1,657 @@
...
+%% @private
+%% next_state_with_timeout(StateName, State) ->
+%%     next_state_with_timeout(StateName, State, State#state.timeout).
+%% next_state_with_timeout(StateName, State, Timeout) ->
+%%     {next_state, StateName, State, Timeout}.
+
+%% exchange_complete({LocalIdx, _}, {RemoteIdx, RemoteNode}, IndexN, Repaired) ->
+%%     riak_kv_entropy_info:exchange_complete(LocalIdx, RemoteIdx, IndexN, Repaired),
+%%     rpc:call(RemoteNode, riak_kv_entropy_info, exchange_complete,
+%%              [RemoteIdx, LocalIdx, IndexN, Repaired]).
```
lordnull: cruft

src/riak_repl_aae_source.erl
```diff
@@ -0,0 +1,657 @@
...
+%%                         index_n=IndexN}) ->
+%%     lager:info("Timeout during exchange between (local) ~p and (remote) ~p, "
+%%                "(preflist) ~p", [LocalVN, RemoteVN, IndexN]),
+%%     send_exchange_status({timeout, RemoteVN, IndexN}, State),
+%%     {stop, normal, State}.
+
+%% @private
+send_exchange_status(Status, State) ->
+    throw(Status),
+    State.
```
lordnull: send_exhange_status always throws an exception? Why?

src/riak_repl_aae_source.erl
```diff
@@ -0,0 +1,657 @@
...
+%% send a tagged message with type and msg. return the reply
+send_asynchronous_msg(MsgType, Msg, State) ->
+    Data = term_to_binary(Msg),
+    send_asynchronous_msg(MsgType, Data, State).
+
+%% send a message with type tag only, no data
+send_asynchronous_msg(MsgType, #state{transport=Transport,
+                                      socket=Socket}) ->
+    ok = Transport:send(Socket, <<MsgType:8>>).
+
```
lordnull, line 575: while waiting for a reply, any status request will be blocked, as well as cancel messages. It may be better to use gen_tcp:recv/3 (with the timeout) for the blocking receive.

src/riak_repl_aae_source.erl
```diff
@@ -0,0 +1,657 @@
...
+    receive
+        {_, Socket, [?MSG_REPLY|Data]} ->
+            trace_recv("MSG_REPLY"),
+            binary_to_term(Data);
+        {Error, Socket} ->
+            throw(Error);
+        {Error, Socket, Reason} ->
+            throw({Error, Reason})
+    end.
+
```
lordnull: trace/1 and trace/2 are basically no-ops. lager:debug/n should likely be used instead.

src/riak_repl_util.erl
```diff
@@ -767,7 +767,7 @@ make_pg_name(Remote) ->
 
 deduce_wire_version_from_proto({_Proto,{CommonMajor,CMinor},{CommonMajor,HMinor}}) ->
     %% if common protocols are both >= 1.1, then we know the new binary wire protocol
-    case CommonMajor >= 1 andalso CMinor >= 1 andalso HMinor >= 1 of
+    case ((CommonMajor >= 2) or (CommonMajor == 1 andalso CMinor >= 1 andalso HMinor >= 1)) of
```
lordnull: orelse is the short-circuit operator, which would mean the CommonMajor == 1 can be removed.

src/riak_repl2_fssource.erl
buddhisthead: But don't we want the {active, once} to take affect for both keylist and aae? I am missing something here.

src/riak_repl2_fssource.erl
buddhisthead: Yep. I changed aae's case to use work_dir=undefined and only rm workdir if != undefined on cleanup.

src/riak_repl_aae_source.erl
buddhisthead: added note re: possible future wire formats

src/riak_repl_aae_source.erl
buddhisthead, line 131: Can you refer me to a similar pattern in some other code? I'm not sure what you mean, but it sounds good.

src/riak_repl_aae_source.erl
buddhisthead, line 269: If I understand you correctly, you are suggesting that we instead store Bloom in the state and enter a new state that allows the compare function to run asynchronously, and handle the "done" message with handle_info; and then continue with the process of sending the differences?

This seems like a good idea because it would allow the gen_fsm to return status information during the compare cycle, instead of being blocked.

src/riak_repl_aae_source.erl
buddhisthead: Converted all calls to just throw(xxx). But does a throw cause execution to stop there? Will the next line even be executed?

src/riak_repl_util.erl
buddhisthead: I agree that it could be written as ((CommonMajor >= 2) orelse (CMinor >= 1 andalso HMinor >= 1)), but would that make it easier to read? It implies that CommonMajor == 1, but it requires thinking to see it. I like the explicit version better, but that's just my opinion.

src/riak_repl_aae_source.erl
lordnull, line 131: riak_repl2_leader uses the described pattern to monitor the leader.

Here's some code for the general pattern as well:

``` erlang
handle_cast({monitor_one, OtherPid}, {_One, Two}) ->
    MonRef = erlang:monitor(process, OtherPid),
    {noreply, {MonRef, Two}};

handle_cast({monitor_two, OtherPid}, {One, _Two}) ->
    MonRef = erlang:monitor(process, OtherPid),
    {noreply, {One, MonRef}}.

handle_info({'DOWN', MonOne, process, Pid, Why}, {MonOne, Two}) ->
    io:formar("Monitored pid one ~p went down due to ~p", [Pid, Why]),
    {noreply, {undefined, Two}};

handle_info({'DOWN', MonTwo, process, Pid, Why}, {One, MonTwo}) ->
    io:format("Monitored pid two ~p went down due to ~p", [Pid, Why]),
    {noreply, {One, undefined}}.
```

src/riak_repl2_fssource.erl
lordnull: The key difference is in the case of aae strategy, you're giving controll of the socket away, while in keylist you are not. Socket messages when using {active, once} and {active, true} always go to the controlling process. For keylist, that is the same process, but for aae, it is not; thus the potential race condition.

src/riak_repl_util.erl
lordnull: We could get even crazier and test equivalence directly :) :

``` erlang
deduce_wire_version_from_proto({_Proto,ClientVer,HostVer}) ->
    if
        HostVer >= {1,1} andalso ClientVer >= {1,1} ->
```

My main point was suggesting the short-circuit operator so unneeded work can be skipped.

src/riak_repl_aae_source.erl
lordnull: throw/1 halts execution and exits the process with the reason passed in, so the next line is never reached.

http://erlang.org/doc/man/erlang.html#throw-1

```
1> F = fun(A) -> throw(A), io:format("reached it!~n") end.
#Fun<erl_eval.6.82930912>
2> self().
<0.31.0>
3> F(test).
** exception throw: test
4> self().
<0.35.0>
5> catch F(test).
test
6> self().
<0.35.0>
```

src/riak_repl_aae_source.erl
lordnull, line 269: Yes, that sums up what I'm saying.

src/riak_repl2_fssink.erl
buddhisthead, line 208: Added TODO documentation there to add behavior.

src/riak_repl_aae_source.erl
buddhisthead, line 575: If I change the code to use a gen_tcp:recv, then I also need to make it use passive mode. And what if it does time out, then what? Either way, we'd be blocking a status call until the timeout is reached. To do this, we'd really have to change how all the messages are threaded and do async io, yes?

lordnull, line 575: To do a recv, you would have to put it in passive mode. As for what to do if it times out, either loop around again by sending a 'retry' message to self, or do as the code currently does, and just throw an error. This means the minimum amount of change needed (it remains contained in the function below).

A better way is to do async io, as it would not block status or cancel. Since there are certain messages you expect from the socket, a flag in the state could be used to suss out if it should be handled as a reply, or other data.

buddhisthead, line 575: Alright, I added a TODO to use async io; and updated the code to use passive mode with a 60 second timeout. I like this change.

src/riak_repl_util.erl
buddhisthead: Oh, this version is much better. I'm going with that.

buddhisthead:
@lordnull There is a problem with the wire format version code. RT and FS are mixing their versions badly. I need to resolve this before we can finish this PR.

buddhisthead:
@lordnull Ok, I believe I have this back in shape for a final look over. It's rebased against master so it will easy-merge.

lordnull:
I fixed a missed bit from the merge; after that the r_t for this feature passes. Other concerns were addressed, so :+1:

buddhisthead:
Many, many thanks to you for the excellent code review, Micah (@lordnull).

PR #288, opened 2013-05-23
Fixed function clause error when migrating queue

lordnull:
Discovered during a riak test, where shutting down a node would hang
forever chewing up a significant amount of cpu.

The queue migration delivery fun expected a tuple of size 3, but was given
a tuple of size 4 (due to the meta data). Fixed the clause and added a down
cast to push a size 3 to the remote rtq in case it's an older version.


buddhisthead:
I did a code review and it looks clean to me.

bookshelfdave:
+1, replication2 riak_test passes

PR #285, opened 2013-05-16
add proxy_get code missed in rebase

bookshelfdave:
I diffed dip_pg14 & master and I believe everything is accounted for. This PR is **_REQUIRED**_ for 1.4

PR #289, opened 2013-05-24
race conds with cross-node calls and node shutdown

bookshelfdave:
don't cache cluster_id in cluster_mgr

squashed version of https://github.com/basho/riak_repl/pull/285


Vagabond:
Have a well deserved +1, sir.

PR #291, opened 2013-05-31
Teach fullsync to recover from AAE not_responsible

buddhisthead:
AAE fullsync can fail with not_responsible during the compare phase when the n_val for a bucket differs between source and sink clusters. This PR allows the full sync source to recover from that error and restart with the keylist strategy.

riak_test:
See https://github.com/basho/riak_test/pull/289


src/riak_repl_aae_source.erl
```diff
@@ -225,7 +225,7 @@ update_trees(start_exchange, State=#state{tree_pid=TreePid,
 update_trees({not_responsible, Partition, IndexN}, State) ->
     lager:info("VNode ~p does not cover preflist ~p", [Partition, IndexN]),
     send_complete(State),
-    throw({not_responsible, Partition, IndexN}),
+%%    throw({not_responsible, Partition, IndexN}),
```
coderoshi, line 228: :ghost:

PR #292, opened 2013-06-03
Don't crash if the block_provider/requester is not running

Vagabond:
During rolling upgrade to 1.4, this will possibly not be the case.


bookshelfdave:
+1 multiple riak_test runs against replication2 pass

PR #294, opened 2013-06-04
Fix various problems in the eunit suite

Vagabond:
The test suite passes consistently for me now, 15 times in a row.


src/riak_repl_stats.erl
```diff
@@ -396,35 +396,43 @@ test_populate_stats() ->
...
+        [{K1, V} || {K1, V} <- Result, {K2, _} <- Expected, K1 == K2]).
+
 
 test_report() ->
     Bytes = 1024 * 60,
     Now = 10000,
     ReportMinutes = 10,
     for_n_minutes(ReportMinutes, Now, Bytes),
+    %% I think the cache is getting in the way here?
+    timer:sleep(1000),
```
bookshelfdave, line 435: would some type of wait() be more appropriate here?

bookshelfdave:
Is this output expected: 
https://gist.github.com/metadave/3f8f5bae407ed061fc3a
and after merging your branch to master (locally):
https://gist.github.com/metadave/233125176758cb67c1fd

bookshelfdave:
can anything be done about the huge binaries that meck spits out upon shutdown?

bookshelfdave:
With @lordnull's latest patch: `All 140 tests passed.`

Issue #298, opened 2013-06-07
gen_server call review


slfritchie:
From a Riak 1.4.0 prerelease package:

```
16:16:48.877 [error] gen_server riak_snmp_stat_poller terminated with reason: {timeout,{gen_server,call,[riak_core_cluster_manager,get_known_clusters]}}
16:16:48.877 [error] CRASH REPORT Process riak_snmp_stat_poller with 0 neighbours exited with reason: {timeout,{gen_server,call,[riak_core_cluster_manager,get_known_clusters]}} in gen_server:terminate/6 line 747
16:16:48.877 [error] Supervisor riak_snmp_sup had child riak_snmp_stat_poller started with riak_snmp_stat_poller:start_link() at <0.1341.0> exit with reason {timeout,{gen_server,call,[riak_core_cluster_manager,get_known_clusters]}} in context child_terminated
```

slfritchie:
PR #299 closed: Merged with commit https://github.com/basho/riak_repl/commit/66c343dacd6abdfba5f33ce8a4382f610062a5f9

PR #299, opened 2013-06-07
gen_server:call timeout review & fixes

slfritchie:
Addresses #298 

In general, I used `?LONG_TIMEOUT` whenever the callee was not an atom and therefore not 100% clear that the callee is on the local node.  A more knowledgable reviewer should take a careful look at those cases and, when you _do_ know that the call is staying on the same node, change the timeout to `infinity`.

### DIFF
--- a/include/riak_repl.hrl
+++ b/include/riak_repl.hrl
@@ -99,3 +99,5 @@
 -define(REPL_MODES, [{mode_repl12,?REPL_HOOK12}, {mode_repl13,?REPL_HOOK_BNW}]).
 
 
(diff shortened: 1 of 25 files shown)


buddhisthead:
Will try and push this into 1.4 after running riak_tests

buddhisthead:
+1 with my changes

riak_test(s) passed = replication2, replication2_upgrade, repl_aae_fullsync

slfritchie:
Merged with commit https://github.com/basho/riak_repl/commit/66c343dacd6abdfba5f33ce8a4382f610062a5f9, not sure why it didn't autoclose.

PR #303, opened 2013-06-12
Make the interval the rtsink rechecks the active flag configurable.

bowrocker:
Addresses https://github.com/basho/riak_repl/issues/252 by adding app.config riak_repl, reactivate_socket_interval configuration to schedule_reactivate_socket/1. Added test for helper function.

### DIFF
--- a/src/riak_repl2_rtsink_conn.erl
+++ b/src/riak_repl2_rtsink_conn.erl
@@ -12,6 +12,10 @@
 %% API
 -include("riak_repl.hrl").
 
+-ifdef(TEST).
+-include_lib("eunit/include/eunit.hrl").
+-endif.
+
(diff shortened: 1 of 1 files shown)


src/riak_repl2_rtsink_conn.erl
```diff
@@ -23,6 +27,9 @@
 -export([init/1, handle_call/3, handle_cast/2, handle_info/2,
          terminate/2, code_change/3]).
 
+% how long to wait to reschedule socket reactivate.
+-define(REACTIVATE_SOCK_INT, 10).
```
bookshelfdave: it might be nice to have the uom in the constant: 
REACTIVATE_SOCK_INT_MILLIS

PR #311, opened 2013-06-14
Mw eunit fix redux

lordnull:
Resolves a few issues left over from andrew's eunit fixes. There remains a single test that will occasionally fail (repl_leader_eqc), but it passes often enough to remain useful.


buddhisthead:
+1 passed 15x

Issue #247, opened 2013-04-17
rtsource/rtsink_conn peername should be stashed in the state so we can give nicer error messages


bookshelfdave:

bowrocker:
Managed to manually reproduce by:
1. Firewall-ing the cluster-mgr port (10016) from the source
2. This forces a heartbeat timeout on the source end
3. Inserting a inet:close(Socket) before the error logging
4. The logger calls inet:peername() which fails, giving:

erlang.log.2:15:34:04.854 [warning] Realtime connection error:einval to "ec21" heartbeat timeout after 15000 milliseconds

Above, "error:einval" is the error return of inet:peername() when the socket is closed.

The fix is, as the subject reads, to put the peername in the state when the socket is established, so that the logging can still have access to it. 

The problem remains how to test this condition, as just firewalling the port leaves the local (source) socket open.

PR #313, opened 2013-06-19
Cache the inet:peername in the State record in rtsink & rtsource connections for better error condition logging.

bowrocker:
Cache peername in State instead of computing from the socket. If the socket is closed, "error:einval" is returned by peername() instead of the peer name. This will allow the error logging to print the peername even if the local socket is closed.

### DIFF
--- a/src/riak_repl2_rtsink_conn.erl
+++ b/src/riak_repl2_rtsink_conn.erl
@@ -12,6 +12,10 @@
 %% API
 -include("riak_repl.hrl").
 
+-ifdef(TEST).
+-include_lib("eunit/include/eunit.hrl").
+-endif.
+
 -export([register_service/0, start_service/5]).
 -export([start_link/2,
          stop/1,
@@ -27,6 +31,7 @@
                 transport,        %% Module for sending
                 socket,           %% Socket
                 proto,            %% Protocol version negotiated
+                peername,         %% peername of socket
                 ver,              %% wire format agreed with rt source
                 max_pending,      %% Maximum number of operations
                 active = true,    %% If socket is set active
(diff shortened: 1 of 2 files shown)


lordnull:
I have concerns about how this is being tested. Testing how the internal state of a gen_server and calling the handle_\* functions directly makes for very brittle tests, and decreases the ease of refactoring what is being tested.

The problem is not "we fail to cache to the peername", the problem is when getting stats after the socket is closed, the peername is no longer available. Set up the test checking peername is available after a socket is closed, rather than the specific solution (caching the peername), and you'll have a better, more long-lived test.

bowrocker:
Thanks for the feedback, Micah.

Sorry I am so long in responding, I failed to watch my own issue, so I missed your comment.

Testing this one seemed tricky. Your point about "Set up the test checking peername is available after a socket is closed" is well taken, but my intention was do exactly that, in the test. cache_peername_test_case() opens a socket, sets it, and then closes the socket and validates that the cached peername is still available:

``` erlang
%% Now close the socket
inet:close(Socket),

%% Now make sure we still get the peername, after socket close
?assertEqual(reply, Reply),
?assertEqual(ok, Res),
?assertEqual({?LOOPBACK_TEST_PEER, ?LOOPBACK_TEST_SOCK}, peername(State)),
```

Am I missing something?

lordnull:
It's a long one, sorry.

In general, it is best to write tests that only use the exported functions excluding the behavior functions (such as handle_call and handle_cast). However, the nature of the problem makes this difficult at best.

To sum up from the issue, we're not getting a peername in the log messages after the socket is closed. We could try setting up an interception and test all the code paths to get to those messages and ensure they have a peername, but that ties the tests very tightly to the internal state and flow, make them very brittle.

So the basic idea you have is right: ensure the internal peername function has the correct data. The diffuculty there is you need the state to test that. sys:get_status/1 (sys:get_state/1 once we move to r16b01) can be used (it comes with a nice warning that's it's intended for debugging).

So given all of that, for the source I'd suggest a flow of:

```
Start a source using start_link/1
connect a tcp socket to a listening socket
call connected/6
set [{active, false}] on the socket
close the socket
get the state
use peerneme/1 and ensure it is what you expect
```

This means that instead of simply testing that a manufactured state fed to peername/1 returns the expected information, the test ensures the peername is set when the socket is set. Setting the socket to [{active, false}] before closing it prevents a close message from being sent to the source process.

The sink can use a similar strategy. (In fact the sink test doesn't actually test much as the socket is closed after the test is run, so it would still pass even if the change was reversed).

```
Start a sink
correct a tcp socket to a listening socket, and use the listening accepted socket
call start_service/5
set [{active, false}] on the socket
close the socket
get the state
use peername/1 and ensure it returns what you expect
```

While is doesn't test the log messages, it does test a fundamental aspect to making it work: that the peername is cached using normal code flow.

PR #315, opened 2013-06-21
cleanup bad return value for an undefined leader

bookshelfdave:


lordnull:
Compiles, code understandable, and all important tests pass. :+1:

PR #321, opened 2013-07-03
removed delayed_write option to file:open

bookshelfdave:
use of delayed_write has been resolved in R16B01, however we're still using R15. We're hoping that this may be the cause of `premature_eof` errors that we've seen with some customers.

See also:
- http://www.erlang.org/download/otp_src_R16B01.readme
- https://github.com/erlang/otp/commit/a73414d2e8ad03538b28546756e7e7264654466d

```
OTP-10984  Fixed a race condition when using delayed_write when writing
          to a file which would cause the same data to be written
          multiple times.
```


lordnull:
Quick eyeballing says it's good.

PR #325, opened 2013-07-08
Cache the inet:peername for better error condition logging.

bowrocker:
Cache the inet:peername in State in both rtsource & rtsink connections instead of computing from the socket. If the socket is closed, "error:einval" is returned by the local function instead of the previously connected peer name -- this means users will not know which connection was closed. This change will allow the error logging to print the peername even if the local socket is closed.

Updated after feedback from Micah Warren. Changes include using only public interfaces, with the exception of the use of sys:get_status to look at the State variable when replication is running, in order to verify. This understood to be brittle, but cannot be helped in this particular instance:

https://github.com/basho/riak_repl/pull/313

### DIFF
--- a/src/riak_repl2_rtsink_conn.erl
+++ b/src/riak_repl2_rtsink_conn.erl
@@ -12,6 +12,10 @@
 %% API
 -include("riak_repl.hrl").
 
+-ifdef(TEST).
+-include_lib("eunit/include/eunit.hrl").
+-endif.
+
 -export([register_service/0, start_service/5]).
 -export([start_link/2,
          stop/1,
@@ -27,6 +31,7 @@
                 transport,        %% Module for sending
                 socket,           %% Socket
                 proto,            %% Protocol version negotiated
+                peername,         %% peername of socket
                 ver,              %% wire format agreed with rt source
                 max_pending,      %% Maximum number of operations
                 active = true,    %% If socket is set active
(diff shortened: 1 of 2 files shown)


src/riak_repl2_rtsink_conn.erl
```diff
@@ -89,7 +94,7 @@ init([OkProto, Remote]) ->
     %% TODO: remove annoying 'ok' from service mgr proto
     {ok, Proto} = OkProto,
     Ver = riak_repl_util:deduce_wire_version_from_proto(Proto),
-    lager:debug("RT sink connection negotiated ~p wire format from proto ~p", [Ver, Proto]),
+    lager:error("RT sink connection negotiated ~p wire format from proto ~p", [Ver, Proto]),
```
bookshelfdave, line 97: this should be set back to lager:debug

src/riak_repl2_rtsink_conn.erl
```diff
@@ -361,4 +371,120 @@ schedule_reactivate_socket(State = #state{transport = T,
...
+
+    catch(meck:unload(riak_core_service_mgr)),
+    meck:new(riak_core_service_mgr, [passthrough]),
+    meck:expect(riak_core_service_mgr, register_service, fun(HostSpec, _Strategy) ->     
+        {_Proto, {TcpOpts, _Module, _StartCB, _CBArg}} = HostSpec,
+        {ok, Listen} = gen_tcp:listen(?SINK_PORT, [binary, {reuseaddr, true} | TcpOpts]),
+        TellMe ! sink_listening,
+        {ok, Socket} = gen_tcp:accept(Listen),
+        {ok, Pid} = riak_repl2_rtsink_conn:start_link({ok, ?PROTOCOL(?VER1)}, "source_cluster"),
+        %unlink(Pid),
```
bookshelfdave, line 424: cleanup

src/riak_repl2_rtsink_conn.erl
```diff
@@ -361,4 +371,120 @@ schedule_reactivate_socket(State = #state{transport = T,
...
+        % check to make sure peername is cached, not calculated from (now closed) Socket
+        case peername(State) of 
+           {Peer, _Port} ->
+               ?assertEqual(?LOOPBACK_TEST_PEER, Peer),
+               ?debugMsg("test passed");
+           _ ->
+               ?assertError(test_failed, "peer not cached")
+        end,
+        
+        %?assertEqual("error:einval", peername(inet, Socket)),
```
bookshelfdave, line 443: cleanup

bookshelfdave:
+1 from me if you fix the lager:debug + cleanup extra whitespace, however, @lordnull should give his +1 on this too.

PR #326, opened 2013-07-08
Cache the inet:peername for better error condition logging. 

bowrocker:
Cache the inet:peername in State in both rtsource & rtsink connections instead of computing from the socket. If the socket is closed, "error:einval" is returned by the local function instead of the previously connected peer name -- this means users will not know which connection was closed. This change will allow the error logging to print the peername even if the local socket is closed.

Updated after feedback from Micah Warren. Changes include using only public interfaces, with the exception of the use of sys:get_status to look at the State variable when replication is running, in order to verify. This understood to be brittle, but cannot be helped in this particular instance:

https://github.com/basho/riak_repl/pull/313

### DIFF
--- a/src/riak_repl2_rtsink_conn.erl
+++ b/src/riak_repl2_rtsink_conn.erl
@@ -401,9 +401,17 @@ setup() ->
     {ok, _} = riak_repl2_rtq:start_link(),
     riak_repl_test_util:kill_and_wait(riak_core_tcp_mon),
     {ok, _TCPMon} = riak_core_tcp_mon:start_link(),
-
     ok.
+
 cleanup(_Ctx) ->
+    riak_repl_test_util:kill_and_wait(riak_core_tcp_mon),
+    riak_repl_test_util:kill_and_wait(riak_repl2_rtq),
+    riak_repl_test_util:kill_and_wait(riak_repl2_rt),
+    riak_repl_test_util:stop_test_ring(),
+    riak_repl_test_util:maybe_unload_mecks(
+      [riak_core_service_mgr,
+       riak_core_connection_mgr,
+       gen_tcp]),
     ok.
 
 %% test for https://github.com/basho/riak_repl/issues/247
@@ -432,13 +440,7 @@ cache_peername_test_case() ->
         {status,Pid,_,[_,_,_,_,[_,_,{data,[{_,State}]}]]} = sys:get_status(Pid),
 
         % check to make sure peername is cached, not calculated from (now closed) Socket
-        case peername(State) of 
-           {Peer, _Port} ->
-               ?assertEqual(?LOOPBACK_TEST_PEER, Peer),
-               ?debugMsg("test passed");
-           _ ->
-               ?assertError(test_failed, "peer not cached")
-        end,
+        ?assertMatch({?LOOPBACK_TEST_PEER, _Port}, peername(State)),
         
         TellMe ! {sink_started, Pid}
     end),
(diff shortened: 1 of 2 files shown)


src/riak_repl2_rtsink_conn.erl
```diff
@@ -361,4 +371,117 @@ schedule_reactivate_socket(State = #state{transport = T,
...
+
+        % check to make sure peername is cached, not calculated from (now closed) Socket
+        case peername(State) of 
+           {Peer, _Port} ->
+               ?assertEqual(?LOOPBACK_TEST_PEER, Peer),
+               ?debugMsg("test passed");
+           _ ->
+               ?assertError(test_failed, "peer not cached")
+        end,
+
```
lordnull, line 444: The entire case statement can be squished down into a single assertMatch:

``` erlang
?assertMatch({?LOOPBACK_TEST_PEER, _Port}, peername(State)),
```

This way the error reported is at the point of failure (unexpected return from the peername function) rather than an unrelated test ("peer not cached" as a statement not throwing an error).

src/riak_repl2_rtsink_conn.erl
```diff
@@ -361,4 +371,117 @@ schedule_reactivate_socket(State = #state{transport = T,
...
+    riak_repl_test_util:kill_and_wait(riak_repl2_rt),
+    {ok, _RT} = riak_repl2_rt:start_link(),
+    riak_repl_test_util:kill_and_wait(riak_repl2_rtq),
+    {ok, _} = riak_repl2_rtq:start_link(),
+    riak_repl_test_util:kill_and_wait(riak_core_tcp_mon),
+    {ok, _TCPMon} = riak_core_tcp_mon:start_link(),
+
+    ok.
+cleanup(_Ctx) ->
+    ok.
```
lordnull, line 415: All those thing started in the setup should be stopped / closed / destroyed or otherwise cleaned up in the cleanup.

src/riak_repl2_rtsource_conn.erl
```diff
@@ -371,3 +382,88 @@ safe_now_diff({_,_,_} = Sooner, {_,_,_} = Later) ->
...
+setup() ->
+    process_flag(trap_exit, true),
+    riak_repl_test_util:stop_test_ring(),
+    riak_repl_test_util:start_test_ring(),
+    riak_repl_test_util:abstract_gen_tcp(),
+    riak_repl_test_util:abstract_stats(),
+    riak_repl_test_util:abstract_stateful(),
+    ok.
+cleanup(_Ctx) ->
+    ok.
```
lordnull, line 419: Ensure that which was started gets cleaned up here; there's a very full setup function.

lordnull:
Tests do pass when run on their own. Currently testing to ensure other tests are not effected...

lordnull:
Does not interfere with the rtsink and rtsource eqc tests.

src/riak_repl2_rtsink_conn.erl
bowrocker, line 444: Thanks Micah, this is great feedback. I was looking at assertMatch to use instead, but the doc was a little opaque.

src/riak_repl2_rtsink_conn.erl
bowrocker, line 415: It is not always clear what to do. Building up and tearing down the REPL "stack" does not appear to be a very well defined or documented process, unless I am missing something. Is there a good example of doing this?

I think this area might need some re-structuring at some point, so it's very clear how to do it.

bookshelfdave, line 415: I agree with @bowrocker, it would be nice to have a template and/or doc on the wiki that shows how to setup a repl test like this.

bowrocker:
Added clean up and changed the test to use assertMatch.

lordnull:
Tests pass and code looks viable :+1:

Given these tests use a similar setup/tear down as the eqc tests, we should look into unifying that code. That seems a task for another pull request.

Issue #328, opened 2013-07-10
Integrate repl hook to avoid synchronizing expired objects


bookshelfdave:
Joe Caswell has worked up a workaround for Netic to avoid synchronizing expired objects. https://gist.github.com/joecaswell/5959715

cmeiklejohn:
Did this get deployed at Netic?

cmeiklejohn:
Moving to 2.1 milestone.

engelsanchez:
If anybody has information on this, let me know. If you see objects that have expired in Bitcask, it's because expiration happened between the time it was fetched and the time it is processed. We have the basic problem of expiration happening at the backend level using the node's local time. Replica repair is unaware of differences between nodes and could possibly resurrect expired objects when there is enough drift to catch them. This hook solution doesn't seem robust enough to implement. I'd rather leave this until a robust Riak wide solution to expiration is developed.  Closing for now. Feel free to add any additional information that may lead to a different conclusion.

engelsanchez:
Actually, leaving duplicate issue #328 open for now and consulting with Mr. Chief Architect about this.

PR #329, opened 2013-07-15
Integrate repl hook to avoid synchronizing expired objects 

bowrocker:
## Overview

The riak_repl2_fullsync_expiry_hook module is a repl_helper hook function that operates on the sending side of a fullsync operation to cancel sending keys whose age, according to its internal last updated metadata, exceeds the configured expiry setting for its storage backend.

Code contributed by Joe Caswell.
### Configuration & Configuration

At system start, the gen_server will read the backend configuration from the application environment and identify any backends that are configured for key expiry. If no expiry backends are found, the gen_server terminates. If any expiry backends are found, the gen_server stores the expiry times in its internal state and registers the module as a repl_helper with riak_core.

During fullsync, the send/2 function checks the bucket properties to get the configured backend, requests the expiry time for that backend from the gen_server, and checks the key's age as determined by the X-Riak-Last-Updated metadata. If the object is older than the limit, the helper cancels sending the key. In the event that an error occurs, the helper fails open such that the key will be sent.
## Testing

An eunit tests has been introduced that starts up a test ring, configures a bitcask backend with an expiration of 60 seconds, and creates 1 bucket with a very old expiration, and one with a very young expiration, and then exercises the send/2 in the hook to confirm the return values.
## Issue

https://github.com/basho/riak_repl/issues/328

### DIFF
--- a/ebin/riak_repl.app
+++ b/ebin/riak_repl.app
@@ -23,6 +23,7 @@
                   'riak_repl_console',
                   'riak_repl_fsm_common',
                   'riak_repl_fullsync_helper',
+                  'riak_repl2_fullsync_expiry_hook',
                   'riak_repl2_fscoordinator',
                   'riak_repl2_fscoordinator_sup',
                   'riak_repl2_fscoordinator_serv',
--- a/src/riak_repl2_fullsync_expiry_hook.erl
+++ b/src/riak_repl2_fullsync_expiry_hook.erl
@@ -150,7 +150,7 @@ get_config() ->
           app_helper:get_env(?MODULE,backends,?DEFAULT_BACKENDS)));
       Other ->
         lager:log(debug,self(),"using ~s backend",[Other]),
-        get_expiry([{<<"riak_repl2_fullsync_expiry_hood_default_backend">>,Other,[]}],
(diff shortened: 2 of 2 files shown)


joecaswell:
~~-+1-~~

joecaswell:
I just noticed a potential name collision.  Line 91,141, and 160, if the user has configured a backend name <<"default">> that is not actually the default backend, replication of its keys may be cancelled whether it is configured for expiry or not.

Also, MAXVER at line 30 and version_check/0 at line 108 and line 49 are not required if this code is merged into the versioned product.

bowrocker:
Thanks.

Clear on MAXVER -- not so sure what the solution is for the "default" name.

So, "default" is not necessarily the default backend, i.e. there is no naming validation. How do we actually find the real default backend, then?

joecaswell:
The default backend is either the only backend, the first one specified in multi_backend, or the one specified by multi_backend_default.  riak_kv validates the config on startup, so I don't bother here.  Finding the default is not difficult, but when storing our default choice, we should pick a name like <<"riak_repl2_fullsync_expiry_hook_default_backend">>, possibly using a defined constant, so the likelihood of collision is significantly lower than using <<"default">>

bowrocker:
Updates to address Joe's input.

joecaswell:
re +1

PR #330, opened 2013-07-16
Integrate repl hook to avoid synchronizing expired objects [JIRA: RIAK-2695]

bowrocker:
# Overview

The riak_repl2_fullsync_expiry_hook module is a repl_helper hook function that operates on the sending side of a fullsync operation to cancel sending keys whose age, according to its internal last updated metadata, exceeds the configured expiry setting for its storage backend.

Code contributed by Joe Caswell.
## Configuration & Operation

At system start, the gen_server will read the backend configuration from the application environment and identify any backends that are configured for key expiry. If no expiry backends are found, the gen_server terminates. If any expiry backends are found, the gen_server stores the expiry times in its internal state and registers the module as a repl_helper with riak_core.

During fullsync, the send/2 function checks the bucket properties to get the configured backend, requests the expiry time for that backend from the gen_server, and checks the key's age as determined by the X-Riak-Last-Updated metadata. If the object is older than the limit, the helper cancels sending the key. In the event that an error occurs, the helper fails open such that the key will be sent.
# Testing

An eunit tests has been introduced that starts up a test ring, configures a bitcask backend with an expiration of 60 seconds, and creates 1 bucket with a very old expiration, and one with a very young expiration, and then exercises the send/2 in the hook to confirm the return values.
### Issue and original PR

https://github.com/basho/riak_repl/issues/328
https://github.com/basho/riak_repl/pull/329

### DIFF
--- a/ebin/riak_repl.app
+++ b/ebin/riak_repl.app
@@ -23,7 +23,6 @@
                   'riak_repl_console',
                   'riak_repl_fsm_common',
                   'riak_repl_fullsync_helper',
-                  'riak_repl2_fullsync_expiry_hook',
                   'riak_repl2_fscoordinator',
                   'riak_repl2_fscoordinator_sup',
                   'riak_repl2_fscoordinator_serv',
--- a/ebin/riak_repl.app
+++ b/ebin/riak_repl.app
@@ -23,6 +23,7 @@
                   'riak_repl_console',
                   'riak_repl_fsm_common',
                   'riak_repl_fullsync_helper',
+                  'riak_repl2_fullsync_expiry_hook',
                   'riak_repl2_fscoordinator',
                   'riak_repl2_fscoordinator_sup',
                   'riak_repl2_fscoordinator_serv',
--- a/src/riak_repl2_fullsync_expiry_hook.erl
+++ b/src/riak_repl2_fullsync_expiry_hook.erl
@@ -1,0 +1,281 @@
+%% Riak EnterpriseDS
+%% Copyright (c) 2007-2012 Basho Technologies, Inc.  All Rights Reserved.
+-module(riak_repl2_fullsync_expiry_hook).
+%%
+%% @doc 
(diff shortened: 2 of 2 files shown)


bookshelfdave:
Can you please rebase against `develop`?

bookshelfdave:
to be addressed after the 2.0 release

cmeiklejohn:
Marking 2.1.

nerophon:
A customer has requested that MDC support bitcask expiry. This PR would go a long way towards fulfilling that requirement.

angrycub:
create jira issue

Basho-JIRA:
[~pbrewer] [~angrycub] and I reviewed the bugs and this is one that we think should be addressed in 2.2 (From 2013 wow and brought up again by a customer) 

**_[posted via JIRA by Seema Jethani]**_

Basho-JIRA:
[~pbrewer] [~angrycub] have you two seen this?

https://github.com/basho/riak_repl/issues/328

Seems that this was never merged in the past because Engel had concerns about clock drift leading to expired data being resurrected. I'm not super familiar with Bitcask expiry so I'm trying to learn more about it and figure out exactly what scenarios we need to be worried about, but it looks like this may or may not be something we can merge as is. 

**_[posted via JIRA by Nick Marino]**_

bsparrow435:
This functionality is opt in and compiles and was tested manually. While there are some concerns around clock skew I do not believe that it warrants holding off on the merge.

nickelization:
Discussed this with @bsparrow435 and we agreed to revert this. In theory it may be harmless because it's opt-in and has been tested, but the issue is that it does not really solve the problem that it claims to solve. It makes the replication of expired data marginally less likely, but it is still very possible, so anyone enabling this hook is likely going to be frustrated that it did not fix the issues they expected it to. Because of this, and because of potential performance concerns, we're going to remove this module and close the PR.

(Additionally, Riak 2.3 will contain the long-awaited sweeper code, which provides a much more robust expiry mechanism.)

PR #340, opened 2013-07-26
backport riak_core PR 313, format _stat_ts bug

bookshelfdave:
https://github.com/basho/riak_core/pull/313

see also: https://github.com/basho/riak_repl/issues/339

PR #342, opened 2013-07-26
backport riak_core PR 313, format _stat_ts bug

bookshelfdave:
see https://github.com/basho/riak_repl/pull/340


bookshelfdave:
#340 has already been +1'd and tested against 1.3.2.

Issue #339, opened 2013-07-26
Address urgent customer issues in 1.3.2


bookshelfdave:
AKA: 1.3.3 release notes
- [x] increase the cluster manager timeouts as in 1.4 (https://github.com/basho/riak_repl/pull/344) (https://github.com/basho/riak_repl/pull/343)
- [x] address the formatting of that _ts stat crash (https://github.com/basho/riak_repl/pull/342)
- [x] dont' crash on stats poll (https://github.com/basho/riak_repl/pull/345)
- [ ] 1.3.3 release notes (https://github.com/basho/internal_wiki/wiki/Riak-1.3.3-release-notes-(Enterprise-only))
- [x] https://github.com/basho/riak_repl/pull/360
- [x] https://github.com/basho/riak_repl/pull/353
- [x] https://github.com/basho/riak_repl/pull/364
- [x] https://github.com/basho/riak_repl/pull/366
- [ ] email dev@ and TAMs

Note: .beams built with R15B01 + scheduler patch
Note: See https://github.com/basho/riak_repl/issues/359 for more info regarding realtime overflow 
Note: https://github.com/basho/riak_repl/pull/344 won't be included in beams shipped to OpenX, however riak_core_cluster_manager WILL be to increase timeout

buddhisthead:
With all PRs of this issue addressed, I think we can close this out now.

buddhisthead:
Re-open to track https://github.com/basho/riak_repl/pull/345, which I think should be added.

lordnull:
#345 has been merged, so all are checked again.

bookshelfdave:
See release notes at https://github.com/basho/internal_wiki/wiki/Riak-1.3.3-release-notes-(Enterprise-only)

bookshelfdave:
See also 1.4.1 issues list: https://github.com/basho/riak_repl/issues/362

PR #343, opened 2013-07-26
All node local calls to cluster manager are infinite

lordnull:
By infinite I mean the timeout has been set to infinite instead of 5
seconds. proxy'ed calls to the leader node still have a 30 second timeout.

This addresses item 2 in #339 .


bookshelfdave:
closed in favor of https://github.com/basho/riak_repl/pull/344

PR #344, opened 2013-07-27
Backport use of Long Timeouts from 1.4 to 1.3

buddhisthead:
Port most of https://github.com/basho/riak_repl/pull/299 from 1.4 to 1.3

The bulk of changes are to use explicit timeout values, rather than the very short default of 5 seconds.
Most will use the new LONG_TIMEOUT of 2 minutes, when they are possibly calls to remote nodes.
Known calls to local nodes now use infinity.
One additional catch of a timeout in the node_reserver makes it operate smoothly instead of crashing - it returns "down" which just mimics a node that is unavailable for work.

The commits from PR #299 were cherry-picked and edited to remove any cruft from 1.4. The EQC changes and changes to files that don't exist in 1.3 (proxy-get files) were removed.

Please note that this does address the gen_server call to get known ipaddrs in the cluster manager :-)

### DIFF
--- a/include/riak_repl.hrl
+++ b/include/riak_repl.hrl
@@ -25,6 +25,8 @@
 -define(DEFAULT_MAX_SINKS_NODE, 1).
 %% 20 seconds. sources should claim within 5 seconds, but give them a little more time
 -define(RESERVATION_TIMEOUT, (20 * 1000)).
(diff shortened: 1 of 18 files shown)


lordnull:
Verified via riak_test, and code is fine. :+1: .

buddhisthead:
Partially resolves issue https://github.com/basho/riak_repl/issues/339

PR #345, opened 2013-07-30
don't crash on bad stats

bookshelfdave:


bookshelfdave:
yeah, that's why I added the lager:warning. I agree that it should return the default. Not sure what the hell I was thinking.

lordnull:
I feel like this is just papering over the issue without actually fixing it. Looking over the code again, instead of just returning the default, a better fix might be checking if we've pulled the final value needed. Change the proplists_get to:

``` erlang
proplists_get(Path, Props) ->
    proplists_get(Path, Props, undefined).

proplists_get([], undefined, Default) ->
    Default;
proplists_get([], Value, _Default) ->
    Value;
proplists_get([Key], Props, Default) when is_list(Props) ->
    Value = proplists:get_value(Key, Props),
    proplists_get([], Value, Default);
proplists_get([Key | Path], Props, Default) when is_list(Props) ->
    case proplists:get_value(Key, Props) of
        undefined ->
            Default;
        AList when is_list(AList) ->
            proplists_get(Path, AList, Default);
        Wut ->
            erlang:error({badarg, {[Key | Path], Props, Wut}})
    end.
```

This checks if we're on the last key, thus returning whatever it got, where the previous version would crash. It also preserves a (in my opion) legitimate crash, but gives it a more useful message. My reason for wanting to keep the error rather than a warning is a message in the crash log is much more likely to get addressed an a warning in the console log.

buddhisthead:
@lordnull I like your version. Can we change to that?

buddhisthead:
Is this already fixed in 1.4? If not, we should fix in 1.4.1 tomorrow.

bookshelfdave:
See also: https://github.com/basho/riak_repl/pull/348 for 1.4

We'll need to get these changes to master as well.

lordnull:
Compiles, stats don't crash under a manual test, so this doesn't break existing stuff, and will give an informational message if it does break.

:+1:

lordnull:
We've gotten to see this is live now. The crash message is much more useful. Seems one of the values was 'too_busy' when a list was expected. If we can handle those, then we can drop back down to a warning and return the default.

Last clause:

``` erlang
proplists_get([Key | Path], Props, Default) when is_list(Props) ->
    case proplists:get_value(Key, Props) of
        undefined ->
            Default;
        AList when is_list(AList) ->
            proplists_get(Path, AList, Default);
        Wut ->
            lager:warning("Non-list return when trying to get ~p from ~p", [[Key | Path], Props]),
            Default
    end.
```

PR #348, opened 2013-07-31
don't crash on bad stats (1.4)

bookshelfdave:
See https://github.com/basho/riak_repl/pull/345


lordnull:
Compiles, manual test show no explosions. Basically same as #345 . :+1:

PR #353, opened 2013-08-02
Mw bad stats crash redux

lordnull:
An alteration to #345 based on results in the wild.


Vagabond:
I'd rather not generate a warning when encountering too_busy.

bookshelfdave:
+1

PR #356, opened 2013-08-05
Use {active, once} for the rt_sink socket.

lordnull:
Using {active, once} instead of {active, true} will limit the number
of messages and provide a built-in means of back pressure the the sink.
This change respects the sinks's ability to place the socket in passive
mode ({active, false}) if it gets too over loaded.


bookshelfdave:
+1, replication2 passes multiple times

PR #357, opened 2013-08-05
don't blow up on stat errors

lordnull:
Status 'too_busy' returns default value. This means A warning is only
generated if a truely unexpected value is found.

This is a squash and minor change to the 1.3.3 patch.

### DIFF
--- a/src/riak_repl_console.erl
+++ b/src/riak_repl_console.erl
@@ -960,6 +960,7 @@ proplists_get([Key | Path], Props, Default) when is_list(Props) ->
         undefined ->
             Default;
         too_busy ->
+            lager:debug("Something was too busy to give stats"),
             Default;
         AList when is_list(AList) ->
             proplists_get(Path, AList, Default);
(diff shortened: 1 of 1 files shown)


src/riak_repl_console.erl
```diff
@@ -952,14 +952,20 @@ proplists_get([], undefined, Default) ->
...
     Value;
+proplists_get([Key], Props, Default) when is_list(Props) ->
+    Value = proplists:get_value(Key, Props),
+    proplists_get([], Value, Default);
 proplists_get([Key | Path], Props, Default) when is_list(Props) ->
     case proplists:get_value(Key, Props) of
         undefined ->
             Default;
+        too_busy ->
+            Default;
```
bookshelfdave, line 964: Should a lager:debug() be used here in case we want to see what's truly going on?

bookshelfdave:
Should this PR be against 1.4 instead of master?

lordnull:
We've got caa8cf4160a6c60d9757c2d5ed4cd424f29be70a in 1.4. This is not a big enough change to merit back porting. No data loss, no new crashes, nor any different handling of existing crashes (or in this case recently fixed crashes).

With that said, the branch name is wonderfully misleading.

bookshelfdave:
+1

PR #361, opened 2013-08-06
Backport 356 for 1.4

buddhisthead:
Not known to pass riak_test yet, but expect it to. Please help test. Thank you.


bookshelfdave:
+1

Issue #362, opened 2013-08-06
Address urgent customer issues in 1.4.1


bookshelfdave:
The following PR's should be delivered to customers using _Riak Enterprise 1.4.1_:
- [x] https://github.com/basho/riak_repl/pull/361
- [x] https://github.com/basho/riak_repl/pull/357 (WON'T BE INCLUDED)
- [x] https://github.com/basho/riak_repl/pull/363
- [x] https://github.com/basho/riak_repl/pull/365
- [x] ship beams to OpenX, Symantec

requiring the following beams:
- riak_repl2_rtsink_conn
- riak_repl_console

.beams built with R15B01 + scheduler patch

Beam installation instructions:
https://help.basho.com/entries/21823664-PENDING-awaiting-final-beam-Prevent-RT-Overload-Patch

lordnull:
If the beams have been shipped (haven't they?) we should be able to close this issue.

PR #363, opened 2013-08-07
don't blow up on stats errors

bookshelfdave:
also, don't log too_busy
See https://github.com/basho/riak_repl/issues/362


bowrocker:
+1

PR #366, opened 2013-08-07
don't blow up on stat errors

bookshelfdave:
also, don't log too_busy
https://github.com/basho/riak_repl/issues/339


Vagabond:
+1

Issue #368, opened 2013-08-10
bad call to timer:nowdiff causes frequent repl source reconnections to the sink


bowrocker:
In the context of performance testing 1.3.3 and patches for 1.3.2, set up 1 node clusters on r2s15 and r2s16 in the boston colo and built 1.3.3. During a basho_bench run, noticed source reconnections, and saw the following in the log on the source side:

2013-08-09 20:58:03.268 [error] <0.8060.154> gen_server <0.8060.154> terminated with reason: no function clause matching timer:now_diff({1376,96283,268662}, undefined) line 206
2013-08-09 20:58:03.271 [error] <0.8060.154> CRASH REPORT Process <0.8060.154> with 0 neighbours exited with reason: no function clause matching timer:now_diff({1376,96283,268662}, undefined) line 206 in gen_server:terminate/6 line 747
2013-08-09 20:58:03.271 [info] <0.23038.161>@riak_repl2_rtsource_conn:handle_call:208 Established realtime connection to site "r2s16" address 10.0.24.26:10016
2013-08-09 20:58:03.273 [error] <0.2063.0> Supervisor riak_repl2_rtsource_conn_sup had child "r2s16" started with riak_repl2_rtsource_conn:start_link("r2s16") at <0.8060.154> exit with reason no function clause matching timer:now_diff({1376,96283,268662}, undefined) line 206 in context child_terminated

The reconnects appeared to cause drops and errors on the source side:

realtime_queue_stats: [{bytes,720},
                       {max_bytes,104857600},
                       {consumers,[{"r2s16",
                                    [{pending,0},
                                     {unacked,0},
                                     {drops,2212510},
                                     {errs,18}]}]}]

jonmeredith:
Based on the crash log entry

```
2013-08-09 20:35:51 =ERROR REPORT====
** Generic server <0.17043.129> terminating 
** Last message in was {tcp,#Port<0.16557>,<<0,0,0,1,0,0,0,0,1,0,0,0,0,1,0,0,0,0,1,0>>}
** When Server state == {state,"r2s16",{"10.0.24.26",10016},#Ref<0.0.45.214986>,ranch_tcp,#Port<0.16557>,{realtime,{1,1},{1,1}},<0.17061.129>,15000,15000,#Ref<0.0.49.60848>,{1376,94951,637085},119,<<>>}
** Reason for termination == 
[...]

PR #369, opened 2013-08-12
Do not calculate heartbeat RTT in riak_repl2_rtsource_conn if state#hb_sent is undefined

bowrocker:
### Issue

https://github.com/basho/riak_repl/issues/368
### Discussion

As described in the issue, occasionally more than one (4) heartbeat message arrive at once. This means that between calls to recv, state#hb_sent is never set to a valid timestamp, as in send_heartbeat/2. The current RTT calculation assumes hb_sent is valid and sends the value into timer:now_diff, but in this case it is not.

This PR is a simple fix to this problem: do not calculate the RTT hb_sent is undefined. However, this will have the effect of possibly skewing the RTT calculations, at least in this case.

This fix does appear to stop the errors from happening, since 'undefined' leads to a badmatch of timer:now_diff and consequently a server crash and RT connection restart.

### DIFF
--- a/src/riak_repl2_rtsource_conn.erl
+++ b/src/riak_repl2_rtsource_conn.erl
@@ -317,7 +317,7 @@ recv(TcpBin, State = #state{remote = Name,
             case HBSent of
                 undefined ->
                     lager:debug("hb_sent undefined in heartbeat RTT calc"),
-                    {noreply, State};
+                    recv(Cont, State);
                 _ ->
                     HBRTT = timer:now_diff(now(), HBSent) div 1000,
                     erlang:cancel_timer(HBTRef),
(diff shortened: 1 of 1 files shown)


src/riak_repl2_rtsource_conn.erl
```diff
@@ -314,12 +314,18 @@ recv(TcpBin, State = #state{remote = Name,
         {ok, heartbeat, Cont} ->
             %% Compute last heartbeat roundtrip in msecs and
             %% reschedule next
-            HBRTT = timer:now_diff(now(), HBSent) div 1000,
-            erlang:cancel_timer(HBTRef),
-            State2 = State#state{hb_sent = undefined,
+            case HBSent of
+                undefined ->
+                    lager:debug("hb_sent undefined in heartbeat RTT calc"),
+                    {noreply, State};
```
lordnull: This should be:

``` erlang
recv(Cont, State)
```

Otherwise you risk losing data; there may be a relavent frame after the irrelavent heartbeat.

PR #370, opened 2013-08-13
Queue implementation for heartbeat tracking

bowrocker:
#### Issue

Issue: https://github.com/basho/riak_repl/issues/368
#### Overview

Per discussions with Jon M and Andrew T, this solution is an alternative to simply not calculating RTT of heartbeats when multiple heartbeat messages arrive. Here, a queue of heartbeat sent messages is used to track the sent heartbeats (state#hb_sent_q), in case more than one is outstanding. The RTT calculation uses the first element in the queue.

### DIFF
--- a/src/riak_repl2_rtsource_conn.erl
+++ b/src/riak_repl2_rtsource_conn.erl
@@ -261,7 +261,7 @@ handle_info({tcp_error, _S, Reason},
 
 handle_info(send_heartbeat, State) ->
     {noreply, send_heartbeat(State)};
-handle_info({heartbeat_timeout}, State = #state{hb_sent_q = HBSentQ,
+handle_info(heartbeat_timeout, State = #state{hb_sent_q = HBSentQ,
                                                 remote = Remote}) ->
     case queue:out(HBSentQ) of 
         {TimeSent, HBSentQ2} ->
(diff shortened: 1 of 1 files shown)


src/riak_repl2_rtsource_conn.erl
```diff
@@ -255,21 +257,42 @@ handle_info({tcp_error, _S, Reason},
...
+            lager:info("Realtime connection ~s to ~p heartbeat rtt ~p "
+                       "milliseconds exceeded timeout ~p, "
+                       "but saved by incoming data\n",
+                       [peername(State), Remote, Duration, HBTimeout]),
+            {noreply, State2};
+        false ->
+            lager:warning("Realtime connection ~s to ~p heartbeat timeout "
+                          "after ~p milliseconds\n",
+                          [peername(State), Remote, Duration]),
+            lager:debug("hb_sent_q_len after heartbeat_timeout: ~s",
```
buddhisthead, line 292: This ~s needs to be a ~p (~d crashes too. why?)

or else...
[debug] <0.5101.0>@riak_repl2_rtsource_conn:handle_info:299 FORMAT ERROR: "hb_sent_q_len after heartbeat_timeout: ~s" [0]

buddhisthead:
ok, I'm confused. After looking at the code, I don't see how we can get more than 1 entries in the HB sent q.

buddhisthead:
I am also concerned about performance if we cancel and restart a timer on every single ack we get from the sink...

https://github.com/basho/riak_repl/blob/a5ffa41b028b8731a579a851b3656caf0391d1cf/src/riak_repl2_rtsource_conn.erl#L327

buddhisthead:
Finally, we can't cancel the HB timeout timer without guarding against the case where it is undefined. It can be undefined in the case where we have back-to-back acks or heartbeats arriving and we haven't scheduled a recent heartbeat yet.

buddhisthead:
superseded by #379

PR #372, opened 2013-08-14
Add overload check and recovery to realtime queue

lordnull:


src/riak_repl2_rtq.erl
```diff
@@ -73,7 +88,27 @@
...
+%% testing some options a bit easier as it removes a dependance on app_helper.
+-type overload_threshold_option() :: {'overload_threshold', pos_integer()}.
+-type overload_recover_option() :: {'overload_recover', pos_integer()}.
+-type start_option() :: overload_threshold_option() | overload_recover_option().
+-type start_options() :: [start_option()].
+-spec start_link(Options :: start_options()) -> {'ok', pid()}.
+start_link(Options) ->
+    case ets:info(?overload_ets) of
+        undefined ->
+            ets:new(?overload_ets, [named_table, public, {read_concurrency, true}]),
```
Vagabond, line 107: Should this not be a protected table?

Vagabond, line 107: Also, in start_link we are running in the context of the supervisor, not the RTQ?

src/riak_repl2_rtq.erl
```diff
@@ -73,7 +88,27 @@
 %% @doc Start linked, registeres to module name.
 -spec start_link() -> {ok, pid()}.
 start_link() ->
-    gen_server:start_link({local, ?SERVER}, ?MODULE, [], []).
+    Overload = app_helper:get_env(riak_repl, rtq_overload_threshold, ?DEFAULT_OVERLOAD),
+    Recover = app_helper:get_env(riak_repl, rtq_overload_recover, ?DEFAULT_RECOVER),
+    Opts = [{overload_threshold, Overload}, {overload_recover, Recover}],
```
Vagabond, line 94: Why default overload and recover here, but also in init()?

Vagabond:
I think you forgot to add the overload_counter module.

test/riak_repl2_rtq_tests.erl
```diff
@@ -0,0 +1,196 @@
...
+            % msg queue = 7 (blocked by deliver)
+            unblock_rtq_pull(),
+            % msq queue = 5 (push handled, blocking deliver handled)
+            % that push should have flipped the overload switch
+            % meaning these will be dropped
+            % these will end up dropped
+            [riak_repl2_rtq:push(1, term_to_binary([<<"object">>])) || _ <- lists:seq(1,5)],
+            % msq queue = 7, drops = 5
+            unblock_rtq_pull(),
+            timer:sleep(1200),
```
Vagabond, line 143: If possible, don't use timers likethis, because the test will fail on slower machines. Can we do something like send the RTQ a status request so we can serialize on the mailbox? Even a sys:get_status may do the trick...

src/riak_repl2_rtq.erl
lordnull, line 94: It's defaulted here to ensure it's sent to start_link/1 based on what is in app.config, but still with a valid value. Having it in init/1 means start_link([]) can be used for testing.

test/riak_repl2_rtq_tests.erl
lordnull, line 143: The timer isn't waiting for rtq, though, it's waiting for a message from rtq_overload_counter to be sent to the rtq, which itself is triggered by a timer. I agree it's better not to use a timer here, but I'm not sure how I would do it the 'right' way.

src/riak_repl2_rtq.erl
lordnull, line 107: Commented earlier, but github ate it.

There was discussion about the actual implementation detail of counting the drops. One of the options was putting the counter in the ets, and having the overload just read and reset it occasionally. The only way to do this is by making the ets public. Putting it under the supervisor means drop count doesn't get reset if the rtq would die.

src/riak_repl2_rtq.erl
```diff
@@ -340,6 +402,25 @@ terminate(Reason, #state{cs = Cs}) ->
 code_change(_OldVsn, State, _Extra) ->
     {ok, State}.
 
+maybe_flip_overload(State) ->
+    #state{overloaded = Overloaded, overload = Overload, recover = Recover} = State,
+    {message_queue_len, MsgQLen} = erlang:process_info(self(), message_queue_len),
+    lager:debug("checking overload: ~p", [MsgQLen]),
```
Vagabond: debug messages aren't entirely free, so leaving one of doubtful value in the hot path should be avoided.

Vagabond:
I'm generally +1 on this from a code perspective (nuking that lager:debug would be nice). I have not run the code at all, so maybe @bowrocker should also +1 it from a testing perspective.

bowrocker:
+1 from running lots and lots of  tests.

PR #373, opened 2013-08-15
Avoid arithmetic on folsom error tuples

engelsanchez:
Folsom errors happen, and they could make the repl stats process crash
because it tried to perform some calculations on those. This change
avoids those calculations when something other than a valid value is
returned.


bookshelfdave:
@engelsanchez could you resubmit against `develop`?

engelsanchez:
Yep, no problem.

cmeiklejohn:
Closing in favor of #482.

PR #379, opened 2013-08-16
Queue implementation for heartbeat tracking

buddhisthead:
NOTE: this is a replacement for PR #370 

Issue

Issue: #368

Overview

Per discussions with Jon M and Andrew T, this solution is an alternative to simply not calculating RTT of heartbeats when multiple heartbeat messages arrive. Here, a queue of heartbeat sent messages is used to track the sent heartbeats (state#hb_sent_q), in case more than one is outstanding. The RTT calculation uses the first element in the queue.

Test: https://github.com/basho/riak_test/pull/352

### DIFF
--- a/src/riak_repl2_rtsink_conn.erl
+++ b/src/riak_repl2_rtsink_conn.erl
@@ -19,6 +19,9 @@
          status/1, status/2,
          legacy_status/1, legacy_status/2]).
 
+%% Export for intercept use in testing
+-export([send_heartbeat/2]).
+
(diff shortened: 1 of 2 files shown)


buddhisthead:
I should also note that I refactored the "send_heartbeat" function on the RT sink side, to mimic the one in RT source, so that I could use intercepts in the new HB riak_test.

jonmeredith:
(thumbsup) +1, please merge.

PR #380, opened 2013-08-16
Drop bad data from the backend.

evanmcc:
wherever we're handling data straight out of the backend, don't trust it and make sure all binary_to_term calls are done inside of try-catch statements, then attempt to noop or exit safely.

### DIFF
--- a/src/riak_repl_fullsync_helper.erl
+++ b/src/riak_repl_fullsync_helper.erl
@@ -481,36 +481,48 @@ missing_key(PBKey, DiffState) ->
 %%
 %% @see http://www.javalimit.com/2010/05/passing-funs-to-other-erlang-nodes.html
 merkle_fold(K, V, Pid) ->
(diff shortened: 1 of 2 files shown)


Vagabond:
Looks safe enough, +1 on a code review side at least.

src/riak_repl_fullsync_helper.erl
```diff
@@ -481,36 +481,48 @@ missing_key(PBKey, DiffState) ->
...
-            {MPid, 0, Total+1};
-        _ ->
-            {MPid, Count+1, Total+1}
+    try
+        H = hash_object(V),
+        Bin = term_to_binary({pack_key(K), H}),
+        %% write key/value hash to file
+        riak_core_gen_server:cast(MPid, {keylist, Bin}),
+        case Count of
+            100 ->
```
evanmcc, line 501: It's a backpressure mechanism, so it likely should be, but it isn't really
germane here.

bookshelfdave:
+1 for code review also, I'll run it through existing riak_tests

PR #385, opened 2013-08-22
don't pop from the hb queue on timeout

bookshelfdave:


bowrocker:
I believe if there is more than one outstanding hb, and therefore 2 or more items in the queue, this may incorrectly calculate the RTT. Is that your understanding?

I think that's OK, but I just wanted to be sure I understood.

bookshelfdave:
@bowrocker yes, this could lead to rtt inaccuracies

we'll circle back and fix that in the future.

src/riak_repl2_rtsource_conn.erl
```diff
@@ -283,22 +280,19 @@ handle_info({heartbeat_timeout, HBSent}, State = #state{hb_sent_q = HBSentQ,
     %%   heartbeat interval timer is pending - so tref NOT undefined
     %%   heartbeat interval timer fired, new hb_sent entry added - so NOT qempty
 
-    State2 = State#state{hb_rtt = Duration, hb_sent_q = HBSentQ2},
     case IntervalTRef /= undefined orelse
-         queue:len(HBSentQ2) /= 0 of
+         queue:len(HBSentQ) /= 0 of
         true ->
```
lordnull, line 285: These lines hurt to read. If I'm parsing the logic right, then the following is easier to grok (logging removed just for example, real code would log):

``` erlang
case {IntervalTRef, queue:len(HBSentQ)} of
    {undefined, 0} -> {stop, normal, State};
    _ -> {noreply, State}
end
```

bookshelfdave, line 285: agreed that it's hard to read, but it was there before and I'd like to NOT touch it ;-) (well, minus the HBSentQ2 -> HBSentQ change)

bowrocker:
+1

PR #386, opened 2013-08-23
don't pop from the hb queue on timeout - 1.4

bookshelfdave:
See also https://github.com/basho/riak_repl/pull/385


lordnull:
I feel like I've seen somewhere before...

+1 since code is the same.

PR #391, opened 2013-08-27
Fix for heartbeat timeout bug discovered in riak_test repl_rt_heartbeat

bowrocker:
The test uses an intercept to drop heartbeat messages and then waits to ensure RT restart. This was discovered not to be working on 1.4 in advance of the ATT upgrade this evening.

The problem was the code was checking the queue for zero len, and not stopping if true.

Fix was to remove check of heartbeat queue, and make  #state.hb_timeout_tref in riak_repl2_rtsource_conn:handle_info/3 the authority of whether to stop, and restart the connection.


jonmeredith:
+1 merge.

Reviewed code.  Confirmed r_t failed before and passes after.  Needs to be backported to 1.3 branch.

```
repl_rt_heartbeat-bitcask: pass
---------------------------------------------
0 Tests Failed
1 Tests Passed
That's 100.0% for those keeping score
```

PR #392, opened 2013-08-28
Fix for heartbeat timeout bug discovered in riak_test repl_rt_heartbeat

bowrocker:
Merge to 1.4 of following:

The test uses an intercept to drop heartbeat messages and then waits to ensure RT restart. This was discovered not to be working on 1.4 in advance of the ATT upgrade this evening.

The problem was the code was checking the queue for zero len, and not stopping if true.

Fix was to remove check of heartbeat queue, and make #state.hb_timeout_tref in riak_repl2_rtsource_conn:handle_info/3 the authority of whether to stop, and restart the connection.

Original PR:

https://github.com/basho/riak_repl/pull/391


jonmeredith:
+1 merge

PR #398, opened 2013-09-11
Move tcp-mon, connection and service manager supervisors from replication to core

buddhisthead:
In order for the connection and service manager to be fully moved to core, we need to move
their supervisors. Also, I've moved the tcp_mon supervisor because it's really a core thing too.

The matching PR fro core is here...
https://github.com/basho/riak_core/pull/383

Doesn't that code look cleaner now!

NOTE: Please do not merge after the +1 here. I need to press the merge buttons on both riak_repl and riak_core at nearly the same time :-) Thanks.

### DIFF
--- a/src/riak_core_cluster_mgr_sup.erl
+++ b/src/riak_core_cluster_mgr_sup.erl
@@ -22,23 +22,11 @@ init([]) ->
                           fun riak_repl_app:cluster_mgr_write_cluster_members_to_ring/2,
                           fun riak_repl_app:cluster_mgr_read_cluster_targets_from_ring/0],
     Processes =
-        [%% TODO: move these to riak core
-         %% Service Manager (inbound connections)
-         {riak_core_service_mgr, {riak_core_service_mgr, start_link, []},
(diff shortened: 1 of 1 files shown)


bookshelfdave:
This PR needs to be rebased to pull in the cuttlefish changes. It fails to build otherwise: https://gist.github.com/metadave/14b0a4672ee68e1e872f

buddhisthead:
The poor thing is so old. I will rebase. Thanks!

buddhisthead:
I had to force push after the rebase. Please get a fresh version of this branch. I did confirm correct build after cloning a fresh ee and checking out vet-conn-mgr-supervisors in _both_ riak_repl and riak_core.

cmeiklejohn:
I think last time we discussed this, we had decided to move all of this back to repl until we were able to clean up some of the more problematic areas and get the design right, isn't that correct @jonmeredith?

buddhisthead:
Yes, we are punting on this PR for 2.0.

cmeiklejohn:
Closing.

Issue #306, opened 2013-06-12
Handle dead clusterid's for replicated CS manifests


bookshelfdave:
If a customer disables an entire cluster, how are blocks replicated if manifests have already been replicated with an old clusterid?

Open for discussion.

bowrocker:
Discussed proposed solution with Andrew T:
- Introduce a configuration parameter that maps the current cluster id to one that will serve the blocks after current is disabled. User will configure.
- Mapping will be placed in the ring meta data in riak_repl_ring. 
- riak_repl_pb_get:process/1, which takes a cluster_id, will first check for the mapping, and use the mapped cluster_id if such exists; otherwise it will use the incoming cluster_id

bowrocker:
Closing based on pull requests above.

PR #399, opened 2013-09-12
Handle decommissioned clusterid for replicated CS manifests

bowrocker:
#### Overview

Repl support for feature 306. Provide a redirection mechanism, via configuration, allowing proxy_get to be directed to an alternate cluster id from one that may be decommissioned or otherwise unreachable.

Initially, the configuration will be accomplished using "raw" cluster ids, instead of cluster names, since there is no existing, non dist-erlang API for getting a remote cluster name from a connected cluster. This will be addressed in follow-on versions.
#### Configuration

Four riak-repl sub-commands are added:
- show-local-cluster-id: displays the local cluster_id as a tuple, to be used in further configurations
- add-block-provider-redirect [from-cluster-id] [to-cluster-id]
- delete-block-provider-redirect [from-cluster-id]
- show-block-provider-redirect 
#### Issue

https://github.com/basho/riak_repl/issues/306

### DIFF
--- a/src/riak_repl_console.erl
+++ b/src/riak_repl_console.erl
@@ -32,7 +32,11 @@
          cascades/1,
          show_nat_map/1,
          add_nat_map/1,
-         del_nat_map/1
+         del_nat_map/1,
+         add_block_provider_redirect/1,
+         show_block_provider_redirect/1,
+         show_local_cluster_id/1,
+         delete_block_provider_redirect/1
      ]).
 
 add_listener(Params) ->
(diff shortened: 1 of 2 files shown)


src/riak_repl_console.erl
```diff
@@ -569,7 +573,28 @@ del_nat_map([External, Internal]) ->
...
+            io:format("No mapping for ~p~n", [FromClusterId]);
+        ToClusterId ->
+            io:format("Cluster id ~p redirecting to cluster id ~p~n", [FromClusterId, ToClusterId])
+    end.
+
+delete_block_provider_redirect([FromClusterId]) ->
+    lager:info("Deleting redirect to ~p", [FromClusterId]),
+    riak_core_metadata:delete({<<"replication">>, <<"cluster-mapping">>}, FromClusterId).
+
+show_local_cluster_id([]) ->
```
lordnull, line 598: I was about to comment on how there's a function for clustername already, but realized that clustername =/= cluster id. A comment or note to remind future coders about that distinction may be a good idea.

bowrocker, line 598: Added comment explaining that the feature operated with cluster_ids for the present time.

lordnull:
:+1:

PR #400, opened 2013-09-12
Fixed realtime protocol version check

lordnull:
The wrong check was causing the ack to be proxied through the helper.
The helper builds a map for acks from v1 to the actual v2 sequence
number. However, because this was a version 2 protocol, the map was
not needed, and thus not built. Without finding the v1 sequence number,
and ack is never sent to the rtq, leaving the object forever unacked.

### DIFF
--- a/src/riak_repl2_rtsource_conn.erl
+++ b/src/riak_repl2_rtsource_conn.erl
@@ -344,7 +344,7 @@ recv(TcpBin, State = #state{remote = Name,
     case riak_repl2_rtframe:decode(TcpBin) of
         {ok, undefined, Cont} ->
             {noreply, State#state{cont = Cont}};
-        {ok, {ack, Seq}, Cont} when ProtoMajor == 2 ->
+        {ok, {ack, Seq}, Cont} when ProtoMajor >= 2 ->
             %% TODO: report this better per-remote
             riak_repl_stats:objects_sent(),
             ok = riak_repl2_rtq:ack(Name, Seq),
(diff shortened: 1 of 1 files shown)


src/riak_repl2_rtsource_conn.erl
```diff
@@ -340,10 +340,11 @@ recv(TcpBin, State = #state{remote = Name,
                             hb_timeout_tref = HBTRef}) ->
     %% hb_timeout_tref might be undefined if we have are getting
     %% acks/heartbeats back-to-back and we haven't sent a heartbeat yet.
+    {realtime, {ProtoMajor, _}, {ProtoMajor, _}} = State#state.proto,
     case riak_repl2_rtframe:decode(TcpBin) of
         {ok, undefined, Cont} ->
             {noreply, State#state{cont = Cont}};
-        {ok, {ack, Seq}, Cont} when State#state.proto == {2,0} ->
+        {ok, {ack, Seq}, Cont} when ProtoMajor == 2 ->
```
jonmeredith: which makes least surprise `==2` or `>= 2` ?

lordnull: Depends on what needs to happen if/when we get to major version 3 of the protocol. If it's a big enough change, there'll be a whole new clause for it. If it doesn't need to change much (in which case why is it a new major version) then >= 2 is better.

jonmeredith: I think >=2 is better, least surprise. It's on whoever introduces major=3 to make sure it's right.

bowrocker:
Passes the rt_cascading test for it in this PR:

https://github.com/basho/riak_test/pull/386

+1

PR #406, opened 2013-09-23
upgrade rebar, added Makefile, make cluster_mgr tests less chatty

bookshelfdave:
Makefile has an eunit target and an eqc_dev target.
The eqc_dev target uses `compile_only=true` from the new version of rebar.


bookshelfdave:
@jaredmorrow any problems with adding a Makefile to this rebar?

jaredmorrow:
Fine by me.

PR #408, opened 2013-09-23
Fixed repl mutator registration

lordnull:
Registers only if {riak_kv, mutators} capability is true (it
defaults to false).

There is not test because building and starting a cluster _is_ the test (ergo, any riak_test tests this code path). The old version of the code crashed develop on startup due to undef'ed function. This will also ensure shrunk objects can be consistently unshrunk across the cluster.


bookshelfdave:
+1
the replication2 test fails related to [config changes](https://github.com/basho/riak_test/blob/master/tests/replication2.erl#L544), but the cluster starts.

PR #417, opened 2013-10-03
Bugfix/repl hook att from 1.4

bookshelfdave:


lordnull:
#418 Seems to include this pr, so should we just close this and use that?

bookshelfdave:
er, not sure how I created to PR's for this. Thanks @lordnull

PR #425, opened 2013-10-18
fix upgrades from 1.2.5anya -> 1.4.2

bookshelfdave:
Specifically for Apple


lordnull:
Code is reasonably clear.

I set the anya_fs_compat on a homogenous cluster, and it was able to complete.
Same sink (1.4) was able to complete a fullsync with a 1.2.5 cluster.

So, this gets a :+1: from me.

bookshelfdave:
I suppose this shouldn't be merged directly against 1.4. I'll create a 1.4anya tomorrow AM.

bookshelfdave:
This was merged right to 1.4 as it was decided that it won't affect customers other than apple.

PR #430, opened 2013-11-08
Reregister cluster name locator.

cmeiklejohn:
When connection_mgr crashes, and the supervisor restarts, ensure we
reregister the CLUSTER_NAME_LOCATOR_TYPE.  This solution is suboptimal,
but works because he supervisor is a rest_for_all.


cmeiklejohn:
#431 is the real solution for 2.0+ here, but shouldn't we merge this considering we did ship these beams to OpenX as a patch for 1.4.4? @metadave

bookshelfdave:
+1

cmeiklejohn:
We need to get a solution to this for the 2.0 RC, because it's currently still broken there.

PR #435, opened 2013-11-11
Handle race by treating cluster as disconnected.

cmeiklejohn:
Prevent a hard crash when the cluster has been connected, but the
gen_server has not been registered because we're still in the process of
reconfiguring after a topology change in riak_repl2_pg_block_requester.

PR against develop branch. @metadave


bookshelfdave:
+1, builds fine

PR #436, opened 2013-11-13
fix misleading error message

bookshelfdave:
this will need to be made against develop as well


bowrocker:
Accurate error messages are always good. 

:+1:

PR #440, opened 2013-11-21
Ensure re-election connects clusters.

cmeiklejohn:
In the event that a node is added to a cluster, when it's already the
leader in a singleton cluster, if it happens to be re-elected ensure
that we have all of the necessary connections to remote clusters that
might be necessary.

Needs testing, we need to identify how to reproduce.


bookshelfdave:
passes replication2, replication2:test_bidirectional_pg
+1

cmeiklejohn:
@joecaswell since they are running purely in a test environment, do we want to ship them a beam for this as we can't reproduce locally yet?  not sure how to handle this...

cmeiklejohn:
Closing; irrelevant.

PR #442, opened 2013-11-22
Make the rt sink socket mode configurable

bowrocker:
#### Overview

Until 1.4, the RT sink mode was hardcoded to {active, true}. Because this behavior sometimes leads to overload on the sink, because no flow control is possible with this setting, it was changed to {active, once}. OpenX is seeing underperformance with this setting. This PR makes the setting configurable using app.config.
#### Configuration

The realtime sink socket mode is configurable using the following in app.config:

{riak_repl, sink_socket_mode, {mode}

Where valid values for mode are:

{active, true}
{active, once}

NB: though {active, false} is a valid setting socket setting, the real time sink module does not support this.

The semantics of these settings are outlined here:

http://www.erlang.org/doc/man/inet.html#setopts-2

Specifically:

"If the value is true, which is the default, everything received from the socket will be sent as messages to the receiving process.

If the value is once ({active, once}), one data message from the socket will be sent to the process. To receive one more message, setopts/2 must be called again with the {active, once} option."

### DIFF
--- a/src/riak_repl2_rtsink_conn.erl
+++ b/src/riak_repl2_rtsink_conn.erl
@@ -33,6 +33,7 @@
                 ver,              %% wire format agreed with rt source
                 max_pending,      %% Maximum number of operations
                 active = true,    %% If socket is set active
+                socket_mode = {active, once}, %% Configurable, default is {active, once}
                 deactivated = 0,  %% Count of times deactivated
                 source_drops = 0, %% Count of upstream drops
                 helper,           %% Helper PID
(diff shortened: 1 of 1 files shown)


cmeiklejohn:
If `{active, false}` is not supported, we probably should error if we detect it's been configured that way, right?  

Also, why not have the user just provide the second part of the tuple, such as `true` or `once`?

jaredmorrow:
This is not destined / intended for 1.4.4 anymore.

cmeiklejohn:
Should we just close this out?

bookshelfdave:
Closed, possibly moving to 2.0+

PR #448, opened 2013-12-03
Ensure we re-connect to leader.

cmeiklejohn:
Prevent connections to non-leader nodes, when the sink leader changes,
because that's where all proxy_get requests were originating.  This
prevents a non-leader node from being the primary connection for the
proxy_get requests.

### DIFF
--- a/src/riak_core_cluster_mgr.erl
+++ b/src/riak_core_cluster_mgr.erl
@@ -206,7 +206,7 @@ init(Defaults) ->
         false ->
             ServiceProto = {?CLUSTER_PROTO_ID, [{1,0}]},
             ServiceSpec = {ServiceProto, {?CTRL_OPTIONS, ?MODULE, ctrlService, []}},
-            riak_core_service_mgr:register_service(ServiceSpec, {round_robin,?MAX_CONS});
+            riak_core_service_mgr:sync_register_service(ServiceSpec, {round_robin,?MAX_CONS});
         true ->
             ok
     end,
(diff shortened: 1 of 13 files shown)


src/riak_repl2_pg_block_requester_sup.erl
```diff
@@ -18,9 +19,15 @@ started(Node) ->
     [{Remote, Pid} || {Remote, Pid, _, _} <-
                       supervisor:which_children({?MODULE, Node}), is_pid(Pid)].
 
+%% @doc Terminate all connections.
+terminate_connections() ->
+    [exit(Pid, kill) || {_, Pid, _, _} <- supervisor:which_children(?MODULE)],
```
jonmeredith: Does this generate cruft in crash.log?  Would `catch supervisor:terminate_child(Pid)` be better than a brutal kill?

src/riak_core_cluster_mgr.erl
```diff
@@ -695,28 +695,30 @@ connect_to_targets(Targets) ->
...
     lager:info("ClusterManager: ~p becoming the leader", [LeaderNode]),
-    %% start leading and tell ourself to connect to known clusters in a bit.
-    %% Wait enough time for the ring to be stable
-    %% so that the call into the repl_ring handler won't crash.
-    %% We can try several time delays because it's idempotent.
+    ok = riak_repl2_pg_block_requester:register_service(),
     erlang:send_after(5000, self(), connect_to_clusters),
     State#state{is_leader = true};
 become_leader(State, LeaderNode) ->
     lager:debug("ClusterManager: ~p still the leader", [LeaderNode]),
```
jonmeredith, line 709: Is sending the connect_to_clusters message unnecessary now?  What triggers the connection when a singleton node is joined to a cluster and elected leader immediately?

src/riak_repl_app.erl
```diff
@@ -88,10 +87,15 @@ start(_Type, _StartArgs) ->
             register_cluster_name_locator(),
 
             %% makes service manager start connection dispatcher
-            riak_repl2_rtsink_conn:register_service(),
-            riak_repl2_fssink:register_service(),
-            riak_repl2_fscoordinator_serv:register_service(),
-            riak_repl2_pg_block_requester:register_service(),
+            ok = riak_repl2_rtsink_conn:register_service(),
```
jonmeredith: These should become sync_register_service calls as you still have the potential for racey behavior ahead of the `service_up`, but I suppose that expands this PR to hit many more modules.

src/riak_core_cluster_mgr.erl
cmeiklejohn, line 709: `connect_to_clusters` is only needed on the source side, when the leader changes.  I don't have an explicit case for testing whether the addition of that line is necessary or not.  On the sink side, with unidirectional replication, since no cluster information is stored in the ring, this call does nothing.

src/riak_repl_app.erl
cmeiklejohn: I'm amending this commit with these switch to synchronous calls.

cmeiklejohn:
Closing; this patch is irrelevant.

PR #450, opened 2013-12-03
Return error when we can't find IP locally.

cmeiklejohn:
This prevent a crash when attempting to sync a particular partition and
we can't figure out who to connect to because of natting.


cmeiklejohn:
@Vagabond tagging you.  Looks like this should also be a candidate for the `develop` branch as well.

Vagabond:
I'm +1 on the code, although it is unclear if anyone wants to lab up a NAT setup to test it.

cmeiklejohn:
Merging; we'll address setting up a NAT testing environment in the future.

PR #451, opened 2013-12-03
kill current nodes() pg_proxies upon leader election

cmeiklejohn:
kill current nodes() pg_proxies upon leader election 

this will force a reconnect from the source

### DIFF
--- a/src/riak_repl2_leader.erl
+++ b/src/riak_repl2_leader.erl
@@ -121,7 +121,7 @@ handle_cast({register_notify_fun, Fun}, State) ->
     %% Notify the interested party immediately, in case leader election
     %% has already occured.
     Fun(State#state.leader_node, State#state.leader_pid),
-    {noreply, State#state{notify_funs=[Fun | State#state.notify_funs]}};
+    {noreply, State#state{notify_funs=State#state.notify_funs ++ [Fun]}};
 
 handle_cast({set_candidates, CandidatesIn, WorkersIn}, State) ->
     Candidates = lists:sort(CandidatesIn),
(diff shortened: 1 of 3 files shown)


cmeiklejohn:
:-1: does not resolve the failure, sits in a constant loop of this:

```
2013-12-03 16:52:26.395 [info] <0.2858.0>@riak_repl_pb_get:process:78 doing replicated GET using cluster id <<"{'dev1@127.0.0.1',{1386,107197,711403}}">>
2013-12-03 16:52:26.396 [info] <0.2858.0>@riak_repl_pb_get:process:129 not connected to cluster <<"{'dev1@127.0.0.1',{1386,107197,711403}}">>
```

cc: @Vagabond

cmeiklejohn:
:+1:, this passes both tests for me and looks good.  This should also be applied to 2.0/develop.

cmeiklejohn:
This should be validated with the following riak_test: https://github.com/basho/riak_test/pull/453

cmeiklejohn:
This patch needs to be ported forward to develop, as this is still failing on the 2.0 branch.

cmeiklejohn:
rt_cascading is failing, but it's because of the test environment as far as I can tell; both without this patch and against the 1.4.3 tag I'm getting the same failures.

cmeiklejohn:
running replication2_upgrade right now...

cmeiklejohn:
I can't get replication2_upgrade to complete locally because of timeouts.

cmeiklejohn:
I'm :+1: on this patch, it appears to be passing all tests except rt_cascading and replication2_upgrade.  

The rt_cascading failures appear to be related to assumptions that the previous and legacy releases support cascading writes, which they do not in the case of 1.4.4.  

replication2_upgrade appears to be timing out for me around line 366 of the test, and I believe that's related to running this test locally.  Considering @metadave's got the builder configured for to run tests as his user, and I would have to change all of the permissions to be able to run these myself, I'll let him decide if he wants to run the test again before merging or defer to letting the builder run considering the passes we are getting on the rest of the suite.

cmeiklejohn:
replication2_upgrade passes in BOS.

PR #461, opened 2013-12-06
spawn the heartbeat from the source

bookshelfdave:
to prevent a pull from blocking for eva


cmeiklejohn:
What tests specifically should I validate this PR against?

cmeiklejohn:
repl_rt_heartbeat passes, merging 461.

PR #464, opened 2013-12-07
Resolve race condition when joining a node.

cmeiklejohn:
In the event that a joining node happens to trigger an election, which
completes prior to ring convergence, it will never receive the repl
configuration on the source side in time to perform outbound connections
to the sink cluster.

Manually trigger the connection handler code when a ring event occurs,
and the current node is the leader.

This fix, unfortunately, ties riak_repl2_leader to cluster manager,
which is probably undesirable moving forward, but provides a compatible
patch for the upcoming 1.4.4 release.

This is a patch for the 1.4 branch, but should also be ported forward to
2.0.

### DIFF
--- a/src/riak_core_cluster_mgr.erl
+++ b/src/riak_core_cluster_mgr.erl
@@ -94,7 +94,8 @@
          get_connections/0,
          get_ipaddrs_of_cluster/1,
          set_gc_interval/1,
-         stop/0
+         stop/0,
+         connect_to_clusters/0
          ]).
 
 %% gen_server callbacks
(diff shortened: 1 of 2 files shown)


joedevivo:
+1 looks good.

bsparrow435:
You could use `riak_core_cluster_mgr:get_is_leader().` instead of riak_repl2_leader if you didn't want them tied together. But this probably has it's own set of tradeoffs.

cmeiklejohn:
Amended commit to remove additional logging.

src/riak_core_cluster_mgr.erl
```diff
@@ -245,6 +251,11 @@ handle_call({get_my_members, MyAddr}, _From, State) ->
 handle_call(leader_node, _From, State) ->
     {reply, State#state.leader_node, State};
 
+handle_call(connect_to_clusters, _From, State) ->
+    lager:info("handling sync connect_to_clusters call"),
```
bookshelfdave: maybe a more customer friendly lager msg?

bookshelfdave:
:+1:

src/riak_core_cluster_mgr.erl
cmeiklejohn: Removed as it's not very necessary.

PR #465, opened 2013-12-09
Resolve race condition when joining a node.

cmeiklejohn:
Port ring handler fix to the 2.0 branch.

In the event that a joining node happens to trigger an election, which
completes prior to ring convergence, it will never receive the repl
configuration on the source side in time to perform outbound connections
to the sink cluster.

Manually trigger the connection handler code when a ring event occurs,
and the current node is the leader.

This fix, unfortunately, ties riak_repl2_leader to cluster manager,
which is probably undesirable moving forward, but provides a compatible
patch for the upcoming 1.4.4 release.

This is a patch for the 1.4 branch, but should also be ported forward to
2.0.

### DIFF
--- a/src/riak_core_cluster_mgr.erl
+++ b/src/riak_core_cluster_mgr.erl
@@ -94,7 +94,8 @@
          get_connections/0,
          get_ipaddrs_of_cluster/1,
          set_gc_interval/1,
-         stop/0
+         stop/0,
+         connect_to_clusters/0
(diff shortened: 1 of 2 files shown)


cmeiklejohn:
@metadave

cmeiklejohn:
@metadave another.

bookshelfdave:
+1

PR #467, opened 2013-12-09
Prevent potential race condition.

cmeiklejohn:
Prevent a potential race condition where services may or may not be
registered by the service manager because of the ordering of
registration, and the nature of cast'ing the registration messages.

Conflicts:
    src/riak_repl2_rtsink_conn.erl
    src/riak_repl_app.erl

@metadave


bookshelfdave:
+1, I just reviewed the same change for 1.4

cmeiklejohn:
MERGE DAT.

PR #472, opened 2013-12-11
Hardcode locator functions to prevent race/deadlocks.

cmeiklejohn:
Prevent races where locators are registered after requests come in
because of the startup dependency between riak_core_connection_mgr and
riak_core_cluster_mgr.


kellymclaughlin:
:+1:

PR #476, opened 2013-12-11
Add riak_core_connection and tests.

cmeiklejohn:
Move riak_core_connection, headers and tests back
to riak_repl as it's not needed in riak_core.


bookshelfdave:
+1, tested basic connectivity between 2 clusters

PR #477, opened 2013-12-12

bowrocker:
#### Overview

###### Pros
###### Cons
#### Configuration

| Source | Sink | Result |
| --- | :-: | --: |
| {2,1} | {2,1} | "default" and user defined typed buckets replicated |
| {2,1} | {2,0} or less | only "default" buckets replicated |
#### Dependencies

### DIFF
--- a/src/riak_repl2_rt.erl
+++ b/src/riak_repl2_rt.erl
@@ -143,7 +143,7 @@ postcommit(RObj) ->
         %% on what the RT source and sink negotiated as the common version).
         Objects0 when is_list(Objects0) ->
             Objects = Objects0 ++ [RObj],
-            Meta = check_for_typed_bucket(RObj),
+            Meta = set_bucket_meta(RObj),
             BinObjs = riak_repl_util:to_wire(w1, Objects),
             %% try the proxy first, avoids race conditions with unregister()
             %% during shutdown
@@ -217,12 +217,18 @@ do_ring_trans(F, A) ->
             ER
     end.
 
-check_for_typed_bucket(Obj) ->
+set_bucket_meta(Obj) ->
     M = orddict:new(),
     case riak_object:bucket(Obj) of
-        {_T, _B } ->
-            lager:info("found typed bucket, setting meta data"),
-            orddict:store(typed_bucket, true, M);
+        {Type, _B} ->
+            AllProps = riak_core_bucket_type:get(Type),
+            CompareProps = proplists:delete(claimant, AllProps),
+            PropsHash = erlang:phash2(proplists:delete(claimant, AllProps)),
+            lager:debug("typed bucket, setting meta data: Type:~p, HashProps:~p", [Type, PropsHash]),
+            M1 = orddict:store(typed_bucket, true, M),
+            M2 = orddict:store(type, Type, M1),
+            orddict:store(props_hash, PropsHash, M2);
         _B -> 
+            lager:debug("un-typed bucket, setting typed_bucket to false"),
             orddict:store(typed_bucket, false, M)
     end.
--- a/src/riak_repl2_rt.erl
+++ b/src/riak_repl2_rt.erl
@@ -143,15 +143,16 @@ postcommit(RObj) ->
         %% on what the RT source and sink negotiated as the common version).
         Objects0 when is_list(Objects0) ->
             Objects = Objects0 ++ [RObj],
+            Meta = check_for_typed_bucket(RObj),
             BinObjs = riak_repl_util:to_wire(w1, Objects),
             %% try the proxy first, avoids race conditions with unregister()
             %% during shutdown
             case whereis(riak_repl2_rtq_proxy) of
                 undefined ->
-                    riak_repl2_rtq:push(length(Objects), BinObjs);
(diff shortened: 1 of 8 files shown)


src/riak_repl_util.erl
```diff
@@ -108,44 +108,62 @@ get_partitions(Ring) ->
...
+    Key = riak_object:key(Object),
     case repl_helper_recv(Object) of
         ok ->
-            ReqId = erlang:phash2({self(), os:timestamp()}),
-            B = riak_object:bucket(Object),
-            K = riak_object:key(Object),
-            Opts = [asis, disable_hooks, {update_last_modified, false}],
+            case Bucket of
+               {T,B} ->
+                   case riak_core_bucket_type:status(T) of
```
jrwest: I may be missing some other area that does the additional checking but isn't it also necessary to check the bucket types are equal\* here? Only checking for existence doesn't prevent us from replicating bucket type `X` with `consistent=true` on the source to the sink who has bucket type `X` set to `consistent=false, datatype=map`. 

\* where equal means some subset of "important" properties are the same on both sides of the cluster?

bowrocker: Agreed. I would think the ones you mention: strongly consistent objects, and CRDTs, at the least.

Is there an example of this sort of checking anywhere? I am guessing what needs to happen is the properties of the incoming bucket type need to be compared with the properties of the type on the sink...

jrwest: The only checking I can think of is `riak_kv_bucket:validate` but thats not totally the same. There are certainly some other properties we care about as well (n_val? allow_mult?). I guess we could implement it w/o knowing the set of properties we need to check upfront (just do all?) and before merge we can get the proper list. Thoughts?

src/riak_repl2_rt.erl
```diff
@@ -143,15 +143,16 @@ postcommit(RObj) ->
...
         Objects0 when is_list(Objects0) ->
             Objects = Objects0 ++ [RObj],
+            Meta = check_for_typed_bucket(RObj),
             BinObjs = riak_repl_util:to_wire(w1, Objects),
             %% try the proxy first, avoids race conditions with unregister()
             %% during shutdown
             case whereis(riak_repl2_rtq_proxy) of
                 undefined ->
-                    riak_repl2_rtq:push(length(Objects), BinObjs);
+                    riak_repl2_rtq:push(length(Objects), BinObjs, Meta);
```
bookshelfdave, line 152: I admit I haven't looked, but can the call to riak_repl2_rtq:push/3 break backwards compat w/ repl 1.3 + 1.4?

src/riak_repl2_rt.erl
```diff
@@ -217,3 +216,13 @@ do_ring_trans(F, A) ->
         ER ->
             ER
     end.
+
+check_for_typed_bucket(Obj) ->
+    M = orddict:new(),
+    case riak_object:bucket(Obj) of
+        {_T, _B } ->
+            lager:info("found typed bucket, setting meta data"),
```
bookshelfdave: lager:debug

src/riak_repl2_rtsink_conn.erl
```diff
@@ -54,7 +54,7 @@
 
 %% Register with service manager
 register_service() ->
-    ProtoPrefs = {realtime,[{2,0}, {1,4}, {1,1}, {1,0}]},
+    ProtoPrefs = {realtime,[{2,1}, {2,0}, {1,4}, {1,1}, {1,0}]},
```
bookshelfdave, line 57: general comment, we should document what each version does :-(
not required for a PR +1 though
I see you have it in the PR writeup, maybe we should keep track of these elsewhere?

src/riak_repl_util.erl
```diff
@@ -108,44 +108,62 @@ get_partitions(Ring) ->
...
+                       undefined ->
+                           lager:error("Bucket type ~p not defined on sink!", [T]);
```
bookshelfdave: probably don't need the !'s, customers will shit their pants regardless

src/riak_repl_util.erl
```diff
@@ -108,44 +108,62 @@ get_partitions(Ring) ->
...
+                       created ->
+                           lager:error("Bucket type ~p exists on sink, but is not active!", [T]);
+                       ready ->
+                           lager:error("Bucket type ~p exists on sink, but is not active!", [T]);
+                       active ->
+                           lager:debug("Bucket type ~p of object ~p found on sink and is active.", [B, T]),
+                           do_put(B, Object)
```
bookshelfdave: anything you need to match against for `do_put`?

src/riak_repl_util.erl
```diff
@@ -108,44 +108,62 @@ get_partitions(Ring) ->
             riak_core_ring:all_owners(riak_core_ring:upgrade(Ring))]).
 
 do_repl_put(Object) ->
```
bookshelfdave, line 110: does `do_repl_put` need to return an explicit value?

bowrocker:
Updated for comments and rebased. New PR:

https://github.com/basho/riak_repl/pull/486

PR #482, opened 2013-12-12
Avoid arithmetic on folsom error tuples

cmeiklejohn:
Folsom errors happen, and they could make the repl stats process crash
because it tried to perform some calculations on those. This change
avoids those calculations when something other than a valid value is
returned.


cmeiklejohn:
@engelsanchez best way to test this?

cmeiklejohn:
Not tested, but visually :+1:'d by AT/DP/CM.

PR #486, opened 2013-12-13

bowrocker:
#### Overview

###### Pros
###### Cons
#### Configuration

| Source | Sink | Result |
| --- | :-: | --: |
| {2,1} | {2,1} | "default" and user defined typed buckets replicated |
| {2,1} | {2,0} or less | only "default" buckets replicated |
#### Dependencies

#### Original PR

https://github.com/basho/riak_repl/pull/477

### DIFF
--- a/src/riak_repl2_rt.erl
+++ b/src/riak_repl2_rt.erl
@@ -143,15 +143,16 @@ postcommit(RObj) ->
         %% on what the RT source and sink negotiated as the common version).
         Objects0 when is_list(Objects0) ->
             Objects = Objects0 ++ [RObj],
+            Meta = set_bucket_meta(RObj),
             BinObjs = riak_repl_util:to_wire(w1, Objects),
             %% try the proxy first, avoids race conditions with unregister()
             %% during shutdown
             case whereis(riak_repl2_rtq_proxy) of
                 undefined ->
-                    riak_repl2_rtq:push(length(Objects), BinObjs);
+                    riak_repl2_rtq:push(length(Objects), BinObjs, Meta);
                 _ ->
                     %% we're shutting down and repl is stopped or stopping...
-                    riak_repl2_rtq_proxy:push(length(Objects), BinObjs)
+                    riak_repl2_rtq_proxy:push(length(Objects), BinObjs, Meta)
             end;
         cancel -> % repl helper callback requested not to send over realtime
             ok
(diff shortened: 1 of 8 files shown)


src/riak_repl2_rt.erl
```diff
@@ -217,3 +216,18 @@ do_ring_trans(F, A) ->
         ER ->
             ER
     end.
+
+set_bucket_meta(Obj) ->
+    M = orddict:new(),
+    case riak_object:bucket(Obj) of
+        {Type, _B} ->
+            AllProps = riak_core_bucket_type:get(Type),
+            PropsHash = erlang:phash2(proplists:delete(claimant, AllProps)),
```
jrwest: this is probably why long term its better to have a "whitelist" of properties we include in the hash rather than a "blacklist" of ones we don't.

src/riak_repl2_rtsink_conn.erl
```diff
@@ -397,6 +406,28 @@ schedule_reactivate_socket(State = #state{transport = T,
 get_reactivate_socket_interval() ->
     app_helper:get_env(riak_repl, reactivate_socket_interval_millis, ?REACTIVATE_SOCK_INT_MILLIS).
 
+write_object(Meta) ->
+    case orddict:fetch(typed_bucket, Meta) of 
+        false -> true;
+        true ->
+            BucketType = orddict:fetch(type, Meta),
+            case riak_core_bucket_type:get(BucketType) of
+            undefined ->
```
jrwest: EWEIRDWHITESPACE ;)

src/riak_repl2_rtsink_conn.erl
```diff
@@ -397,6 +406,28 @@ schedule_reactivate_socket(State = #state{transport = T,
 get_reactivate_socket_interval() ->
     app_helper:get_env(riak_repl, reactivate_socket_interval_millis, ?REACTIVATE_SOCK_INT_MILLIS).
 
+write_object(Meta) ->
+    case orddict:fetch(typed_bucket, Meta) of 
+        false -> true;
+        true ->
+            BucketType = orddict:fetch(type, Meta),
+            case riak_core_bucket_type:get(BucketType) of
```
jrwest, line 414: seems like it would be nice to refactor this and the sink side into a common function if possible. That way we ensure we are checking the same properties on both sides when we change things.

src/riak_repl2_rt.erl
bowrocker: Agreed -- we should discuss once we have an idea of what those properties are, then we get do it that way, I'd greatly prefer that.

jrwest:
I also wonder if we should make the hashes of the properties exposed to the user so that when things aren't replicating they can check whether or not they in fact match or not on both sides.

src/riak_repl2_rt.erl
jrwest: we sort of need to discuss soon don't we :)

bowrocker: Yeah, but we can probably use what's here for now, then open an issue to fix it...

jrwest: well i guess it depends. its getting kinda of late for the open an issue thing (that basically means its a bug for 2.0 and a fix in 2.0.1 or 2.1). One immediate problem I see is if a user has properties that are only meaninful to them -- bucket (type) properties support arbitrary key/value pairs. By not limiting this list we open ourselves to all sorts of weirdness there. 

I don't think it would be too bad to gather up a list of things that we should check. Previously repl did no checking iirc so its kind of hard for us to get it wrong isn't it? Off the top of my head the ones that probably matter are:
- `consistent`
- `datatype`
- `n_val`
- `allow_mult`
- `last_write_wins`

Those are the ones I can think of that affect how data is stored. We shouldn't require that users have the same default `w` or `r` values on both sides, etc.

bowrocker:
Closing -- new PR here:

https://github.com/basho/riak_repl/pull/490

PR #487, opened 2013-12-13
added a v2 deprecation lager:warning

bookshelfdave:
the wording of the message might change before the official release, but here's the first cut.


bowrocker:
Testing

[x] manual verification of code
[x ] riak_test: https://gist.github.com/bowrocker/7951511
[ ] Dialyzer
[ ] XRef
[ ] Coverage reports

:+1:

PR #488, opened 2013-12-13
Integrate fullsync with the background manager

buddhisthead:
NOTE: DO NOT MERGE YET. requires two additional PR from riak_core...

The riak_core background manager PR is here: https://github.com/basho/riak_core/pull/475

The riak_core integration with handoff and AAE is here: https://github.com/basho/riak_core/pull/484

The background manager allows subsystems to coordinate with locks (and tokens) in a non-blocking API when desired. In our case, we've seen several customer support issues related to a production cluster being overloaded by the combined weight of MDC fullsync replication in combination with either handoff, or AAE tree rebuilds (or all three!).

Each of these subsystems is being taught, for 2.0, to "take" a lock on the kv vnode/partition before it does a fold over that partition. Locks are returned when the process dies or exists normally. Each subsystem (at this point in time) is coded to keep working on other partitions if one partition is denied a lock via the return value of max_concurrency, which means progress is made even when one partition is busy.

The fullsync coordinator is tolerant of fullsync source processes "crashing" if they return the right kind of error. This PR adds another error code that signals it's busy; the coodinator will just move that partition to the end of the TODO list and try it later. This applies to both key_list and aae strategies because ultimately they both fold over the vnode after creating a bloom filter.

Use of this feature can be bypassed in production. Besides a global skill switch on the background manager itself, the MDC configuration supports a new parameter: fullsync.skip_bg_manager, which defaults to false. Setting it to true will not call the background manager and proceed like it used to.

### DIFF
--- a/priv/riak_repl.schema
+++ b/priv/riak_repl.schema
@@ -127,10 +127,10 @@
 %% @doc By default, fullsync replication will try to coordinate with other
 %% riak subsystems that may be contending for the same resources. This will help
 %% to prevent system response degradation under times of heavy load from multiple
-%% background tasks. To disable background coordination, set this parameter to true.
+%% background tasks. To disable background coordination, set this parameter to false.
 %% Enterprise 2.0+.
-{mapping, "mdc.fullsync.skip_bg_manager", "riak_repl.fullsync_skip_background_manager", [
+{mapping, "mdc.fullsync.use_bg_manager", "riak_repl.fullsync_use_background_manager", [
     {datatype, {enum, [true, false]}},
     {level, advanced},
-    {default, false}
+    {default, true}
 ]}.
--- a/priv/riak_repl.schema
+++ b/priv/riak_repl.schema
@@ -131,5 +131,6 @@
 %% Enterprise 2.0+.
 {mapping, "mdc.fullsync.skip_bg_manager", "riak_repl.fullsync_skip_background_manager", [
     {datatype, {enum, [true, false]}},
+    {level, advanced},
     {default, false}
 ]}.
--- a/priv/riak_repl.schema
+++ b/priv/riak_repl.schema
@@ -123,3 +123,13 @@
     {datatype, {duration, s}},
     {default, "15s"}
 ]}.
+
+%% @doc By default, fullsync replication will try to coordinate with other
+%% riak subsystems that may be contending for the same resources. This will help
+%% to prevent system response degradation under times of heavy load from multiple
+%% background tasks. To disable background coordination, set this parameter to true.
+%% Enterprise 2.0+.
+{mapping, "mdc.fullsync.skip_bg_manager", "riak_repl.fullsync_skip_background_manager", [
+    {datatype, {enum, [true, false]}},
+    {default, false}
+]}.
--- a/src/riak_repl2_fssource.erl
+++ b/src/riak_repl2_fssource.erl
@@ -1,5 +1,6 @@
 -module(riak_repl2_fssource).
 -include("riak_repl.hrl").
+-include_lib("riak_core/include/riak_core_locks.hrl").
 
 -behaviour(gen_server).
 %% API
@@ -62,7 +63,16 @@ init([Partition, IP]) ->
     OurCaps = decide_our_caps(DefaultStrategy),
     SupportedStrategy = proplists:get_value(strategy, OurCaps, DefaultStrategy),
 
-    connect(IP, SupportedStrategy, Partition).
+    %% Possibly try to obtain the per-vnode lock before connecting.
+    %% If we return error, we expect the coordinator to start us again later.
+    case maybe_get_vnode_lock(Partition) of
+        ok ->
+            %% got the lock, or ignored it.
+            connect(IP, SupportedStrategy, Partition);
+        {error, Reason} ->
+            %% the vnode is probably busy. Try again later.
+            {stop, Reason}
+    end.
 
 handle_call({connected, Socket, Transport, _Endpoint, Proto, Props},
             _From, State=#state{ip=IP, partition=Partition, strategy=DefaultStrategy}) ->
@@ -298,4 +308,37 @@ connect(IP, Strategy, Partition) ->
         {error, Reason}->
             lager:warning("Error connecting to remote"),
             {stop, Reason}
+    end.
+
+%% @private
+%% @doc Return true iff neither of the "skip background manager" configuration
+%%      settings are defined as anything other than false. 'skip_background_manager'
+%%      turns off all use of bg-mgr. 'fullsync_skip_background_manager' only stops
+%%      fullsync from using bg-mgr.
+-spec use_bg_mgr() -> boolean().
+use_bg_mgr() ->
+    %% note we're tolerant here of any non-boolean value as well. Iff it's 'false', we don't skip.
+    GlobalSkip = app_helper:get_env(riak_core, skip_background_manager, false) =/= false,
+    FSSkip = app_helper:get_env(riak_repl, fullsync_skip_background_manager, false) =/= false,
+    not (GlobalSkip orelse FSSkip).
+
+%% @private
+%% @doc Unless skipping the background manager, try to acquire the per-vnode lock.
+%%      Sets our task meta-data in the lock as 'aae_rebuild', which is useful for
+%%      seeing what's holding the lock via @link riak_core_background_mgr:ps/0.
+-spec maybe_get_vnode_lock(SrcPartition::integer()) -> ok | {error, Reason::term()}.
+maybe_get_vnode_lock(SrcPartition) ->
+    Lock = ?VNODE_LOCK(riak_kv_vnode, SrcPartition),
+    case use_bg_mgr() of
+        true  ->
+            case riak_core_bg_manager:get_lock(Lock, self(), [{task, repl_fullsync}]) of
+                {ok, _Ref} ->
+                    ok;
+                max_concurrency ->
+                    Reason = {max_concurrency, Lock},
+                    {error, Reason}
+            end;
+        false ->
+            lager:debug("AAE tree rebuild is skipping the background manager vnode lock: ~p", [Lock]),
+            ok
     end.
--- a/src/riak_repl2_fssource.erl
+++ b/src/riak_repl2_fssource.erl
@@ -1,6 +1,6 @@
 -module(riak_repl2_fssource).
 -include("riak_repl.hrl").
--include_lib("riak_core/include/riak_core_locks.hrl").
+-include_lib("riak_kv/include/riak_kv_vnode.hrl").
 
 -behaviour(gen_server).
 %% API
@@ -311,26 +311,14 @@ connect(IP, Strategy, Partition) ->
     end.
 
 %% @private
-%% @doc Return true iff neither of the "skip background manager" configuration
-%%      settings are defined as anything other than false. 'skip_background_manager'
-%%      turns off all use of bg-mgr. 'fullsync_skip_background_manager' only stops
-%%      fullsync from using bg-mgr.
--spec use_bg_mgr() -> boolean().
-use_bg_mgr() ->
-    %% note we're tolerant here of any non-boolean value as well. Iff it's 'false', we don't skip.
-    GlobalSkip = app_helper:get_env(riak_core, skip_background_manager, false) =/= false,
-    FSSkip = app_helper:get_env(riak_repl, fullsync_skip_background_manager, false) =/= false,
-    not (GlobalSkip orelse FSSkip).
-
-%% @private
 %% @doc Unless skipping the background manager, try to acquire the per-vnode lock.
-%%      Sets our task meta-data in the lock as 'aae_rebuild', which is useful for
+%%      Sets our task meta-data in the lock as 'repl_fullsync', which is useful for
 %%      seeing what's holding the lock via @link riak_core_background_mgr:ps/0.
 -spec maybe_get_vnode_lock(SrcPartition::integer()) -> ok | {error, Reason::term()}.
 maybe_get_vnode_lock(SrcPartition) ->
-    Lock = ?VNODE_LOCK(riak_kv_vnode, SrcPartition),
-    case use_bg_mgr() of
+    case riak_core_bg_manager:use_bg_mgr(riak_repl, fullsync_use_background_manager) of
         true  ->
+            Lock = ?VNODE_LOCK(SrcPartition),
             case riak_core_bg_manager:get_lock(Lock, self(), [{task, repl_fullsync}]) of
                 {ok, _Ref} ->
                     ok;
@@ -339,6 +327,5 @@ maybe_get_vnode_lock(SrcPartition) ->
                     {error, Reason}
             end;
         false ->
-            lager:debug("AAE tree rebuild is skipping the background manager vnode lock: ~p", [Lock]),
             ok
     end.
--- a/src/riak_repl2_fssource.erl
+++ b/src/riak_repl2_fssource.erl
@@ -65,7 +65,7 @@ init([Partition, IP]) ->
 
     %% Possibly try to obtain the per-vnode lock before connecting.
     %% If we return error, we expect the coordinator to start us again later.
-    case maybe_get_vnode_lock(Partition) of
+    case riak_repl_util:maybe_get_vnode_lock(Partition) of
         ok ->
             %% got the lock, or ignored it.
             connect(IP, SupportedStrategy, Partition);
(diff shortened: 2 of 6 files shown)


cmeiklejohn:
Just to be clear, the only branch you need to pull on `riak_core` to get this to build is `cet-bg-mgr-vnode-lock`, since one has been merged into the other.

cmeiklejohn:
However, it appears that these branches are not up to date with the extraction of `riak_core_connection` from `riak_core` to `riak_repl`.

```
==> rel (generate)
ERROR: generate failed while processing /Users/cmeiklejohn/Repositories/riak_ee/rel: {'EXIT',{{badmatch,{error,"Module riak_core_connection potentially included by two different applications: riak_core and riak_repl."}},
         [{rebar_reltool,generate,2,[]},
          {rebar_core,run_modules,4,[]},
          {rebar_core,execute,5,[]},
          {rebar_core,process_dir1,6,[]},
          {rebar_core,process_commands,2,[]},
          {rebar,main,1,[]},
          {escript,run,2,[{file,"escript.erl"},{line,747}]},
          {escript,start,1,[{file,"escript.erl"},{line,277}]}]}}
make: *** [dev1] Error 1
```

priv/riak_repl.schema
```diff
@@ -123,3 +123,13 @@
...
 ]}.
+
+%% @doc By default, fullsync replication will try to coordinate with other
+%% riak subsystems that may be contending for the same resources. This will help
+%% to prevent system response degradation under times of heavy load from multiple
+%% background tasks. To disable background coordination, set this parameter to true.
+%% Enterprise 2.0+.
+{mapping, "mdc.fullsync.skip_bg_manager", "riak_repl.fullsync_skip_background_manager", [
+    {datatype, {enum, [true, false]}},
+    {default, false}
```
cmeiklejohn: I thought we wanted to default this to being disabled, is that not correct?

src/riak_repl2_fssource.erl
```diff
@@ -299,3 +309,36 @@ connect(IP, Strategy, Partition) ->
...
+
+%% @private
+%% @doc Return true iff neither of the "skip background manager" configuration
+%%      settings are defined as anything other than false. 'skip_background_manager'
+%%      turns off all use of bg-mgr. 'fullsync_skip_background_manager' only stops
+%%      fullsync from using bg-mgr.
+-spec use_bg_mgr() -> boolean().
+use_bg_mgr() ->
+    %% note we're tolerant here of any non-boolean value as well. Iff it's 'false', we don't skip.
+    GlobalSkip = app_helper:get_env(riak_core, skip_background_manager, false) =/= false,
```
cmeiklejohn: Minor nitpick; this should probably be a function in `riak_core` that's exposed so all callers can call to `riak_core` to find out of the global skip is enabled.

cmeiklejohn:
Code looks fairly straightforward.  Going to wait to do further testing until the `riak_core` API stabilized and is merged as to avoid work duplication.

priv/riak_repl.schema
buddhisthead: No, we want to default to true. The switch is mostly as a failsafe to disable in production if needed in production. But I think the default should be "don't melt my cluster on full sync".

src/riak_repl2_fssource.erl
buddhisthead: Yeah, you're right. I was thinking that as I typed it again in the handoff code. At least for the "global" switch.

buddhisthead:
The API should be stable, especially with regard to the use by replication, handoff, and aee. But, thanks for your efforts so far.

buddhisthead:
I rebased on replication's develop branch right before I pushed the PR. Thanks move fast around release time :-)

priv/riak_repl.schema
seancribbs: Please add `{level, advanced}` if this is a setting that most installs won't change.

buddhisthead: Sounds good. Thanks.

buddhisthead: Addressed by 68d6c89https://github.com/basho/riak_repl/commit/68d6c890c1dbda89965eae74e879fc6f46bf5519
.

cmeiklejohn:
Ah, maybe I wasn't super clear, sorry!  The code that needs to be rebased is the related riak_core branch.

buddhisthead:
Hey Chris. Thanks for the note. I've rebased the riak_core branch against
develop and got your changes for the connection and service managers. The
core PR is pretty much wrapped up, so if you don't mind having another
look, I would appreciate it. Looking forward to your comments.

cmeiklejohn:
I'm on it first thing tomorrow.

cmeiklejohn:
@buddhisthead Which branch did you rebase?  Slightly confused as to which branch of core I should checkout considering basho/riak_core#481 appears to contain basho/riak_core#451 merged into it?

src/riak_repl2_fssource.erl
```diff
@@ -299,3 +309,36 @@ connect(IP, Strategy, Partition) ->
...
+    FSSkip = app_helper:get_env(riak_repl, fullsync_skip_background_manager, false) =/= false,
+    not (GlobalSkip orelse FSSkip).
+
+%% @private
+%% @doc Unless skipping the background manager, try to acquire the per-vnode lock.
+%%      Sets our task meta-data in the lock as 'aae_rebuild', which is useful for
```
Vagabond: But we set it as repl_fullsync here? Copy-paste-o?

src/riak_repl2_fssource.erl
```diff
@@ -62,7 +63,16 @@ init([Partition, IP]) ->
     OurCaps = decide_our_caps(DefaultStrategy),
     SupportedStrategy = proplists:get_value(strategy, OurCaps, DefaultStrategy),
 
-    connect(IP, SupportedStrategy, Partition).
+    %% Possibly try to obtain the per-vnode lock before connecting.
+    %% If we return error, we expect the coordinator to start us again later.
+    case maybe_get_vnode_lock(Partition) of
```
Vagabond: Do we need to do this in the worker process because the lock is tied to a pid? Can we check in the coordinator if the lock is available before spawning a worker?

src/riak_repl2_fssource.erl
```diff
@@ -299,3 +309,36 @@ connect(IP, Strategy, Partition) ->
...
+        true  ->
+            case riak_core_bg_manager:get_lock(Lock, self(), [{task, repl_fullsync}]) of
+                {ok, _Ref} ->
+                    ok;
+                max_concurrency ->
+                    Reason = {max_concurrency, Lock},
+                    {error, Reason}
+            end;
+        false ->
+            lager:debug("AAE tree rebuild is skipping the background manager vnode lock: ~p", [Lock]),
```
Vagabond: Also copy-paste-o.

src/riak_repl2_fssource.erl
Vagabond: With AAE fullsync, is the assumption that the AAE rebuild/exchange will have a lock, so we'll bail at this point?

Also, the aae_source looks like it retries forever if the remote tree is locked:

https://github.com/basho/riak_repl/blob/a6b19e75ce7d267d15c8452e098017fd9811a3ba/src/riak_repl_aae_source.erl#L180

If biredirectional fullsync is on, and a aae_source has a lock on the local tree on each side, will there not be a deadlock forever? And with this new lock manager not only fullsync will deadlock, but AAE and whatever else uses this lock manager will as well.

src/riak_repl2_fssource.erl
cmeiklejohn: This function is duplicated in `riak_kv` and should be moved to `riak_core`.

src/riak_repl2_fssource.erl
```diff
@@ -299,3 +309,36 @@ connect(IP, Strategy, Partition) ->
...
+    GlobalSkip = app_helper:get_env(riak_core, skip_background_manager, false) =/= false,
+    FSSkip = app_helper:get_env(riak_repl, fullsync_skip_background_manager, false) =/= false,
+    not (GlobalSkip orelse FSSkip).
+
+%% @private
+%% @doc Unless skipping the background manager, try to acquire the per-vnode lock.
+%%      Sets our task meta-data in the lock as 'aae_rebuild', which is useful for
+%%      seeing what's holding the lock via @link riak_core_background_mgr:ps/0.
+-spec maybe_get_vnode_lock(SrcPartition::integer()) -> ok | {error, Reason::term()}.
+maybe_get_vnode_lock(SrcPartition) ->
```
cmeiklejohn: This function is duplicated in `riak_kv` and should be moved to `riak_core`.

src/riak_repl2_fssource.erl
buddhisthead: We do need to take the lock with the pid of the worker, but we don't HAVE to do it from the worker. It would be a pain, though, to spawn the worker and then take the lock from somewhere else. We could, in theory, check that the lock is available, but then we'd have a race condition since it might get taken before we actually try.

buddhisthead: As for deadlock, it appears that AAE fullsync is susceptible to a deadlock as written today, during bi-directional replication if the two sides both start the same partition at the same time. Because the aae source retries the remote lock forever, the two sides could fail to make progress. The addition of the vnode lock could block other subprocesses from making progress on that vnode.

Since it is only the source side that takes the vnode lock, the problem would be avoided if the aae fullsync source gave up it's local lock when it got `already_locked` from the remote and re-queued the partition for later. Funny that we have a comment in the code that says that. However, I don't know if that's easy. It might be.

Since this is a problem in AAE full sync, as it exists today, I have filed issue https://github.com/basho/riak_repl/issues/499.  It's already a potentially serious problem because the hashtree lock would also block handoff and aee tree rebuilds, even without the addition of the new node lock using the background manager.

@jtuple may wish to comment as well since he helped develop the original aee_source

buddhisthead:
@cmeiklejohn In answer to your question about needed branches...
Please note that this PR now requires https://github.com/basho/riak_core/pull/484 (instead of 481) which has been rebased to contain the latest background manager.

priv/riak_repl.schema
```diff
@@ -123,3 +123,14 @@
...
+
+%% @doc By default, fullsync replication will try to coordinate with other
+%% riak subsystems that may be contending for the same resources. This will help
+%% to prevent system response degradation under times of heavy load from multiple
+%% background tasks. To disable background coordination, set this parameter to false.
+%% Enterprise 2.0+.
+{mapping, "mdc.fullsync.use_bg_manager", "riak_repl.fullsync_use_background_manager", [
+    {datatype, {enum, [true, false]}},
+    {level, advanced},
+    {default, true}
```
cmeiklejohn, line 135: Just double checking, didn't we want this disabled by default?

Vagabond:
Does the fssink not also need to grab a lock for the vnode on the sink cluster?

buddhisthead:
So, this turns out to be a nastier problem. On the source side, we know that both strategies will do a fold over the vnode, which is what the vnode_lock is trying to reduce contention on. So we can take the vnode_lock from the fssource level. On the sink side, it's not so easy. Only the keylist server needs to take the vnode_lock. But threading that is hard. We don't have the partition until we've already started up the specific strategy. We can't change that because the old key lister handles multiple partitions in a single process lifetime. And without adding new protocol versions, etc, we can't just add the get_lock into the key list server. It's too bad, because it's only the old keylist server that needs to take the lock - the AAE sink side does not fold over the vnode and therefore doesn't need to take the lock. However, to simplify the code immensely, we should probably just take the lock at the fssink level and pay the penalty. This would have the added benefit of solving the bi-directional MDC deadlock problem since we'll just immediately give up if we can't get the lock. We need to think if a deadlock could still occur in an N-way directional replication, where N > 2.

So, I'll proceed with this approach unless someone has a better suggestion.

priv/riak_repl.schema
buddhisthead, line 135: I feel that we want it enabled by default. But in testing, we'll make the final determination. I think @Vagabond 's new tests will help answer this question.

Vagabond:
Can you just not take the lock here, and if that fails just {stop, ...}

https://github.com/basho/riak_repl/blob/develop/src/riak_repl2_fssink.erl#L169

The work CMJ and I are doing on the AAE fullsync will ensure that the other side gets torn down as well.

buddhisthead:
Sadly, no. We don't have the Partition until after the keylist server has started and we get this message:

https://github.com/basho/riak_repl/blob/develop/src/riak_repl_keylist_server.erl#L159

Actually, now that I think about it. We can just take the lock in the keylist server because it won't cause any problems if we fail. We'll just retry if we handle the failure correctly. And we need to take the lock in the aee_sink as well or handle the already_locked retry issue to solve the deadlock problem.

Vagabond:
OK yes. It might get weird if people are using the keylist strategy and not using BNW repl, though?

buddhisthead:
Well, it's needed in both BNW and old world. Either way, it's good to take turns on the vnode folding. Someone could be using old repl with new AAE, handoff, etc.

buddhisthead:
89b8727 Addresses the need for locking on the sink. Note this is only needed in the keylist server since the AAE replication doesn't do a vnode fold on the sink side.

My remaining concern is that the fssink does not handle the case of it's worker stopping abnormally.

Vagabond:
After the AAE branch lands, the fssink should handle unclean worker exit.

buddhisthead:
Alright then, I'd like to get a +1 on this PR, but wait for your AAE work to complete before we merge.

cmeiklejohn:
The AAE works, as far as I know, is not slated to hit 2.0, so we need to figure out what we're doing here with merging this feature.

cmeiklejohn:
I'm :-1:

Not sure why, but this branch is failing replication2, and it might be related to it being stale.  Unsure.  However, replication2 passes on develop.

```
14:54:05.684 [warning] replication2 failed: {{assertEqual_failed,[{module,repl_util},{line,176},{expression,"Res"},{expected,ok},{value,{fail,false}}]},[{repl_util,'-start_and_wait_until_fullsync_complete/1-fun-3-',2,[{file,"tests/repl_util.erl"},{line,176}]},{repl_util,start_and_wait_until_fullsync_complete,1,[{file,"tests/repl_util.erl"},{line,176}]},{replication2,replication,3,[{file,"tests/replication2.erl"},{line,141}]},{replication2,confirm,0,[{file,"tests/replication2.erl"},{line,52}]}]}
14:54:05.685 [error]
================ replication2 failure stack trace =====================
{{assertEqual_failed,[{module,repl_util},
                      {line,176},
                      {expression,"Res"},
                      {expected,ok},
                      {value,{fail,false}}]},
 [{repl_util,'-start_and_wait_until_fullsync_complete/1-fun-3-',2,
             [{file,"tests/repl_util.erl"},{line,176}]},
  {repl_util,start_and_wait_until_fullsync_complete,1,
             [{file,"tests/repl_util.erl"},{line,176}]},
  {replication2,replication,3,[{file,"tests/replication2.erl"},{line,141}]},
  {replication2,confirm,0,[{file,"tests/replication2.erl"},{line,52}]}]}
=======================================================================
```

```
2013-12-23 14:51:51.276 [error] <0.20442.1> gen_fsm <0.20442.1> in state wait_for_partition terminated with reason: bad record state in riak_repl_keylist_server:terminate/3 line 572
```

src/riak_repl_keylist_server.erl
```diff
@@ -160,20 +160,31 @@ wait_for_partition({partition, Partition}, State=#state{work_dir=WorkDir}) ->
...
+                                                    kl_ref=KeyListRef, kl_fn=KeyListFn,
+                                                    partition=Partition,
+                                                    partition_start=os:timestamp(),
+                                                    stage_start=os:timestamp(),
+                                                    pending_acks=0, generator_paused=false,
+                                                    their_kl_fn=TheirKeyListFn,
+                                                    their_kl_fh=undefined}};
+        {error, Reason} ->
+            %% the vnode is probably busy. Quit all the way back.
+            {stop, Reason, state}
```
Vagabond: Lowercase S in State is making the 'badrecord' error happen.

buddhisthead: Yep. Good eyes.

buddhisthead:
This branch has been rebased off develop and force pushed to make merging easy peasy. The issue was that I had the sink-side lock in the wrong module. I needed to be in the client. The lock was getting taken twice, which of course didn't work well. With this change and the rebase (which was needed to make cuttlefish happy in the riak_test), replication2 now passes.

@cmeiklejohn please take one more pass when you have a chance. Thank you very much.

cmeiklejohn:
Reviewing this now...

cmeiklejohn:
I'm getting timeouts when starting realtime with `replication2` but I'm not sure if there is a legitimate problem yet.  I will continue to test.

```
================ replication2 failure stack trace =====================
{{assertEqual_failed,[{module,repl_util},
                      {line,176},
                      {expression,"Res"},
                      {expected,ok},
                      {value,{fail,false}}]},
 [{repl_util,'-start_and_wait_until_fullsync_complete/1-fun-3-',2,
             [{file,"tests/repl_util.erl"},{line,176}]},
  {repl_util,start_and_wait_until_fullsync_complete,1,
             [{file,"tests/repl_util.erl"},{line,176}]},
  {replication2,replication,3,[{file,"tests/replication2.erl"},{line,184}]},
  {replication2,confirm,0,[{file,"tests/replication2.erl"},{line,52}]}]}
=======================================================================

22:31:44.479 [notice] replication2 Test Run Complete
```

cmeiklejohn:
So, `replication2` is passing for me now.  I assume the original race was something to do with my laptop.  

I'm still a bit concerned that we don't have all of the deadlock cases covered, but given that `replication2` is passing, I'll give this a :+1:, however, I will strongly state that we absolutely, _no exceptions_ need an actual test to ensure this operates correctly, something that should have been done along with this PR.  

It appears that Chris is offline now, so given that, I'm going to merge it so it's guaranteed to make the `pre8` build tomorrow morning.

PR #490, opened 2013-12-16

bowrocker:
#### Overview

###### Pros
###### Cons
#### Configuration

| Source | Sink | Result |
| --- | :-: | --: |
| {3,0} | {3,0} | "default" and user defined typed buckets replicated |
| {3,0} | {2,0} or less | only "default" buckets replicated |
#### Dependencies

https://github.com/basho/riak_kv/tree/jdb-conditional-postcommit
#### Previous PR

https://github.com/basho/riak_repl/pull/486
#### Riak Test

https://github.com/basho/riak_test/pull/477

### DIFF
--- a/include/riak_repl.hrl
+++ b/include/riak_repl.hrl
@@ -32,6 +32,10 @@
 -define(RETRY_AAE_LOCKED_INTERVAL, 1000).
 -define(DEFAULT_FULLSYNC_STRATEGY, keylist). %% keylist | aae
 -define(LOG_USER_CMD(Msg, Params), lager:notice("[user] " ++ Msg, Params)).
+%% the following are keys for bucket types related meta data
+-define(BT_META_TYPED_BUCKET, typed_bucket).
+-define(BT_META_TYPE, bucket_type).
+-define(BT_META_PROPS_HASH, properties_hash_val).
 
 -type(ip_addr_str() :: string()).
 -type(ip_portnum() :: non_neg_integer()).
--- a/include/riak_repl.hrl
+++ b/include/riak_repl.hrl
@@ -57,7 +57,11 @@
 %% Also see analagous binary_version() in riak_object, which is carried
 %% inside the wire format in "BinObj". w0 implies v0. w1 imples v1.
 -type(wire_version() :: w0   %% simple term_to_binary() legacy encoding
-                      | w1). %% <<?MAGIC:8/integer, 1:8/integer,
+                      | w1   %% <<?MAGIC:8/integer, 1:8/integer,
+                             %% BLen:32/integer, B:BLen/binary,
+                             %% KLen:32/integer, K:KLen/binary, BinObj/binary>>.
+                      | w2). %% <<?MAGIC:8/integer, ?W2_VER:8/integer,
+                             %% TLen:32/integer, T:TLen/binary,
                              %% BLen:32/integer, B:BLen/binary,
                              %% KLen:32/integer, K:KLen/binary, BinObj/binary>>.
 
--- a/include/riak_repl.hrl
+++ b/include/riak_repl.hrl
@@ -56,6 +56,8 @@
 %% the to_wire() and from_wire() operations, see riak_repl_util.erl.
 %% Also see analagous binary_version() in riak_object, which is carried
 %% inside the wire format in "BinObj". w0 implies v0. w1 imples v1.
+%% w2 still uses v1, but supports encoding the type information
+%% for bucket types.
 -type(wire_version() :: w0   %% simple term_to_binary() legacy encoding
                       | w1   %% <<?MAGIC:8/integer, 1:8/integer,
                              %% BLen:32/integer, B:BLen/binary,
--- a/src/riak_repl2_rt.erl
+++ b/src/riak_repl2_rt.erl
@@ -151,10 +151,8 @@ postcommit(RObj) ->
             
             BinObjs = case orddict:fetch(?BT_META_TYPED_BUCKET, Meta) of
                 false ->
-                    lager:debug("encoding w1 format object"),
                     riak_repl_util:to_wire(w1, Objects);
                 true ->
-                    lager:debug("encoding w2 format object"),
                     riak_repl_util:to_wire(w2, Objects)
             end,
             %% try the proxy first, avoids race conditions with unregister()
@@ -234,11 +232,9 @@ set_bucket_meta(Obj) ->
     case riak_object:bucket(Obj) of
         {Type, _B} ->
             PropsHash = riak_repl_util:get_bucket_props_hash(riak_core_bucket_type:get(Type)),
-            lager:debug("typed bucket, setting meta data: Type:~p, HashProps:~p", [Type, PropsHash]),
             M1 = orddict:store(?BT_META_TYPED_BUCKET, true, M),
             M2 = orddict:store(?BT_META_TYPE, Type, M1),
             orddict:store(?BT_META_PROPS_HASH, PropsHash, M2);
         _B -> 
-            lager:debug("un-typed bucket, setting typed_bucket to false"),
             orddict:store(?BT_META_TYPED_BUCKET, false, M)
     end.
--- a/src/riak_repl2_rt.erl
+++ b/src/riak_repl2_rt.erl
@@ -2,6 +2,8 @@
 %% Copyright (c) 2007-2012 Basho Technologies, Inc.  All Rights Reserved.
 -module(riak_repl2_rt).
 
+-include("riak_repl.hrl").
+
 %% @doc Realtime replication
 %%
 %% High level responsibility...
@@ -15,6 +17,8 @@
          terminate/2, code_change/3]).
 
 -define(SERVER, ?MODULE).
+
+
 -record(state, {sinks = []}).
 
 %% API - is there any state? who watches ring events?
@@ -223,10 +227,10 @@ set_bucket_meta(Obj) ->
         {Type, _B} ->
             PropsHash = riak_repl_util:get_bucket_props_hash(riak_core_bucket_type:get(Type)),
             lager:debug("typed bucket, setting meta data: Type:~p, HashProps:~p", [Type, PropsHash]),
-            M1 = orddict:store(typed_bucket, true, M),
-            M2 = orddict:store(type, Type, M1),
-            orddict:store(props_hash, PropsHash, M2);
+            M1 = orddict:store(?BT_META_TYPED_BUCKET, true, M),
+            M2 = orddict:store(?BT_META_TYPE, Type, M1),
+            orddict:store(?BT_META_PROPS_HASH, PropsHash, M2);
         _B -> 
             lager:debug("un-typed bucket, setting typed_bucket to false"),
-            orddict:store(typed_bucket, false, M)
+            orddict:store(?BT_META_TYPED_BUCKET, false, M)
     end.
--- a/src/riak_repl2_rt.erl
+++ b/src/riak_repl2_rt.erl
@@ -143,15 +143,16 @@ postcommit(RObj) ->
         %% on what the RT source and sink negotiated as the common version).
         Objects0 when is_list(Objects0) ->
             Objects = Objects0 ++ [RObj],
+            Meta = set_bucket_meta(RObj),
             BinObjs = riak_repl_util:to_wire(w1, Objects),
             %% try the proxy first, avoids race conditions with unregister()
             %% during shutdown
             case whereis(riak_repl2_rtq_proxy) of
                 undefined ->
-                    riak_repl2_rtq:push(length(Objects), BinObjs);
+                    riak_repl2_rtq:push(length(Objects), BinObjs, Meta);
                 _ ->
                     %% we're shutting down and repl is stopped or stopping...
-                    riak_repl2_rtq_proxy:push(length(Objects), BinObjs)
+                    riak_repl2_rtq_proxy:push(length(Objects), BinObjs, Meta)
             end;
         cancel -> % repl helper callback requested not to send over realtime
             ok
@@ -208,12 +209,24 @@ terminate(_Reason, _State) ->
 code_change(_OldVsn, State, _Extra) ->
     {ok, State}.
 
-
-
 do_ring_trans(F, A) ->
     case riak_core_ring_manager:ring_trans(F, A) of
         {ok, _} ->
             ok;
         ER ->
             ER
+    end.
+
+set_bucket_meta(Obj) ->
+    M = orddict:new(),
+    case riak_object:bucket(Obj) of
+        {Type, _B} ->
+            PropsHash = riak_repl_util:get_bucket_props_hash(riak_core_bucket_type:get(Type)),
+            lager:debug("typed bucket, setting meta data: Type:~p, HashProps:~p", [Type, PropsHash]),
+            M1 = orddict:store(typed_bucket, true, M),
+            M2 = orddict:store(type, Type, M1),
+            orddict:store(props_hash, PropsHash, M2);
+        _B -> 
+            lager:debug("un-typed bucket, setting typed_bucket to false"),
+            orddict:store(typed_bucket, false, M)
     end.
--- a/src/riak_repl2_rt.erl
+++ b/src/riak_repl2_rt.erl
@@ -151,11 +151,11 @@ postcommit(RObj) ->
             
             BinObjs = case orddict:fetch(?BT_META_TYPED_BUCKET, Meta) of
                 false ->
-                     lager:debug("encoding w1 format object"),
-                     riak_repl_util:to_wire(w1, Objects);
+                    lager:debug("encoding w1 format object"),
+                    riak_repl_util:to_wire(w1, Objects);
                 true ->
-                     lager:debug("encoding w2 format object"),
-                     riak_repl_util:to_wire(w2, Objects)
+                    lager:debug("encoding w2 format object"),
+                    riak_repl_util:to_wire(w2, Objects)
             end,
             %% try the proxy first, avoids race conditions with unregister()
             %% during shutdown
--- a/src/riak_repl2_rt.erl
+++ b/src/riak_repl2_rt.erl
@@ -148,7 +148,15 @@ postcommit(RObj) ->
         Objects0 when is_list(Objects0) ->
             Objects = Objects0 ++ [RObj],
             Meta = set_bucket_meta(RObj),
-            BinObjs = riak_repl_util:to_wire(w1, Objects),
+            
+            BinObjs = case orddict:fetch(?BT_META_TYPED_BUCKET, Meta) of
+                false ->
+                     lager:debug("encoding w1 format object"),
+                     riak_repl_util:to_wire(w1, Objects);
+                true ->
+                     lager:debug("encoding w2 format object"),
+                     riak_repl_util:to_wire(w2, Objects)
+            end,
             %% try the proxy first, avoids race conditions with unregister()
             %% during shutdown
             case whereis(riak_repl2_rtq_proxy) of
--- a/src/riak_repl2_rtsink_conn.erl
+++ b/src/riak_repl2_rtsink_conn.erl
@@ -246,7 +246,6 @@ recv(TcpBin, State = #state{transport = T, socket = S}) ->
             send_heartbeat(T, S),
             recv(Cont, State#state{hb_last = os:timestamp()});
         {ok, {objects_and_meta, {Seq, BinObjs, Meta}}, Cont} ->
-            lager:info("get objects_and_meta"),
             recv(Cont, do_write_objects(Seq, {BinObjs, Meta}, State));
         {error, Reason} ->
             %% TODO: Log Something bad happened
--- a/src/riak_repl2_rtsink_conn.erl
+++ b/src/riak_repl2_rtsink_conn.erl
@@ -56,8 +56,8 @@
 
 %% Register with service manager
 sync_register_service() ->
-    %% version {2,1} supports typed bucket replication
-    ProtoPrefs = {realtime,[{2,1}, {2,0}, {1,4}, {1,1}, {1,0}]},
+    %% version {3,0} supports typed bucket replication
+    ProtoPrefs = {realtime,[{3,0}, {2,0}, {1,4}, {1,1}, {1,0}]},
     TcpOptions = [{keepalive, true}, % find out if connection is dead, this end doesn't send
                   {packet, 0},
                   {nodelay, true}],
@@ -246,6 +246,7 @@ recv(TcpBin, State = #state{transport = T, socket = S}) ->
             send_heartbeat(T, S),
             recv(Cont, State#state{hb_last = os:timestamp()});
         {ok, {objects_and_meta, {Seq, BinObjs, Meta}}, Cont} ->
+            lager:info("get objects_and_meta"),
             recv(Cont, do_write_objects(Seq, {BinObjs, Meta}, State));
         {error, Reason} ->
             %% TODO: Log Something bad happened
--- a/src/riak_repl2_rtsink_conn.erl
+++ b/src/riak_repl2_rtsink_conn.erl
@@ -54,9 +54,9 @@
                 cont = <<>>       %% Continuation from previous TCP buffer
                }).
 
-%% @doc Register with service manager
+%% Register with service manager
 sync_register_service() ->
-    ProtoPrefs = {realtime,[{2,0}, {1,4}, {1,1}, {1,0}]},
+    ProtoPrefs = {realtime,[{2,1}, {2,0}, {1,4}, {1,1}, {1,0}]},
     TcpOptions = [{keepalive, true}, % find out if connection is dead, this end doesn't send
                   {packet, 0},
                   {nodelay, true}],
@@ -258,7 +258,7 @@ make_donefun({Binary, Meta}, Me, Ref, Seq) ->
         gen_server:cast(Me, {ack, Ref, Seq, Skips}),
         maybe_push(Binary, Meta)
     end,
-    {Done, Binary};
+    {Done, Binary, Meta};
 make_donefun(Binary, Me, Ref, Seq) when is_binary(Binary) ->
     Done = fun() ->
         gen_server:cast(Me, {ack, Ref, Seq, 0})
@@ -285,8 +285,17 @@ do_write_objects(Seq, BinObjsMeta, State = #state{max_pending = MaxPending,
                                               acked_seq = AckedSeq,
                                               ver = Ver}) ->
     Me = self(),
-    {DoneFun, BinObjs} = make_donefun(BinObjsMeta, Me, Ref, Seq),
-    riak_repl2_rtsink_helper:write_objects(Helper, BinObjs, DoneFun, Ver),
+    case make_donefun(BinObjsMeta, Me, Ref, Seq) of
+        {DoneFun, BinObjs, Meta} ->
+            case write_object(Meta) of
+            true ->
+                riak_repl2_rtsink_helper:write_objects(Helper, BinObjs, DoneFun, Ver);
+            false ->
+                lager:info("Bucket is of a type that are not equal on both the source and sink; not writing object.")
+            end;
+        {DoneFun, BinObjs} ->
+            riak_repl2_rtsink_helper:write_objects(Helper, BinObjs, DoneFun, Ver)
+    end,
     State2 = case AckedSeq of
                  undefined ->
                      %% Handle first received sequence number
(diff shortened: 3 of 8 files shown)


src/riak_repl2_rt.erl
```diff
@@ -217,3 +216,17 @@ do_ring_trans(F, A) ->
...
             ER
     end.
+
+set_bucket_meta(Obj) ->
+    M = orddict:new(),
+    case riak_object:bucket(Obj) of
+        {Type, _B} ->
+            PropsHash = riak_repl_util:get_bucket_props_hash(riak_core_bucket_type:get(Type)),
+            lager:debug("typed bucket, setting meta data: Type:~p, HashProps:~p", [Type, PropsHash]),
+            M1 = orddict:store(typed_bucket, true, M),
```
bookshelfdave: it might be nice to define `typed_bucket` as a constant.

src/riak_repl2_rt.erl
```diff
@@ -217,3 +216,17 @@ do_ring_trans(F, A) ->
...
+            M2 = orddict:store(type, Type, M1),
```
bookshelfdave: or anything that's a key to the orddict

src/riak_repl2_rtsink_conn.erl
```diff
@@ -54,9 +54,9 @@
                 cont = <<>>       %% Continuation from previous TCP buffer
                }).
 
-%% @doc Register with service manager
+%% Register with service manager
 sync_register_service() ->
-    ProtoPrefs = {realtime,[{2,0}, {1,4}, {1,1}, {1,0}]},
+    ProtoPrefs = {realtime,[{2,1}, {2,0}, {1,4}, {1,1}, {1,0}]},
```
bookshelfdave: it might be worth commenting on the diff between 2.0 + 2.1 inline

src/riak_repl2_rtsink_conn.erl
```diff
@@ -285,8 +285,17 @@ do_write_objects(Seq, BinObjsMeta, State = #state{max_pending = MaxPending,
...
     Me = self(),
-    {DoneFun, BinObjs} = make_donefun(BinObjsMeta, Me, Ref, Seq),
-    riak_repl2_rtsink_helper:write_objects(Helper, BinObjs, DoneFun, Ver),
+    case make_donefun(BinObjsMeta, Me, Ref, Seq) of
+        {DoneFun, BinObjs, Meta} ->
+            case write_object(Meta) of
+            true ->
+                riak_repl2_rtsink_helper:write_objects(Helper, BinObjs, DoneFun, Ver);
+            false ->
+                lager:info("Bucket is of a type that are not equal on both the source and sink; not writing object.")
```
bookshelfdave: this message could get very spammy in the logs, also 

grammar? "Bucket is of a type that _is_ not equal on both the source and sink"

src/riak_repl2_rtsink_conn.erl
```diff
@@ -285,8 +285,17 @@ do_write_objects(Seq, BinObjsMeta, State = #state{max_pending = MaxPending,
...
+            end;
+        {DoneFun, BinObjs} ->
+            riak_repl2_rtsink_helper:write_objects(Helper, BinObjs, DoneFun, Ver)
```
bookshelfdave, line 299: I'm assuming this is for previous versions of repl. If so, can you comment it as such?

src/riak_repl2_rt.erl
bowrocker: Is riak_repl.hrl a good place to add these definitions? I don't see a common place...is it safe to add includes of that file?

bookshelfdave: riak_repl.hrl is fine

src/riak_repl2_rtsink_conn.erl
```diff
@@ -397,6 +406,27 @@ schedule_reactivate_socket(State = #state{transport = T,
...
+            case riak_core_bucket_type:get(BucketType) of
+                undefined ->
+                    lager:debug("No properties found for bucket type:~p", [BucketType]),    
+	            false;
+	        {error, _T} ->
+                    lager:debug("No properties found for bucket type:~p", [BucketType]),
+	            false;
+                AllProps ->
+                    Sink = riak_repl_util:get_bucket_props_hash(AllProps),
+                    Source = orddict:fetch(props_hash, Meta),
```
bookshelfdave: `type`, `props_hash`, `typed_bucket` constants here etc

src/riak_repl2_rtsink_conn.erl
```diff
@@ -397,6 +406,27 @@ schedule_reactivate_socket(State = #state{transport = T,
...
+write_object(Meta) ->
+    case orddict:fetch(typed_bucket, Meta) of
+        false -> true;
+        true ->
+            BucketType = orddict:fetch(type, Meta),
+            case riak_core_bucket_type:get(BucketType) of
+                undefined ->
+                    lager:debug("No properties found for bucket type:~p", [BucketType]),    
+	            false;
+	        {error, _T} ->
```
bookshelfdave: indentation looks screwed up here, and makes it harder to read the multiple case clauses

src/riak_repl2_rtsink_conn.erl
bowrocker: Should it be a debug then? How else we the customer know what's going on?

bookshelfdave: can you keep a list of "un-replable" buckets and expose it as a stat?

src/riak_repl2_rtsource_helper.erl
```diff
@@ -124,13 +124,28 @@ maybe_send(Transport, Socket, QEntry, State) ->
...
-            State2
+            case State#state.proto of 
+                Ver when Ver =:= {2,1} ->
+                    encode_and_send(QEntry, Remote, Transport, Socket, State);
+		_ ->
+                    case is_bucket_typed(Meta) of
+                        false ->
+                            encode_and_send(QEntry, Remote, Transport, Socket, State);
+                        true ->
+   	                    lager:info("Negotiated protocol version:~p does not support typed buckets, not sending")
```
bookshelfdave: this will also be spammy

src/riak_repl2_rtsink_conn.erl
bookshelfdave: although, that stat might get huge as well. hm

src/riak_repl2_rtsource_conn.erl
```diff
@@ -128,7 +128,7 @@ init([Remote]) ->
     % TODO: expand version list or remove comment when we no
     % longer support 1.3.1
     % prefered version list: [{2,0}, {1,5}, {1,1}, {1,0}]
-    ClientSpec = {{realtime,[{2,0}, {1,5}]}, {TcpOptions, ?MODULE, self()}},
+    ClientSpec = {{realtime,[{2,1}, {1,5}]}, {TcpOptions, ?MODULE, self()}},
```
bookshelfdave: was {2,0} removed?

src/riak_repl2_rtsource_helper.erl
```diff
@@ -124,13 +124,28 @@ maybe_send(Transport, Socket, QEntry, State) ->
...
             State;
         false ->
-            QEntry2 = merge_forwards_and_routed_meta(QEntry, Remote),
-            {Encoded, State2} = encode(QEntry2, State),
-            lager:debug("Forwarding to ~p with new data: ~p derived from ~p", [State#state.remote, QEntry2, QEntry]),
-            Transport:send(Socket, Encoded),
-            State2
+            case State#state.proto of 
+                Ver when Ver =:= {2,1} ->
+                    encode_and_send(QEntry, Remote, Transport, Socket, State);
```
bookshelfdave, line 129: `encode_and_send` returns a new state, do you need to propagate that new state?

src/riak_repl2_rtsource_conn.erl
bowrocker: ach -- so, I guess I didn't quite understand the versioning there, I should have asked. 

Is {{realtime,[{2,1},{2,0}, {1,5}]} correct then? 

Why is there only {2,0} and {1,5} on the source?

src/riak_repl2_rtsource_helper.erl
```diff
@@ -124,13 +124,28 @@ maybe_send(Transport, Socket, QEntry, State) ->
...
+            end,
+	    State
     end.
 
+encode_and_send(QEntry, Remote, Transport, Socket, State) ->
+    QEntry2 = merge_forwards_and_routed_meta(QEntry, Remote),
+    {Encoded, State2} = encode(QEntry2, State),
+    lager:debug("Forwarding to ~p with new data: ~p derived from ~p", [State#state.remote, QEntry2, QEntry]),
+    Transport:send(Socket, Encoded),
+    State2.
```
bookshelfdave, line 146: State2 seems to be thrown away above, do you even need the result of `{Encoded, State2} = encode(QEntry2, State),`?

src/riak_repl_util.erl
```diff
@@ -680,7 +685,10 @@ proxy_get_active() ->
 
 log_dropped_realtime_obj(Obj) ->
     DroppedKey = riak_object:key(Obj),
-    DroppedBucket = riak_object:bucket(Obj),
+    DroppedBucket = case riak_object:bucket(Obj) of
+        {_T, B} -> B;
```
bookshelfdave: in repl_helper_send, there's a match like so:
`{_T, B} -> B;`

and log_dropped_realtime_obj:
`{Bucket, _T} -> Bucket;`

one looks wrong.

src/riak_repl_util.erl
```diff
@@ -837,6 +861,27 @@ from_wire(w1, BinObjList) ->
...
+    case T of
+        <<>> -> 
+            lager:debug("Found zero-length type, decoding default bucket-type"),
+            riak_object:from_binary(B, K, BinObj);
+        _ -> 
+            lager:debug("Found non-zero-length type, decoding non-default bucket-type:~p", [T]),
+            riak_object:from_binary({T, B}, K, BinObj)
+    end;
+from_wire(X) when is_binary(X) ->
+    lager:error("unknown wire format: ~p", [X]),
```
bookshelfdave: how about `"An unknown replication wire format has been detected: ~p"`

src/riak_repl_util.erl
```diff
@@ -894,6 +929,12 @@ peer_wire_format(Peer) ->
             w0
     end.
 
+get_bucket_props_hash(Props) ->
+   PB = [{Prop, proplists:get_value(Prop, Props)} || Prop <- ?BUCKET_TYPES_PROPS],
+   lager:debug("Bucket types props: ~p", [PB]),
+   erlang:phash2(PB).
```
bookshelfdave, line 944: a comment on why you need phash2 here would be nice

src/riak_repl_util.erl
```diff
@@ -795,13 +803,24 @@ encode_obj_msg(V, {Cmd, RObj}) ->
             term_to_binary({Cmd, BObj})
     end.
 
-%% @doc Create a new binary wire formatted replication blob, complete with
-%%      bucket and key for reconstruction on the other end. BinObj should be
+%% @doc Create binary wire formatted replication blob for riak 2.0+, complete with
+%%      possible type, bucket and key for reconstruction on the other end. BinObj should be
 %%      in the new format as obtained from riak_object:to_binary(v1, RObj).
-new_w1(B, K, BinObj) when is_binary(B), is_binary(K), is_binary(BinObj) ->
```
buddhisthead, line 801: I feel that the wire version must be increased to w2 and the version increased to denote the new encoding of the Bucket Type in the object. That way, any code that encounters it can either work or safely ignore it and log it. Perhaps even better would be add the bucket type as a property item in a list and say that for wire version 2, we now support meta data along with the bucket and key. This might avoid future wire changes and make the systems more flexible. Let me know if you want help with what that would look like in our encoded binary.

include/riak_repl.hrl
```diff
@@ -53,7 +57,11 @@
 %% Also see analagous binary_version() in riak_object, which is carried
 %% inside the wire format in "BinObj". w0 implies v0. w1 imples v1.
 -type(wire_version() :: w0   %% simple term_to_binary() legacy encoding
```
bookshelfdave, line 61: can you add a comment on what w0/w1 + w2 are?

src/riak_repl2_rtsink_conn.erl
```diff
@@ -245,6 +246,7 @@ recv(TcpBin, State = #state{transport = T, socket = S}) ->
             send_heartbeat(T, S),
             recv(Cont, State#state{hb_last = os:timestamp()});
         {ok, {objects_and_meta, {Seq, BinObjs, Meta}}, Cont} ->
+            lager:info("get objects_and_meta"),
```
bookshelfdave: cruft

src/riak_repl2_rtsource_helper.erl
```diff
@@ -124,13 +124,28 @@ maybe_send(Transport, Socket, QEntry, State) ->
...
+   	                    lager:debug("Negotiated protocol version:~p does not support typed buckets, not sending"),
+                            State
+                    end
+            end
     end.
 
+encode_and_send(QEntry, Remote, Transport, Socket, State) ->
+    QEntry2 = merge_forwards_and_routed_meta(QEntry, Remote),
+    {Encoded, State2} = encode(QEntry2, State),
+    lager:info("Forwarding to ~p with new data: ~p derived from ~p", [State#state.remote, QEntry2, QEntry]),
```
bookshelfdave: can you dial this down to a lager:debug?

jtuple:
Pull-request that adds necessary support to K/V: basho/riak_kv#771

lordnull:
Ran riak_test against this, and passes. Code looks good. :+1: .

src/riak_repl2_rt.erl
```diff
@@ -217,3 +228,17 @@ do_ring_trans(F, A) ->
         ER ->
             ER
     end.
+
+set_bucket_meta(Obj) ->
+    M = orddict:new(),
+    case riak_object:bucket(Obj) of
+        {Type, _B} ->
+            PropsHash = riak_repl_util:get_bucket_props_hash(riak_core_bucket_type:get(Type)),
```
jrwest, line 234: just verifying: is it ok to assume here that the type always exists? `get/1` may return undefined and `get_bucket_props_hash/1` does not account for that case -- it will crash on the `get_value/2` call here: https://github.com/basho/riak_repl/blob/2dfb083d60dcfef40dcb3a6c561adbeb17f7d016/src/riak_repl_util.erl#L961

jrwest, line 234: i guess it is safe to assume because in order to have the `riak_object` the type must have existed/been activated in order to write it.

src/riak_repl_cs.erl
```diff
@@ -58,7 +58,11 @@ skip_common_bucket(Bucket) ->
 
 -spec skip_block_object(riak_object:riak_object()) -> boolean().
 skip_block_object(Object) ->
-    block_bucket(riak_object:bucket(Object)).
+    Bucket = case riak_object:bucket(Object) of
+        {_T, B} -> B;
+        B -> B
+    end,
+    block_bucket(Bucket).
```
jrwest, line 62: I'm not sure what `block_bucket/1` does but stripping of the type here caught my attention. Bucket `<<"a">>` (which is the same as `{<<"default">>, <<"a">>"}`) is not the same as `{<<"sometype">>, <<"a">>}`, so stripping the type and operating on just the bucket name might be a problem?

src/riak_repl_util.erl
```diff
@@ -108,13 +111,14 @@ get_partitions(Ring) ->
             riak_core_ring:all_owners(riak_core_ring:upgrade(Ring))]).
 
 do_repl_put(Object) ->
-    B = riak_object:bucket(Object),
+    B = case riak_object:bucket(Object) of
```
jrwest: the more I look at it, the more i'm noticing where this stripping occurs. I think this may be a bigger problem we need to address. The few places I've followed the code are treating typed buckets and untyped buckets as the same entity as a result.

src/riak_repl_cs.erl
bowrocker, line 62: Not sure -- but if we leave it as it was (Bucket = riak_object:bucket(Object)) -- will it be ok (i.e. not crash)

jrwest, line 62: the functions need to change to account for the fact that `riak_object:bucket()` now returns `binary() | {binary(), binary()}` instead of stripping off the bucket type like is done now.

src/riak_repl_util.erl
bowrocker: Later in do_repl_put it puts the whole object (we do not strip, just don't use the type if it's there) -- I changed it to check for the {Type, Bucket} tuple so it wouldn't crash.

The put:

https://github.com/basho/riak_repl/blob/2dfb083d60dcfef40dcb3a6c561adbeb17f7d016/src/riak_repl_util.erl#L124

The logging:

https://github.com/basho/riak_repl/blob/2dfb083d60dcfef40dcb3a6c561adbeb17f7d016/src/riak_repl_util.erl#L152

I have verified that the object has a proper type on the other side by doing a  riakc_pb_socket:get/3 on the other side.

Does that make sense?

src/riak_repl2_rt.erl
bowrocker, line 234: That was my assumption.

include/riak_repl.hrl
```diff
@@ -52,8 +56,14 @@
 %% the to_wire() and from_wire() operations, see riak_repl_util.erl.
 %% Also see analagous binary_version() in riak_object, which is carried
 %% inside the wire format in "BinObj". w0 implies v0. w1 imples v1.
+%% w2 still uses v1, but supports encoding the type information
+%% for bucket types.
```
Vagabond, line 60: What happens on a cluster configured with v0 objects that is using bucket types?

src/riak_repl2_rt.erl
```diff
@@ -143,15 +147,24 @@ postcommit(RObj) ->
         %% on what the RT source and sink negotiated as the common version).
         Objects0 when is_list(Objects0) ->
             Objects = Objects0 ++ [RObj],
-            BinObjs = riak_repl_util:to_wire(w1, Objects),
+            Meta = set_bucket_meta(RObj),
+            
+            BinObjs = case orddict:fetch(?BT_META_TYPED_BUCKET, Meta) of
+                false ->
+                    lager:debug("encoding w1 format object"),
```
Vagabond: I don't think these debugs add a lot of value...

src/riak_repl2_rt.erl
```diff
@@ -217,3 +228,17 @@ do_ring_trans(F, A) ->
...
+    M = orddict:new(),
+    case riak_object:bucket(Obj) of
+        {Type, _B} ->
+            PropsHash = riak_repl_util:get_bucket_props_hash(riak_core_bucket_type:get(Type)),
+            lager:debug("typed bucket, setting meta data: Type:~p, HashProps:~p", [Type, PropsHash]),
+            M1 = orddict:store(?BT_META_TYPED_BUCKET, true, M),
+            M2 = orddict:store(?BT_META_TYPE, Type, M1),
+            orddict:store(?BT_META_PROPS_HASH, PropsHash, M2);
+        _B -> 
+            lager:debug("un-typed bucket, setting typed_bucket to false"),
```
Vagabond: More debugs in the critical path that don't seem very valuable.

src/riak_repl2_rtsink_conn.erl
```diff
@@ -285,8 +286,18 @@ do_write_objects(Seq, BinObjsMeta, State = #state{max_pending = MaxPending,
...
     Me = self(),
-    {DoneFun, BinObjs} = make_donefun(BinObjsMeta, Me, Ref, Seq),
-    riak_repl2_rtsink_helper:write_objects(Helper, BinObjs, DoneFun, Ver),
+    case make_donefun(BinObjsMeta, Me, Ref, Seq) of
+        {DoneFun, BinObjs, Meta} ->
+            case write_object(Meta) of
+            true ->
+                riak_repl2_rtsink_helper:write_objects(Helper, BinObjs, DoneFun, Ver);
+            false ->
+                lager:info("Bucket is of a type that is not equal on both the source and sink; not writing object.")
```
Vagabond: This has the potential to really log-spam a cluster. A stat counter for this might be a wiser plan.

include/riak_repl.hrl
buddhisthead, line 60: Is that even possible? A cluster that supports bucket types will be using v1 objects. But it is possible that we'll be sending v0 objects to a cluster because it doesn't support binary riak objects. Would we ever be sending bucket types to such a cluster?

src/riak_repl2_rtsink_conn.erl
```diff
@@ -285,8 +286,18 @@ do_write_objects(Seq, BinObjsMeta, State = #state{max_pending = MaxPending,
                                               acked_seq = AckedSeq,
                                               ver = Ver}) ->
     Me = self(),
-    {DoneFun, BinObjs} = make_donefun(BinObjsMeta, Me, Ref, Seq),
-    riak_repl2_rtsink_helper:write_objects(Helper, BinObjs, DoneFun, Ver),
+    case make_donefun(BinObjsMeta, Me, Ref, Seq) of
+        {DoneFun, BinObjs, Meta} ->
+            case write_object(Meta) of
```
Vagabond: Can we rename this maybe_write_object() instead to be clear what is is doing?

src/riak_repl2_rtsource_helper.erl
Vagabond, line 146: No, it ends up as the return value of maybe_send, and is retained.

src/riak_repl2_rtsource_helper.erl
```diff
@@ -124,13 +124,28 @@ maybe_send(Transport, Socket, QEntry, State) ->
             lager:debug("Did not forward to ~p; destination already in routed list", [Remote]),
             State;
         false ->
-            QEntry2 = merge_forwards_and_routed_meta(QEntry, Remote),
-            {Encoded, State2} = encode(QEntry2, State),
-            lager:debug("Forwarding to ~p with new data: ~p derived from ~p", [State#state.remote, QEntry2, QEntry]),
-            Transport:send(Socket, Encoded),
-            State2
+            case State#state.proto of 
+                Ver when Ver =:= {3,0} ->
```
Vagabond: This is not very forwards compatible. If we bump the protocol to 3.1, then this will stop passing.

Vagabond: Something like {Major, _Minor} when Major >= 3 -> is probably better, no?

src/riak_repl_util.erl
```diff
@@ -828,15 +845,51 @@ to_wire(w1, Objects) when is_list(Objects) ->
...
+from_wire(<<131, _Rest/binary>>=BinObjTerm) ->
+    binary_to_term(BinObjTerm);
+%% @doc Convert from wire version w2, which has bucket type information
+from_wire(<<?MAGIC:8/integer, ?W2_VER:8/integer,
+            TLen:32/integer, T:TLen/binary,
+            BLen:32/integer, B:BLen/binary,
+            KLen:32/integer, K:KLen/binary, BinObj/binary>>) ->
+    case T of
+        <<>> -> 
+            lager:debug("Found zero-length type, decoding default bucket-type"),
```
Vagabond: More debugs in the critical path.

src/riak_repl_util.erl
```diff
@@ -67,12 +67,15 @@
...
-         peer_wire_format/1
+         peer_wire_format/1,
+         get_bucket_props_hash/1
         ]).
 
 %% Defines for Wire format encode/decode
 -define(MAGIC, 42). %% as opposed to 131 for Erlang term_to_binary or 51 for riak_object
 -define(W1_VER, 1). %% first non-just-term-to-binary wire format
+-define(W2_VER, 2). %% first non-just-term-to-binary wire format
+-define(BUCKET_TYPES_PROPS, [consistent, datatype, n_val, allow_mult, last_write_wins]).
```
Vagabond, line 78: The CRDT and strong consistency people should sign off on this, if they have not already.

include/riak_repl.hrl
Vagabond, line 60: I think bucket types and object version 1 are orthagonal features. If they aren't than this is not a problem.

Vagabond:
I'm generally OK with this PR, after reading it. My main concern is with all the lager:debugs in the critical path, for every single write. lager:debug is cheap but not free and it also will completely spam the logs if anyone ever turns on global debugging.

I have not run the code, nor do I have time to until next week.

include/riak_repl.hrl
Vagabond, line 60: I just ran the bucket_types riak_test with object_format set to v0, and the test passed. This means that it is a valid permutation and must be handled by this PR. I will try to run the riak_test for this feature with v0 objects at some point, possibly not until Monday.

This should prevent merging until it is dealt with.

Vagabond:
I'm out of time on this for now, but I'm against this feature merging right now. I'm not satisfied that the v0 objects with bucket types edgecase is covered and the riak_test has some serious flaws and omissions.

I will try to continue working with the test next week.

include/riak_repl.hrl
bowrocker, line 60: I do not fully understand this scenario. What does it mean that a cluster is "configured with v0 objects"? 

riak_repl2_rt:postcommit/1 hard-codes the encoding to w1, which appears to encode to v1. I guess what you mean is that the to_wire with w1 can handle v0 objects, but w2 cannot? Is that correct?

src/riak_repl_util.erl
seancribbs, line 78: How is this used? Generally I think hard-coding these is a bad idea because it will tend to bitrot.

bookshelfdave, line 78: I generally like atoms as constants, I've seen too many atom-typo bugs.

seancribbs, line 78: Sure, but I'm more concerned about bucket(-type) properties that may arise in the future, and having that codified in two separate places. Probably bikeshedding.

jrwest, line 78: @seancribbs this is a list of properties, specific to repl, used to determine if two types on either side of the connection can be considered equal. Admittedly, the properties were arbitrarily suggested by me based on `riak_kv_bucket` & my knowledge of Riak and the 2.0 features relying on the creation invariant of bucket types.  I think a discussion about the properties contained in this list w/ other people familiar w/ them would be a smart thing to do.

For the curious, my reasoning for the existing properties was:
- `consistent` -  Just like w/ in a single cluster, we can't have strongly consistent data on one side mixed w/ eventually consistent data on the other side
- `datatype` - same for CRDTs
- `n_val` - at least when `consistent=true`, `n_val` cannot change w/in a single cluster so requiring them to be equal on the source & sink seems necessary as well. Since we don't have a conditional check in repl (if consistent, then n_val must also be equal) we have to consider it in all cases. I didn't see this as too much of an issue since repl hasn't really supported differing Ns that I was aware of -- except the reduced N work but `n_val` isn't affected by that iirc.
- `allow_mult` - it seemed like mixing semantic and syntactically merged data across clusters was a bad idea
- `last_write_wins` - ditto

Granted, now that I type out the above, I'm remember that the MDC story for strong consistency isn't what I had in mind when I suggested some of the properties above. So again, I think a discussion of this list would be prudent.

src/riak_repl2_rtsink_conn.erl
```diff
@@ -285,8 +286,18 @@ do_write_objects(Seq, BinObjsMeta, State = #state{max_pending = MaxPending,
                                               acked_seq = AckedSeq,
                                               ver = Ver}) ->
     Me = self(),
-    {DoneFun, BinObjs} = make_donefun(BinObjsMeta, Me, Ref, Seq),
-    riak_repl2_rtsink_helper:write_objects(Helper, BinObjs, DoneFun, Ver),
+    case make_donefun(BinObjsMeta, Me, Ref, Seq) of
+        {DoneFun, BinObjs, Meta} ->
+            case maybe_write_object(Meta) of
+            true ->
```
Vagabond: I think indenting here is a little off.

Vagabond:
So for posterity, the v0/v1 issue was a non-issue because the addition of bucket type metadata is independant of the wire encoding. v0 objects have their type extracted and placed in the metadata and are thus subject to all the normal rules.

With that, and the addressing of issues in the test, I hereby provide my +1 to merge.

bowrocker:
Changes were squashed and merged:

https://github.com/basho/riak_repl/commit/c51c3659a9c29dcf1b3e3c403f8a5f059c1a4260

PR #491, opened 2013-12-17
official deprecation notice from prod mgmt

bookshelfdave:
See also https://github.com/basho/riak_ee/pull/203


cmeiklejohn:
:+1:

PR #493, opened 2013-12-18
Prevent sink cluster fullsync stalls with tcp errors.

cmeiklejohn:
In the event that the tcp connection errors without completing
full-sync, and when the max fssinks is not configured and defaulted to
one, this causes the aae_sink process to terminate normally, but not
propagate errors to the fssink process.  This causes an abandoned fssink
process, which locks out any future reservations for fullsync causing
replication to stall indefinitely on the sink cluster.

In the event that max fssink setting is not configured to one, and
specified at another value, tcp errors can cause available reservation
slots to be depleted over time.

This problem was discovered with repl_aae_fullsync in riak_test and
should be forward ported to the 2.0 release in addition to the test.

This could also be part of the issue observed by OpenX, in addition to the
not_built handling.

cc: @bsparrow435 @jcapricebasho

PR #495, opened 2013-12-18
Ensure we retry partitions without trees.

cmeiklejohn:
In the event of a race where the tree has not beem computed yet,
schedule for retry in the same manner as if the partition was already
locked.

This will need to be ported forward to 2.0 and can be addressed with the
same test as the other AAE fixes.

cc: @bsparrow435 @jcapricebasho


cmeiklejohn:
It appears this is unnecessary, the default clause will cause it to be rescheduled for later at the end of the queue, which is sufficient.  I believe the root of the problems observed by OpenX is related to #493.

PR #496, opened 2013-12-18
Gracefully handle the not_built message from the hashtree.

cmeiklejohn:
Ensure that when aae_source dies, we handle the DOWN message and
terminate the fsssource process instead of ignoring the message.

Do not throw the error, or send the complete message to the sink, when
the process dies; both are not necessary.

### DIFF
--- a/src/riak_repl2_fssource.erl
+++ b/src/riak_repl2_fssource.erl
@@ -103,8 +103,6 @@ handle_call({connected, Socket, Transport, _Endpoint, Proto, Props},
                                                                    Transport, Socket,
                                                                    Partition,
                                                                    self()),
-            %% We want a 'DOWN' message when the aae worker stops itself for not_responsible
-            erlang:monitor(process, FullsyncWorker),
             %% Give control of socket to AAE worker. It will consume all TCP messages.
             ok = Transport:controlling_process(Socket, FullsyncWorker),
             riak_repl_aae_source:start_exchange(FullsyncWorker),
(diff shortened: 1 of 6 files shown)


buddhisthead:
Hey Chris, what test do you recommend to reproduce this? Will repl_aae_fullsync show it, if I reduce the max source to 1?

src/riak_repl2_fssource.erl
```diff
@@ -182,6 +182,10 @@ handle_info({'DOWN', Ref, process, _Pid, not_responsible}, State=#state{partitio
         {ok, State2} -> {noreply, State2};
         Error -> Error
     end;
+handle_info({'DOWN', Ref, process, _Pid, Reason}, State) ->
+    erlang:demonitor(Ref),
```
Vagabond: You don't need to demonitor a dead process.

src/riak_repl2_fssource.erl
```diff
@@ -182,6 +182,10 @@ handle_info({'DOWN', Ref, process, _Pid, not_responsible}, State=#state{partitio
...
+    lager:info("Received: ~p, stopping process.", [Reason]),
```
Vagabond: If the reason is Error or shutdown, we can probably suppress this.

Vagabond: normal or shutdown, I mean.

src/riak_repl_aae_source.erl
```diff
@@ -170,8 +170,6 @@ prepare_exchange(start_exchange, State=#state{transport=Transport,
         Error ->
             lager:info("AAE source failed get_lock for partition ~p, got ~p",
                        [Partition, Error]),
-            send_complete(State),
```
Vagabond, line 173: We should also remove the other throw()s like this, they will all trigger bad_return_type errors. We should also audit the other send_completes to see if they make sense vs just having the socket close.

cmeiklejohn:
Closing, incorporated into #500.

PR #498, opened 2013-12-19
Convert riak_core_cluster_conn to use the OTP gen_fsm behavior

kellymclaughlin:
Convert `riak_core_cluster_conn` to use the OTP `gen_fsm` behavior. The changes affect the internal operation of the module, but the API is unchanged. Also add an eqc property to exercise the fsm state transitions.
### Review checklist
- [ ] manual verification of code
- [ ] eunit (w/ gist of output)
- [ ] EQC (w/ gist of output)
- [ ] riak_test (w/ gist of output)
- [ ] Dialyzer
- [ ] XRef
- [ ] Coverage reports
- [ ] internal docs (design docs)
- [ ] man pages

### DIFF
--- a/src/riak_core_cluster_conn.erl
+++ b/src/riak_core_cluster_conn.erl
@@ -7,26 +7,68 @@
 
 -module(riak_core_cluster_conn).
 
+-behavior(gen_fsm).
+
 -include("riak_core_cluster.hrl").
 -include("riak_core_connection.hrl").
 
+
 -ifdef(TEST).
+
 -include_lib("eunit/include/eunit.hrl").
+
+%% Test API
+-export([current_state/1]).
+
 %% For testing, we need to have two different cluster manager services running
 %% on the same node, which is normally not done. The remote cluster service is
 %% the one we're testing, so use a different protocol for the client connection
 %% during eunit testing, which will emulate a cluster manager from the test.
 -define(REMOTE_CLUSTER_PROTO_ID, test_cluster_mgr).
--define(TRACE(Stmt),Stmt).
-%%-define(TRACE(Stmt),ok).
 -else.
 -define(REMOTE_CLUSTER_PROTO_ID, ?CLUSTER_PROTO_ID).
--define(TRACE(Stmt),ok).
 -endif.
 
--export([start_link/1]).
+%% API
+-export([start_link/1,
+         start_link/2,
+         status/1,
+         connected/6,
+         connect_failed/3,
+         stop/1]).
 
--export([connected/6, connect_failed/3, ctrlClientProcess/3]).
+%% gen_fsm callbacks
+-export([init/1,
+         initiating_connection/2,
+         initiating_connection/3,
+         connecting/2,
+         connecting/3,
+         waiting_for_cluster_name/2,
+         waiting_for_cluster_name/3,
+         waiting_for_cluster_members/2,
+         waiting_for_cluster_members/3,
+         connected/2,
+         connected/3,
+         handle_event/3,
+         handle_sync_event/4,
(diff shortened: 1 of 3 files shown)


src/riak_core_cluster_conn.erl
```diff
@@ -7,26 +7,68 @@
...
+
+%% Test API
+-export([current_state/1]).
+
 %% For testing, we need to have two different cluster manager services running
 %% on the same node, which is normally not done. The remote cluster service is
 %% the one we're testing, so use a different protocol for the client connection
 %% during eunit testing, which will emulate a cluster manager from the test.
 -define(REMOTE_CLUSTER_PROTO_ID, test_cluster_mgr).
--define(TRACE(Stmt),Stmt).
```
kellymclaughlin, line 20: I replaced `?debugFmt` calls wrapped in this `TRACE` macro with `lager:debug` calls. If there is a real desire to continue using the `TRACE` stuff, I'm happy to replace it.

src/riak_core_cluster_conn.erl
```diff
@@ -7,26 +7,68 @@
...
+         handle_event/3,
+         handle_sync_event/4,
+         handle_info/3,
+         terminate/3,
+         code_change/4]).
 
--export([connected/6, connect_failed/3, ctrlClientProcess/3]).
+-type remote() :: {cluster_by_name, clustername()} | {cluster_by_addr, ip_addr()}.
+-type peer_address() :: {string(), pos_integer()}.
+-record(state, {mode :: atom(),
```
kellymclaughlin, line 60: Some of these record members are unused and can probably be removed, but I just kept them all initally. _e.g._ `connection_ref`. Happy to hear opinions on that.

src/riak_core_cluster_conn.erl
```diff
@@ -35,196 +77,318 @@
...
+
+%% Sync message handling for the `initiating_connection' state
+initiating_connection(status, _From, _State) ->
+    {reply, initiating_connection, initiating_connection, _State};
+initiating_connection(_, _From, _State) ->
+    {reply, ok, initiating_connection, _State}.
+
+%% Async message handling for the `connecting' state
+connecting(timeout, State=#state{remote=Remote}) ->
+    %% @TODO Original comment below says we want to die, but code was
```
kellymclaughlin, line 149: I'd like to have an opinion on this comment as well.

src/riak_core_cluster_conn.erl
```diff
@@ -35,196 +77,318 @@
...
+    {next_state, connecting, State}.
+
+%% Sync message handling for the `connecting' state
+connecting(status, _From, State) ->
+    {reply, connecting, connecting, State};
+connecting(_, _From, _State) ->
+    {reply, ok, connecting, _State}.
+
+%% Async message handling for the `waiting_for_cluster_name' state
+waiting_for_cluster_name({cluster_name, NewName}, State=#state{previous_name="undefined"}) ->
```
kellymclaughlin, line 199: I thought it would be nice to switch to using an atom of `undefined` instead of string, but it requires plumbing outside this module and seemed more appropriate for a separate change.

kellymclaughlin:
The `replication2` riak_test requires an adjustment to pass with these changes. The `repl_util` r_t module uses raw message passing to solicit the connection status and these changes broke that. [This](https://github.com/basho/riak_test/pull/487) riak_test PR changes it to use the `riak_core_cluster_conn` API to get the status which should be less brittle to internal module changes.

cmeiklejohn:
Moving to 2.1.

lordnull:
Covered by #631

PR #500, opened 2013-12-20
Improve active anti-entropy replication.

cmeiklejohn:
Provided below is a summary of bugs fixed.

In the event that the tcp connection errors without completing
full-sync, and when the max fssinks is not configured and defaulted to
one, this causes the aae_sink process to terminate normally, but not
propagate errors to the fssink process.  This causes an abandoned fssink
process, which locks out any future reservations for fullsync causing
replication to stall indefinitely on the sink cluster.

In the event that max fssink setting is not configured to one, and
specified at another value, tcp errors can cause available reservation
slots to be depleted over time.

Gracefully handle the not_built message from the hashtree.

Ensure that when aae_source dies, we handle the DOWN message and
terminate the fsssource process instead of ignoring the message.

Do not throw the error, or send the complete message to the sink, when
the process dies; both are not necessary.

Ensure abnormal exits when the aae_source fails.

Remove monitors, handle not_responsible as a independent message instead
of a failure state.

Provide a bound on the number of times a partition will be retried in
the event of failure through the fullsync coodinator proces.  If we
exceed the number of retries for all queued partitions, end the
fullsync.

If ownership transfer is occuring, ensure we bail and nodify the
fscoordinator with an abnormal exit, which will reschedule the
partition.

No longer attempt to gain locks on hashtrees during the initialization
function of a sink process, which, in the event of failure would cause
connection manager to use an exponential backoff, in addition to
invalidating the endpoint preventing other partitions on the same node
from being synchronized.

Ensure that the AAE source process exits if the socket closes.

Ensure we reserve and unreserve partitions when we detect source
processes exiting.

### DIFF
--- a/src/riak_repl2_fs_node_reserver.erl
+++ b/src/riak_repl2_fs_node_reserver.erl
@@ -114,7 +114,7 @@ handle_cast({claim_reservation, Partition}, State) ->
     Reserved2 = cancel_reservation_timeout(Partition, State#state.reservations),
     {noreply, State#state{reservations = Reserved2}};
 
-handle_cast(Msg, State) ->
+handle_cast(_Msg, State) ->
     {noreply, State}.
 
 
--- a/src/riak_repl2_fs_node_reserver.erl
+++ b/src/riak_repl2_fs_node_reserver.erl
@@ -52,7 +52,7 @@ reserve(Partition) ->
     end.
 
 %% @doc Release a reservation for the given partition on the correct node.
--spec unreserve(Partition :: any()) -> 'ok'.
+-spec unreserve(Partition :: any()) -> 'ok' | 'busy' | 'down'.
 unreserve(Partition) ->
     Node = get_partition_node(Partition),
     try gen_server:call({?SERVER, Node}, {unreserve, Partition}) of
--- a/src/riak_repl2_fs_node_reserver.erl
+++ b/src/riak_repl2_fs_node_reserver.erl
@@ -101,6 +101,8 @@ handle_call({reserve, Partition}, _From, State) ->
     end;
 
 %% @hidden
+%% This message is a call to prevent unreserve/reserve races, as well as
+%% detect failed processes.
 handle_call({unreserve, Partition}, _From, State) ->
     Reserved2 = cancel_reservation_timeout(Partition, State#state.reservations),
     {reply, ok, State#state{reservations = Reserved2}};
--- a/src/riak_repl2_fs_node_reserver.erl
+++ b/src/riak_repl2_fs_node_reserver.erl
@@ -55,7 +55,15 @@ reserve(Partition) ->
 -spec unreserve(Partition :: any()) -> 'ok'.
 unreserve(Partition) ->
     Node = get_partition_node(Partition),
-    gen_server:cast({?SERVER, Node}, {unreserve, Partition}).
+    try gen_server:call({?SERVER, Node}, {unreserve, Partition}) of
+        Out ->
+            Out
+    catch
+        exit:{noproc, _} ->
+            down;
+        exit:{{nodedown, _}, _} ->
+            down
+    end.
 
 %% @doc Indicates a reservation has been converted to a running sink. Usually
 %% used by a sink.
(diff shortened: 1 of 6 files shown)


cmeiklejohn:
Looks like we're missing one commit here from @Vagabond.

buddhisthead:
Are there plans to address the potential deadlock in this PR?

Vagabond:
Just a reminder, we are not quite done here. I think we need to at least provide the option for infinite retries in the coordiantor, and possibly even default to it. I would also like to test a 3-way AAE fullsync (with all the trees built) and check that the locking behaves and that all 6 fullsyncs complete eventually.

buddhisthead:
Yes. Even though you've solved a deadlock between N-way when N=2, the N>2 case is not understood well yet.

cmeiklejohn:
@Vagabond's two concerns:
- Semantics changed away from infinite retry, with a default retry max of 5.
- Test bidirectional between 3 clusters.

src/riak_repl2_fs_node_reserver.erl
```diff
@@ -85,25 +93,25 @@ handle_call({reserve, Partition}, _From, State) ->
...
         true ->
-            lager:debug("Node busy for partition ~p. running=~p reserved=~p max=~p",
+            lager:info("Node busy for partition ~p. running=~p reserved=~p max=~p",
                         [Partition, Running, Reserved, Max]),
             {reply, busy, State}
     end;
 
+handle_call({unreserve, Partition}, _From, State) ->
+    Reserved2 = cancel_reservation_timeout(Partition, State#state.reservations),
+    {reply, ok, State#state{reservations = Reserved2}};
```
lordnull, line 108: Anytime I see a call that always returns 'ok' I wonder why it's not a cast. A comment here indicating detection of down process/nodes would help prevent a quick reading of the code from reaching that conclusion.

src/riak_repl2_fs_node_reserver.erl
```diff
@@ -85,25 +93,25 @@ handle_call({reserve, Partition}, _From, State) ->
...
-handle_cast({unreserve, Partition}, State) ->
-    Reserved2 = cancel_reservation_timeout(Partition, State#state.reservations),
-    {noreply, State#state{reservations = Reserved2}};
-
 handle_cast({claim_reservation, Partition}, State) ->
     Reserved2 = cancel_reservation_timeout(Partition, State#state.reservations),
     {noreply, State#state{reservations = Reserved2}};
 
-handle_cast(_Msg, State) ->
+handle_cast(Msg, State) ->
```
lordnull: Won't this change just cause a warning about an unused variable? I mean, I know we have a bunch of them already, but this wasn't a warning...

src/riak_repl2_fscoordinator.erl
```diff
@@ -330,79 +336,83 @@ handle_cast(_Msg, State) ->
...
+        {value, {Pid, {Index, _, _}=Partition}, Running} ->
 
             % likely a slot on the remote node opened up, so re-enable that
             % remote node for whereis requests.
             {_, _, Node} = Partition,
             NewBusies = sets:del_element(Node, State#state.busy_nodes),
 
+            % ensure we unreserve the partition on the remote node
+            % instead of waiting for a timeout.
+            Transport:send(Socket, term_to_binary({unreserve, Index})),
```
lordnull, line 356: Since this is a new message, shouldn't the protocol number be incremented so we don't send it to older clusters?

src/riak_repl_aae_source.erl
```diff
@@ -73,33 +73,17 @@ init([Cluster, Client, Transport, Socket, Partition, OwnerPid]) ->
     %% for all combined IndexNs and iterate over them. When finished, send
     %% COMPLETE msg to signal sink to die and unlock tree.
```
lordnull: These comments don't seem to have any bearing on what is happening here.

src/riak_repl2_fscoordinator.erl
cmeiklejohn, line 356: It's not a new message, we're just sending it additional places.  See: https://github.com/basho/riak_repl/blob/7124f92bde1e5eee43b9adfd8ae04c2f380366c4/src/riak_repl2_fscoordinator.erl#L706

src/riak_repl2_fs_node_reserver.erl
cmeiklejohn, line 108: Addressed in 7675406.

src/riak_repl_aae_source.erl
cmeiklejohn: Comments removed in ae900de.  The code got moved, and is straightforward and does not require further documentation.

src/riak_repl2_fs_node_reserver.erl
cmeiklejohn: Fixed in 3e5345d.

src/riak_repl2_fscoordinator.erl
lordnull, line 356: You are correct this is already handled by the fscoordinator_serv.

src/riak_repl2_fs_node_reserver.erl
lordnull, line 108: Not actually what I meant, but added docstrings are nice. Might just be me, but seeing handle_call/3 that always returns 'ok' seems like a code smell, indicating the message might better be handled in handle_cast/2 so as not to block the caller of the wrapping function (assuming there is one). By putting a comment here that the semantics of the message require detection of process/node failure, it means one does not need to find the wrapping function(s) to detainee that handle_call is indeed correct even when the answer is always 'ok'.

Bottom line: a nitpick about a comment on why this message is handled as a call rather than a cast even though the answer is always 'ok'.

cmeiklejohn, line 108: I'm of the opinion that unreserve should be a blocking call, as reserve is, to prevent races.  This is in addition to the failure detection case.

cmeiklejohn:
Other than your comments regarding comments, are you :+1: otherwise on this change set?

lordnull:
Yes, I"m :+1: on it. Code is clear as can be (considering) and the associated riak_test demonstrates the issues this pr fixes.

cmeiklejohn:
Merging.

PR #503, opened 2014-01-06
Don't write empty clusters to the ring, ever

Vagabond:
Before, if the cluster already existed in the ring and the new
memberlist was empty, it would be deleted from the ring. However if you
tried to disconnect an unconnected cluster it would try to delete it
from the ring by writing the member list as [], but since it was not in
the ring before, instead of deleting it, the cluster would be written
with a empty memberlist. The connection manager would then, as part of
ensuring a connection to all configured clusters, try to connect to it
and block any legitimate connection attempts to that cluster.

This patch treats a write of an empty member list when the cluster is
not in the ring as a no-op.


cmeiklejohn:
:+1: to merge

PR #505, opened 2014-01-10
Fix and refactor EQC, Eunit tests

bowrocker:
This branch makes many fixes and refactorings to the eunit test. As of this writing, it appears to get all tests passing on 2 MBPs as well as buildbot running R16B02 (- EQC tests). To enumerate all changes here would not be worth the effort, but reviewers please pay particular attention to:
1. The eqc tests have been scaled back considerably to get them passing, using testing_time or numtests. These should be reviewed, and then the testing times or iterations scaled back up to give the tests more teeth. We felt that it was important to get a stake in the ground and get the tests passing first, then have original authors or experts review the changes.
2. All ?TRACE macros have been removed from the core modules, and some in the core tests, replaced with lager:debug statements. There may be dupes/unnecessary calls.
3. riak_repl_util:generate_socket_tag/3 has been refactored to check for a possible error return from the Transport:sockname() call, as it was assuming the call would succeed. This call was failing on builbot with 'einval'. It is not clear whether the refactoring in this PR is sufficiently robust in the case of error, and should be reviewed carefully.


src/riak_core_connection_mgr.erl
```diff
@@ -379,7 +376,7 @@ handle_info({'EXIT', From, Reason}, State = #state{pending = Pending}) ->
             end
     end;
 handle_info(_Unhandled, State) ->
-    ?TRACE(?debugFmt("Unhandled gen_server info: ~p", [_Unhandled])),
+    lager:debug("Unhandled gen_server info: ~p", [_Unhandled]),
```
kellymclaughlin: This line is redundant with the next line so let's take it out.

cmeiklejohn:
Haven't revived the individual commits yet, but :+1: on the tests passing after a bunch of runs in a row.

src/riak_repl2_rtsource_conn.erl
```diff
@@ -314,21 +318,23 @@ handle_info(Msg, State) ->
...
-        _ ->
-            try
-                riak_repl2_rtsource_helper:stop(HelperPid)
-            catch
-                _:Err ->
-                    lager:info("Realtime source did not cleanly stop ~p - ~p\n",
-                               [HelperPid, Err])
-            end
-    end.
+    %% case HelperPid of
```
cmeiklejohn: Can we lose all of this commented code?

include/rt_source_eqc.hrl
```diff
@@ -0,0 +1,40 @@
...
+%% "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+%% KIND, either express or implied.  See the License for the
+%% specific language governing permissions and limitations
+%% under the License.
+%%
+%% -------------------------------------------------------------------
+
+%% all_remotes defines the list of remote clients the test will manage
+
+%% -define(all_remotes, ["a", "b", "c", "d", "e"]).
```
cmeiklejohn, line 24: Why is this here?

rebar.config
```diff
@@ -1,7 +1,7 @@
 %%-*- mode: erlang -*-
 
 {cover_enabled, true}.
-{erl_opts, [debug_info, warnings_as_errors, {parse_transform, lager_transform}]}.
```
cmeiklejohn, line 4: Why was this removed?  We want `riak_repl` built with `warning_as_errors`, as it's caught a bunch of problems.

include/rt_source_eqc.hrl
kellymclaughlin, line 24: This test needs more work. Originally this test used multiple remotes, but each one used the same `fake_sink` process. This lead to the test being flaky. For now we punted on further refactoring and changed it to just use a single source. This line was left as a reminder of how it previously looked and that it still needs more love.

rebar.config
kellymclaughlin, line 4: Pretty sure this is an accidental omission.

src/riak_repl2_rtsource_conn.erl
kellymclaughlin: Probably, I was a bit unsure why it as there in the first place. I don't think it's necessary, but I left it there commented to get some more informed opinions about its removal.

src/riak_repl2_rtsink_conn.erl
```diff
@@ -467,12 +472,9 @@ cleanup(_Ctx) ->
     riak_repl_test_util:kill_and_wait(riak_repl2_rtq),
     riak_repl_test_util:kill_and_wait(riak_repl2_rt),
     riak_repl_test_util:stop_test_ring(),
-    riak_repl_test_util:maybe_unload_mecks(
-      [riak_core_service_mgr,
-       riak_core_connection_mgr,
-       gen_tcp]),
-    application:set_env(riak_repl, reactivate_socket_interval_millis,
-          ?REACTIVATE_SOCK_INT_MILLIS),
+%    application:set_env(riak_repl, reactivate_socket_interval_millis,
```
cmeiklejohn: I assume this is no longer necessary?

src/riak_repl_util.erl
```diff
@@ -698,12 +698,18 @@ dropped_realtime_hook(Obj) ->
...
+        {ok, {_Address, Portnum}} ->
+            lists:flatten(io_lib:format("~s_~p -> ~p.~p.~p.~p:~p",[
                 Prefix,
                 Portnum,
                 O1, O2, O3, O4,
-                PeerPort])).
+                PeerPort]));
+        {error, Err} ->
+            lager:error("~p:sockname returned error:~p", [Transport, Err]),
+            []
```
cmeiklejohn: Nervous about this change, do we know that all consumers of this function will operate properly when the function returns empty list?

bowrocker: It's a good question, and I am not sure I know the right answer. The way it was, it would crash if sockname/1 failed.

test/riak_core_cluster_mgr_tests.erl
```diff
@@ -4,13 +4,12 @@
 -module(riak_core_cluster_mgr_tests).
 -compile(export_all).
 
+-ifdef(TEST).
 -include("riak_core_cluster.hrl").
 -include("riak_core_connection.hrl").
+%% -define(NODEBUG, true).
```
cmeiklejohn: Unnecessary.

src/riak_repl2_rtsink_conn.erl
bowrocker: Looks like I managed to bung that up -- replacing the maybe_unload_mecks call. I don't believe the set_env is necessary when doing a cleanup...

test/riak_core_cluster_mgr_tests.erl
```diff
@@ -321,12 +306,11 @@ start_link_setup(ClusterAddr) ->
     {ok, Pid1} = riak_core_service_mgr:start_link(ClusterAddr),
     {ok, Pid2} = riak_core_connection_mgr:start_link(),
     {ok, Pid3} = riak_core_cluster_conn_sup:start_link(),
-    %unlink(Pid3),
+    %unlink(Pid3ctrlClientProcess),
```
cmeiklejohn: Unnecessary.

test/riak_core_cluster_mgr_tests.erl
```diff
@@ -467,9 +445,14 @@ start_fake_remote_cluster_service() ->
     %% start our cluster_mgr service under a different protocol id,
     %% which the cluster manager will use during testing to connect to us.
     ServiceProto = {test_cluster_mgr, [{1,0}]},
+    %ServiceSpec = {ServiceProto, {?CTRL_OPTIONS, ?MODULE, ctrlService, []}},
```
cmeiklejohn: Unnecessary.

test/rt_source_helpers.erl
```diff
@@ -0,0 +1,312 @@
...
+    timer:sleep(WaitPeriod),
+    case is_process_alive(Pid) of
+        true ->
+            ok;
+        false ->
+            wait_for_pid(Pid, RetriesLeft - 1, WaitPeriod)
+    end.
+
+sink_acceptor(Listen, WhoToTell) ->
+    %% ?debugFmt("sink_acceptor: ~p", [self()]),
```
cmeiklejohn, line 177: Lots of commented out code here, it's hard to figure out what's happening.  Also, maybe docstrings?

cmeiklejohn:
:+1:, the tests are passing for me.

Some general comments:
- I've mentioned the commented code in a few places, but it's all over the place.  It would be great to go through this PR and identify and remove all of the commented code as it goes a long way to making the code a lot more readable.
- `test/rt_source_helpers.erl` would benefit from docstrings on the functions to at least detail in a few sentences what they are there for, and when they should be used.
- I've still got some reservations to the change with `generate_socket_tag` as I don't think the change is fully vetted across the code base, and I definitely do not think the test suite probably executes all code paths that may use that function.

kellymclaughlin:
+1 to those comments. i'd say let's do another cleanup pass through the changes, add some comments for `rt_source_helpers`, and maybe discuss the `generate_socket_tag` stuff a bit more.

bowrocker:
On second thought, I am +1 on Chris' comment except the cleaning up all the commented code, which I think is outside our scope here. I think this PR is far-ranging enough that it is probably a better idea to fix those in a separate effort.

Thoughts?

bookshelfdave:
+1 to a separate cleanup PR

cmeiklejohn:
I mean, I'll defer to you.  I just want to make sure it gets done.

However, I want to make that point that I believe since this PR introduced a ton of the comments code, the removal should be addressed in this PR.

bowrocker:
@cmeiklejohn I see your point. Some of the things we did to get things passing involved commenting out instead of deleting because we weren't quite sure what the original intent was, or what something was doing. In those cases, I think the original author, or someone with more knowledge of the particular test should look at it.  I will do my best to get some eyes on those bits.

I guess I look at this PR as a bit of a group effort, since it is so wide-ranging and involves so many parts of the code.

Also, commented test code that pre-existed this PR, of which there is also quite a bit, should still be a separate PR in my view.

cmeiklejohn:
This looks good; what's left here?

cmeiklejohn:
Left a comment on 0101fb0, which is a blocker for a :+1: from me.

cmeiklejohn:
All test pass for me, :+1: to merge.  Thanks @bowrocker.

PR #511, opened 2014-01-25
Ensure we update the bloom for object differences.

cmeiklejohn:
Fix a bug where we incremented the count of modified objects, but failed
to update the bloom filter accordingly preventing objects from being
replicated during fullsync.

Let me know if this looks good, and I'll backport the change to Riak 1.4.


Vagabond:
+1, fixes the bug.

PR #512, opened 2014-01-25
Prevent deadlock in hashtree compare.

cmeiklejohn:
Prevent races and deadlocks in computing the hashtree by removing
blocking receive calls and refactor into gen_fsm states.

This change prevents a deadlock in replication where a particular worker
process is blocked on waiting for changes to be computed when a status
call is made, which queries all workers from the fscoordinator running
on a leader node.

If this looks good, this will be back ported to Riak 1.4.

### DIFF
--- a/src/riak_repl_aae_source.erl
+++ b/src/riak_repl_aae_source.erl
@@ -15,6 +15,7 @@
          update_trees/2,
          cancel_fullsync/1,
          key_exchange/2,
+         compute_differences/2,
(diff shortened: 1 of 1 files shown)


Vagabond:
I'm +1 on this, although there's some other things I don't like about this module (like the twiddling with socket ownership and the synchronous messages whose return values are ignored).

However, this seems like an unmitigated improvement.

PR #515, opened 2014-01-27
Prevent deadlock in hashtree compare.

cmeiklejohn:
Prevent races and deadlocks in computing the hashtree by removing
blocking receive calls and refactor into gen_fsm states.

This change prevents a deadlock in replication where a particular worker
process is blocked on waiting for changes to be computed when a status
call is made, which queries all workers from the fscoordinator running
on a leader node.


cmeiklejohn:
Backport of #512.

Vagabond:
+1

PR #516, opened 2014-01-28
Removed reduced repl

lordnull:
Due to mutators being removed in kv, this will no longer work.


cmeiklejohn:
Looking at this now...

cmeiklejohn:
:+1:, looks good, getting passes on replication2 (in addition to what appear to be the normal timeouts, I get when running this locally).

PR #517, opened 2014-01-29
Add xref and remove ancient couchDB cruft

Vagabond:
The CouchDB stuff is a holdover from when we were being backwards
compatible with riak 1.0 and earlier, we have not used the 'syncv1'
strategy as the default since then, so we can safely remove it.


cmeiklejohn:
:+1:, replication2 passes, xref clean.

Vagabond:
regular 'replication; should pass as well, as while we've deprecated 'old' replication we have not yet removed it.

cmeiklejohn:
'replication' does not pass with or without this branch currently.  It times out.

Vagabond:
Do you see wacky things like this:

```
2014-01-31 12:25:16.229 [error] <0.8983.0> gen_fsm <0.8983.0> in state request_partition terminated with reason: {max_concurrency,{vnode_lock,548063113999088594326381812268606132370974703616}}
```

In the log?

cmeiklejohn:
So, it appears that this is probably good to merge, and we need to add a wait for the ring to converge in the riak test, correctly?

Vagabond:
Should we just merge this and fix the regular replication riak_test like we have to anyway?

cmeiklejohn:
Yes, please :+1: YOLO merge.

PR #510, opened 2014-01-30
Typed buckets are not replicated with fullsync

kellymclaughlin:
The comparison process does not currently handle _buckets_ of the form `{Type, Bucket}` and as a result any objects in typed buckets do not get replicated using fullsync. 
- [x] Fix fullsync replication problem
- [x] Augment existing bucket types riak_test to also exercise fullsync
### Related PRs
- Add `property_hash` function to `riak_core_bucket_type`: https://github.com/basho/riak_core/pull/516
- Changes to `riak_object` to include hash of the bucket type properties in the metadata dictionary by default: https://github.com/basho/riak_kv/pull/823
- Changes to the riak_test: https://github.com/basho/riak_test/pull/527

_Note_: Currently replicating from a 2.0 source cluster to a 1.4 or earlier sink cluster could break when the sink encounters typed buckets. @bowrocker is working on this as a separate issue.

### DIFF
--- a/src/riak_repl2_fssink.erl
+++ b/src/riak_repl2_fssink.erl
@@ -64,7 +64,8 @@ sync_register_service() ->
     %% 1,0 and up supports keylist strategy
     %% 1,1 and up supports binary object
     %% 2,0 and up supports AAE strategy
-    ProtoPrefs = {fullsync,[{1,1}, {2,0}]},
+    %% 3,0 and up supports Typed Buckets
+    ProtoPrefs = {fullsync,[{1,1}, {2,0}, {3,0}]},
     TcpOptions = [{keepalive, true}, % find out if connection is dead, this end doesn't send
                   {packet, 4},
                   {active, false},
--- a/src/riak_repl2_fssource.erl
+++ b/src/riak_repl2_fssource.erl
@@ -97,6 +97,7 @@ handle_call({connected, Socket, Transport, _Endpoint, Proto, Props},
     lager:info("Got caps: ~p", [TheirCaps]),
     Strategy = decide_common_strategy(OurCaps, TheirCaps),
     lager:info("Common strategy: ~p", [Strategy]),
+    {_, ClientVer, _} = Proto,
 
     case Strategy of
         keylist ->
@@ -106,7 +107,8 @@ handle_call({connected, Socket, Transport, _Endpoint, Proto, Props},
             %% We maintain ownership of the socket. We will consume TCP messages in handle_info/2
             Transport:setopts(Socket, [{active, once}]),
             {ok, FullsyncWorker} = riak_repl_keylist_server:start_link(Cluster,
-                                                                       Transport, Socket, WorkDir, Client),
+                                                                       Transport, Socket, 
+                                                                       WorkDir, Client, ClientVer),
             riak_repl_keylist_server:start_fullsync(FullsyncWorker, [Partition]),
             {reply, ok, State#state{transport=Transport, socket=Socket, cluster=Cluster,
                                     fullsync_worker=FullsyncWorker, work_dir=WorkDir,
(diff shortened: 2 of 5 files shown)


bowrocker:
### Testing

[x] manual verification of code
[x] eunit https://gist.github.com/bowrocker/8726261
[x] EQC https://gist.github.com/bowrocker/8726261
[x] riak_test https://gist.github.com/bowrocker/8726150
[ ] Coverage reports

:+1:

kellymclaughlin:
I have pushed some changes to this PR as well as the KV and Core PRs. The riak_test passes for me with these changes, but I would appreciate another verification of that.

kellymclaughlin:
As Jordan and I have exchanged comments on the `riak_core` [PR](https://github.com/basho/riak_core/pull/516) that stemmed from this work I feel more and more like going to the route of always adding the bucket type hash to each `riak_object` is the wrong move.

To be clear what we have at this point works fine and the test I wrote passes, but when trying to decide where different portions of this change should live none of the options feel like the right fit and I think it is a symptom of a larger issue nagging at my brain.

My main concern now is if by addressing this specific problem with a very specific solution we are missing the larger problem, if not the opportunity to solve it. With the advent of the cluster metadata system in riak there could be many similar uses that follow bucket types which require some knowledge of that metadata to be shared by source and sink clusters involved in fullsync replication. We absolutely do not want to have to implement specific solutions to each new instance of such metadata, especially if it involves more calculation being done for each `riak_object` or other potential performance impacts for users not even using replication. 

I think what is really needed is a general way to exchange metadata as part of the fullsync replication process. Right now we build lists of keys, exchange differences, and transfer data. We could easily add another step, _exchange metadata_, to the process. This step would just be a general way to transfer any metadata and the semantics of that data can be worked out in specific modules of the code on either source or sink side. 

One very clear advantage I can see to this in relation to fullsync and bucket types is that we can save the cost of data transfer when bucket type metadata is exchanged before data transfer and the source can preemptively detect a mismatch of bucket type properties. The current approach represented by this change does not allow for filtering until the data has already reached the sink side and the transfer cost has been incurred. 

Assuming I am not along in thinking this is a good idea, what is unknown to me is what action we can or should take to this end and when we can do it. Time is very short or nonexistent for the 2.0 release. Perhaps we have to live with a less than ideal band-aid for this release and spend time on it for the next release cycle. Then we just have to make it such that we are not painted into any corners with the band-aid in terms of backwards compatibility and commit to the feature for 2.1. 

This makes sense to me, but I would love to hear any other thoughts or opinions.

bowrocker:
@kellymclaughlin I agree with your assessment that exchanged cluster meta-data is the best way to go to avoid wasting resources, etc. There are probably many other uses for it besides typed-bucket replication. For the record, we went with the current solution because time was short. There was also a suggestion to exchanged bucket type information between clusters, but I think your proposal is more general.

However, having a mechanism to test bucket type equivalence doesn't seem like a bad thing in the abstract, and could also be generally useful. We could use it in the short term for typed bucket replication, but there is no reason we can't migrate off of it in future releases, leaving the mechanism in place if it does prove useful.

Doing it all now just doesn't seem possible in the time allowed -- just my two cents.

kellymclaughlin:
I had further discussion with @bowrocker this morning and we have concluded we are definitely stuck with the approach represented by this and the related PRs for the 2.0 release. We also agreed we need to do something better for the 2.1 release and I plan to communicate with Jon and PM to make some plans around this. 

I have pushed a few more commits to try and improve the organization and will try to move these PRs to being ready to merge. 

Ideally we will rip out all of the changes that are going into `riak_object` for the next release and I do not think there should be any implication to backwards compatibility for doing so. This makes it a little easier for me to not be as concerned about finding the absolute best home for the list of default properties that need to be hashed for replication to determine if a typed bucket can be replicated or not. I settled on duplicating the information in `riak_kv` and `riak_repl` for this reason and with the hope that no updates will be required before it is removed from `riak_kv`.

jrwest:
interesting that there seems to be a want for cluster metadata to be hooked into replication across clusters. we had originally (perhaps w/ not much thought for use cases) decided against this. open to discussing it for future releases (obviously not a 2.0 thing).

kellymclaughlin:
> interesting that there seems to be a want for cluster metadata to be hooked into replication across clusters. we had originally (perhaps w/ not much thought for use cases) decided against this. open to discussing it for future releases (obviously not a 2.0 thing). 

Cool, I'd love to discuss it more down the road and hear other thoughts on it. I also need time to better parse out my thoughts and observations on it. Thanks!

kellymclaughlin:
I got a +1 from Jon on the replication aspects of this. Any other thoughts or comments before we try to merge this?

PR #519, opened 2014-02-03
Filter consistent objects during fullsync replication

kellymclaughlin:
Replication of consistent objects is not supported in Riak 2.0. Add
code to check the bucket properties of an object during fullsync and
prevent replication if the consistent property is set. It is not
necessary to filter realtime write at this time because the postcommit
hook used for realtime is not invoked on the consistent code path.

### DIFF
--- a/src/riak_repl_util.erl
+++ b/src/riak_repl_util.erl
@@ -216,16 +216,40 @@ repl_helper_recv([{App, Mod}|T], Object) ->
 
 repl_helper_send(Object, C) ->
     B = riak_object:bucket(Object),
-    case proplists:get_value(repl, C:get_bucket(B)) of
-        Val when Val==true; Val==fullsync; Val==both; Val==undefined ->
+    case fullsync_enabled_for_bucket(B, C) of
+        true ->
             case application:get_env(riak_core, repl_helper) of
                 undefined -> [];
                 {ok, Mods} ->
                     repl_helper_send(Mods, Object, C, [])
             end;
-        _ ->
+        false ->
             lager:debug("Repl disabled for bucket ~p", [B]),
             cancel
+    end.
+
+fullsync_enabled_for_bucket(Bucket, _C) ->
+    case riak_core_bucket:get_bucket(Bucket) of
+        {error, _Reason} ->
+            ok;
+        Props ->
+            not is_consistent_bucket(Props) andalso is_fullsync_enabled(Props)
+    end.
+
+is_consistent_bucket(Props) ->
+    case lists:keyfind(consistent, 1, Props) of
+        {consistent, true} ->
+            true;
+        _ ->
+            false
+    end.
+
+is_fullsync_enabled(Props) ->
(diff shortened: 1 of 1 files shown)


kellymclaughlin:
The riak_test PR to exercise this feature is [here](https://github.com/basho/riak_test/pull/530)

kellymclaughlin:
This PR relies on [this](https://github.com/basho/riak_repl/pull/510) currently outstanding riak_repl PR and the branch for that PR is the base branch of this PR to ensure a proper diff.

src/riak_repl_util.erl
```diff
@@ -216,18 +216,42 @@ repl_helper_recv([{App, Mod}|T], Object) ->
...
             end;
-        _ ->
+        false ->
             lager:debug("Repl disabled for bucket ~p", [B]),
             cancel
     end.
 
+fullsync_enabled_for_bucket(Bucket, _C) ->
+    case riak_core_bucket:get_bucket(Bucket) of
+        {error, _Reason} ->
```
bsparrow435, line 233: Should we log this error?

src/riak_repl_util.erl
```diff
@@ -216,18 +216,42 @@ repl_helper_recv([{App, Mod}|T], Object) ->
 
 repl_helper_send(Object, C) ->
     B = riak_object:bucket(Object),
-    case proplists:get_value(repl, C:get_bucket(B)) of
-        Val when Val==true; Val==fullsync; Val==both; Val==undefined ->
+    case fullsync_enabled_for_bucket(B, C) of
```
bsparrow435, line 219: Do we need to add a wildcard to catch when an `ok` is returned instead of just true/false?

bsparrow435:
Passes attached test. :+1: after addressing comments.

src/riak_repl_util.erl
kellymclaughlin, line 219: No, `fullsync_enabled_for_bucket` should only return a boolean. That's a mistake, will fix.

src/riak_repl_util.erl
kellymclaughlin, line 233: I was thinking no since it will get logged at L227 (once I fix the `ok` issues you found). The only reason that can be returned (if the spec is to be believed) is `no_type` so I don't think the info is much more valuable. Though maybe a debug level statement would be appropriate to distinguish from other cases where `fullsync_enabled_for_bucket` may return `false`. Thoughts?

bsparrow435, line 233: I agree with you, it will be logged when false is returned for the case statement and we'll have the bucket name from that line if further investigation is needed.

src/riak_repl_util.erl
kellymclaughlin, line 219: Fixed with [697762d](https://github.com/basho/riak_repl/commit/697762dd2b2525cc4fb94a6e6fe250dd2e057151)

PR #520, opened 2014-02-04
Add protocol version checks in typed-bucket handling to fullsync for backward compatibility

bowrocker:
- Bumped up the fullsync protocol version to {3,0} for typed buckets
- Removed forcing of protocol version by negotiated strategy
- Added checks to negotiated protocol version so we only send typed buckets to {3,0} versions or above


kellymclaughlin:
- [x] Code review
- [x] Eunit and eqc tests pass
- [x] riak_test verification

kellymclaughlin:
:+1:

PR #538, opened 2014-02-12
Normalize representation.

cmeiklejohn:
Add missing realtime queue stats to the legacy stats call which was
accidentially omitted; in addition, ensure the same representation as
the JSON interface provides.

@bsparrow435 @metadave


bsparrow435:
:+1:  make tests pass:

  All 178 tests passed.

PR #539, opened 2014-02-12
Remove some more couch_merkle holdovers

Vagabond:
I forgot some crap from the couch_merkle removal.

cc @cmeiklejohn


cmeiklejohn:
:+1: yolo

PR #540, opened 2014-02-14
Include a bucket type hash in the message encoding for fullsync

kellymclaughlin:
Certain properties of typed buckets need to be identical on the source and sink as a condition for data replication between buckets using those bucket types.. For fullsync replication encode a hash of these bucket type properties in the messages sent from the source.  The sink side performs a comparison of a hash of its local bucket type properties for the corresponding bucket type and only writes the data if the hashes are identical.

This PR is an alternate approach to the path taken in [this](https://github.com/basho/riak_repl/pull/510) PR to address the fullsync + bucket types. The reason is yesterday I realized an omission to the `riak_kv` changes I did as part of [this](https://github.com/basho/riak_kv/pull/823) PR. I only account for the manner in which objects are written via the [PB client](https://github.com/basho/riak_kv/blob/develop/src/riak_kv_pb_object.erl#L234) which does a write with an empty value and then performs and update with the desired content. Any writes that only use `riak_object:new` result in an object that does not record the bucket type hash until that object is updated. This means any such object would be discarded from fullsync due to the bucket type hash not matching that on the sink side until such time as it was updated.

The fix in `riak_object` is straightforward, but adds yet more dependence on the availability of the cluster metadata service. I really dislike that fact and the more I think about it, the less I like having these changes solely designed to support fullsync replication in KV, especially given that this is a quick hack workaround for the 2.0 release that hopefully will be ripped out in favor of something better for 2.1 (see comments towards the end of #510). It is very convenient just having the hash there in the riak_object, but if it is just going to be ripped out maybe it is better to confine the change to riak_repl. 

These changes represent a fairly small difference in the code, but address the problem without needing any modifications to riak_object. My proposal is to use the changes in this PR and to revert the commit from [KV823](https://github.com/basho/riak_kv/pull/823). I have not run all of the replication riak_test modules yet, but all that I have run have passed and the eunit and eqc tests are passing for me (if run with the riak_object changes reverted; otherwise, `do_wire_list_w1_bucket_type_test` fails).

### DIFF
--- a/src/riak_repl_bucket_type_util.erl
+++ b/src/riak_repl_bucket_type_util.erl
@@ -30,8 +30,6 @@ bucket_props_match(Props) ->
 
 -spec bucket_props_match(riak_object:type(), integer()) -> boolean().
 bucket_props_match(Type, RemoteBucketTypeHash) ->
-    lager:info("Local hash: ~p Remoate Hash: ~p", [property_hash(Type),
-                                                   RemoteBucketTypeHash]),
    property_hash(Type) =:= RemoteBucketTypeHash.
 
 -spec is_bucket_typed({error, no_type} | proplists:proplist()) -> boolean().
--- a/src/riak_repl_bucket_type_util.erl
+++ b/src/riak_repl_bucket_type_util.erl
@@ -36,7 +36,6 @@ bucket_props_match(Type, RemoteBucketTypeHash) ->
 is_bucket_typed({error, no_type}) ->
     false;
 is_bucket_typed(Props) ->
-    lager:info("Props: ~p", [Props]),
     prop_get(?BT_META_TYPED_BUCKET, false, Props).
 
 -spec prop_get(binary(), term(), {error, no_type} | proplists:proplist()) -> term().
--- a/src/riak_repl_bucket_type_util.erl
+++ b/src/riak_repl_bucket_type_util.erl
@@ -28,17 +28,11 @@ bucket_props_match(Props) ->
             undefined =:= prop_get(?BT_META_PROPS_HASH, undefined, Props)
     end.
 
--spec bucket_props_match(riak_object:type(), [integer()]) -> boolean().
-bucket_props_match(Type, []) ->
-    property_hash(Type) =:= undefined;
-bucket_props_match(Type, [RemoteBucketTypeHash]) ->
-   property_hash(Type) =:= RemoteBucketTypeHash;
-bucket_props_match(_Type, _HashList) ->
-    %% Do not replicate if the hash list from the riak_object contains
-    %% more than 1 value because the local bucket type hash can never
-    %% match more than a single value. Must wait for the object to be
-    %% resolved and updated on the source side before replicating.
-    false.
+-spec bucket_props_match(riak_object:type(), integer()) -> boolean().
+bucket_props_match(Type, RemoteBucketTypeHash) ->
+    lager:info("Local hash: ~p Remoate Hash: ~p", [property_hash(Type),
+                                                   RemoteBucketTypeHash]),
+   property_hash(Type) =:= RemoteBucketTypeHash.
 
 -spec is_bucket_typed({error, no_type} | proplists:proplist()) -> boolean().
 is_bucket_typed({error, no_type}) ->
--- a/src/riak_repl_bucket_type_util.erl
+++ b/src/riak_repl_bucket_type_util.erl
@@ -38,7 +38,7 @@ is_bucket_typed({error, no_type}) ->
 is_bucket_typed(Props) ->
     prop_get(?BT_META_TYPED_BUCKET, false, Props).
 
--spec prop_get(binary(), term(), {error, no_type} | proplists:proplist()) -> term().
+-spec prop_get(atom() | binary(), term(), {error, no_type} | proplists:proplist()) -> term().
 prop_get(_Key, Default, {error, no_type}) ->
     Default;
 prop_get(Key, Default, Props) ->
--- a/src/riak_repl_util.erl
+++ b/src/riak_repl_util.erl
@@ -112,7 +112,7 @@ get_partitions(Ring) ->
     lists:sort([P || {P, _} <-
             riak_core_ring:all_owners(riak_core_ring:upgrade(Ring))]).
 
-do_repl_put(Object) ->
+do_repl_put({RemoteTypeHash, Object}) ->
     Bucket = riak_object:bucket(Object),
     Type = riak_object:type(Object),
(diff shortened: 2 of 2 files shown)


kellymclaughlin:
PR to revert the `riak_object` changes: https://github.com/basho/riak_kv/pull/844

andrewjstone:
- [x] eunit + eqc
- [x] riak_test replication tests
- [x] dialyzer
- [x] xref

src/riak_repl_bucket_type_util.erl
```diff
@@ -28,17 +28,9 @@ bucket_props_match(Props) ->
...
-    property_hash(Type) =:= undefined;
-bucket_props_match(Type, [RemoteBucketTypeHash]) ->
-   property_hash(Type) =:= RemoteBucketTypeHash;
-bucket_props_match(_Type, _HashList) ->
-    %% Do not replicate if the hash list from the riak_object contains
-    %% more than 1 value because the local bucket type hash can never
-    %% match more than a single value. Must wait for the object to be
-    %% resolved and updated on the source side before replicating.
-    false.
+-spec bucket_props_match(riak_object:type(), integer()) -> boolean().
```
andrewjstone, line 31: Vast improvement in this function :)

src/riak_repl_util.erl
```diff
@@ -828,13 +830,33 @@ deduce_wire_version_from_proto({_Proto, _Client, _Host}) ->
...
     case V of
         w0 ->
             term_to_binary({Cmd, RObj});
         _W ->
             BObj = riak_repl_util:to_wire(w1,RObj),
             term_to_binary({Cmd, BObj})
+    end;
+encode_obj_msg(V, {Cmd, RObj}, T) ->
+    BTHash = case riak_repl_bucket_type_util:property_hash(T) of
+                 undefined ->
```
andrewjstone, line 848: I wonder if it would be better to leave this as undefined. In theory, things can hash to 0. Not sure how this affects typing though.

andrewjstone:
Seeing the following related dialyzer warnings, that seem to be due to bucket types.

```
riak_repl_bucket_type_util.erl:20: Function bucket_props_match/1 has no local return
riak_repl_bucket_type_util.erl:22: The pattern 'true' can never match the type 'false'
riak_repl_bucket_type_util.erl:28: The call riak_repl_bucket_type_util:prop_get('properties_hash_val','undefined',Props::{'error','no_type'}) breaks the contract (binary(),term(),{'error','no_type'} | proplists:proplist()) -> term()
riak_repl_bucket_type_util.erl:40: The call riak_repl_bucket_type_util:prop_get('typed_bucket','false',Props::any()) breaks the contract (binary(),term(),{'error','no_type'} | proplists:proplist()) -> term()

riak_repl_util.erl:699: Expression produces a value of type 'undefined' | {integer(),integer(),integer()}, but this value is unmatched
riak_repl_util.erl:834: The call riak_object:type(RObj::tuple()) does not have an opaque term of type riak_object:riak_object() as 1st argument
riak_repl_util.erl:846: The pattern <V, {Cmd, RObj}, T> can never match since previous clauses completely covered the type <_,{_,_},'undefined'>
```

andrewjstone:
Seeing the following eunit/eqc error. I doubt it's related as I've seen passes. Likely test order related/race condition elsewhere.

```
42% {rt_source_eqc,connect_to_v1,2}
38% {rt_source_eqc,disconnect,1}
13% {rt_source_eqc,connect_to_v2,2}
8% {rt_source_eqc,push_object,3}
riak_repl2_rtsource_conn: riak_repl2_rtsource_conn_test_...*skipped*
undefined
*unexpected termination of test process*
::{badmatch,{error,eaddrinuse}}
```

@kellymclaughlin is looking into it, but it won't block this PR.

src/riak_repl_util.erl
kellymclaughlin, line 848: I went back and forth on that. I made it zero because the hash uses `erlang:phash2` and according to the [docs](http://www.erlang.org/doc/man/erlang.html#phash2-2):

> Range can be between 1 and 2^32, the function returns a hash value for Term within the range 1..Range.

So I think 0 is safe to use. I guess my motivation was just avoiding having two different encoding formats (1 for an `undefined` atom and 1 for an integer) for a case that should happen very rarely in practice.

andrewjstone, line 848: That works for me. I didn't realize phash started at 1.

andrewjstone:
repl_aae_fullsync passes, but others are failing.

Currently failing the following repl riak_tests:
- repl_aae_fullsync_custom_n
- repl_bucket_types 
- repl_reduced
- repl_rt_heartbeat

I gave up after those four ran. It's just an `ls -l` in order run.

cmeiklejohn:
I can run some of the r_t's on this today.

kellymclaughlin:
> repl_aae_fullsync passes, but others are failing.

There is a setup race prior to `connect_clusters` being called in `repl_bucket_types`. Fixing that now and I'll check out the rest of the failures.

lordnull:
repl_reduced can be skipped; it's testing a feature that has been bumped until 2.1.

kellymclaughlin:
[PR](https://github.com/basho/riak_test/pull/541) to fix `repl_bucket_types` test

kellymclaughlin:
The `repl_rt_heartbeat` failure happend after PR #535 was merged to correct the units of the heartbeat timeout and is unrelated to this change. [Here](https://github.com/basho/riak_test/pull/542) is a PR to fix that test.

kellymclaughlin:
As for the failure of `repl_aae_fullsync_custom_n` there is definitely some issues with this test, but as far as I can see they are all related to aae fullsync and not to my changes. I see the same behavior from the test using the HEAD of `develop` for `riak_repl`. I modified the test to not use aae fullsync and it passed for me. 

I can look more at cleaning up and fixing this test tomorrow, but at this point I don't see why it should block this change.

andrewjstone:
Failing riak_tests are being addressed by other PRs and seem unrelated to this.

Once the rest of the riak_repl_util dialyzer fixes are done I'll +1

kellymclaughlin:
> Failing riak_tests are being addressed by other PRs and seem unrelated to this.
> 
> Once the rest of the riak_repl_util dialyzer fixes are done I'll +1

Oops, I totally missed those the first time I read your comment. Fixed with [ca21266](https://github.com/basho/riak_repl/commit/ca2126691a6ac81e396ca09a647fd8f30cf2d57d)

andrewjstone:
Excellent thought leading on this issue @kellymclaughlin :+1:

PR #541, opened 2014-02-19
Minor dialyzer fixes.

cmeiklejohn:
Address some minor dialyzer warnings.  Remove some incorrect
specifications, add LC ignores.

@seancribbs


seancribbs:
Still lots of warnings, but none involving these lines. :+1:

cmeiklejohn:
Yeah, at this point smaller PRs seem to be the easiest way to make forward progress with repl, so that's going to be my strategy.

PR #542, opened 2014-02-21
Remove consistency check for realtime replication and fix fullsync helper function

kellymclaughlin:
- Refactor the is_fullsync_enabled helper function to more clearly
  enforce the intention that fullsync is enabled for buckets by
  default and must be explicitly disabled by specifying the repl
  property with a value of false or realtime.
- Remove the unnecessary check for a consistent bucket in
  riak_repl_util:repl_helper_send_realtime. It is broken due to a bad
  return value expectation from the call to
  riak_core_bucket:get_bucket and is not needed since the realtime
  postcommit hooks are bypassed by consistent writes.

### DIFF
--- a/src/riak_repl_util.erl
+++ b/src/riak_repl_util.erl
@@ -291,16 +291,9 @@ repl_helper_send([{App, Mod}|T], Object, C, Acc) ->
     end.
 
 repl_helper_send_realtime(Object, C) ->
-    IsConsistentBucket =
-        case riak_core_bucket:get_bucket(riak_object:bucket(Object)) of
-            {ok, Props} ->
-                is_consistent_bucket(Props);
-            {error, _Reason} ->
-                false
-        end,
     case application:get_env(riak_core, repl_helper) of
         undefined -> [];
-        {ok, Mods} when IsConsistentBucket =:= false ->
+        {ok, Mods} ->
             repl_helper_send_realtime(Mods, Object, C, [])
     end.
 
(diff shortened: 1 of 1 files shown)


src/riak_repl_util.erl
```diff
@@ -291,16 +293,9 @@ repl_helper_send([{App, Mod}|T], Object, C, Acc) ->
     end.
 
 repl_helper_send_realtime(Object, C) ->
-    IsConsistentBucket =
-        case riak_core_bucket:get_bucket(riak_object:bucket(Object)) of
-            {ok, Props} ->
```
kellymclaughlin, line 296: This check is not needed at all since consistent writes don't invoke the postcommit hooks, but this line was the problem. `riak_core_bucket:get_bucket` returns just the properties list and not an `ok` tuple; otherwise, this check would have been benign.

src/riak_repl_util.erl
```diff
@@ -260,11 +260,13 @@ is_consistent_bucket(Props) ->
     end.
 
 is_fullsync_enabled(Props) ->
+    %% Default to enabling fullsync for all buckets and only disable
+    %% if explicitly indicated.
     case lists:keyfind(repl, 1, Props) of
-        {repl, Val} when Val==true; Val==fullsync; Val==both; Val==undefined ->
```
kellymclaughlin, line 264: Most of the time the `repl` key will not even be present so needed to change this function so that fullsync would actually be enabled by default as intended.

andrewjstone:
make test, replication2 and repl_consistent_object_filter pass.

:+1:

PR #544, opened 2014-02-24
AAE fullsync may never terminate

kellymclaughlin:
Using AAE fullsync when a bucket has differing `n_val` between the source and sink clusters can cause the fullsync process to go into an infinite loop. This is due to the manner in which AAE fullsync handles this situation combined with a problem with handling strategy specification in `riak_repl2_fssource`. 

In the case of buckets with conflicting `n_val` values the `riak_repl_aae_source` worker is terminated for the partition and `riak_repl2_fssource` [downgrades](https://github.com/basho/riak_repl/blob/develop/src/riak_repl2_fssource.erl#L178) the strategy to `keylist` and establishes a new connection via the connection manager. 

What should happen next is that the fullsync for the partition is retried using the keylist strategy. In actuality what happens is that AAE fullsync is again selected as the strategy and the AAE tree comparison begins again and then again fails. This process repeats _ad infinitum_. This can be demonstrated by running the `repl_aae_fullsync_custom_n` riak_test. 

The reason this happens is that `riak_repl2_fssource` checks the value of the `fullsync_strategy` environment variable every time [`decide_our_caps`](https://github.com/basho/riak_repl/blob/develop/src/riak_repl2_fssource.erl#L250) is called. In the case of AAE fullsync, where `fullsync_strategy` must be set for it to be enabled this has the effect of overriding the attempt to downgrade to the `keylist` strategy and results in the fullsync never terminating. 

The fix to this problem is to change `riak_repl2_fssource` to only check the `fullsync_strategy` environment variable during initialization and to treat the `strategy` field of the state record as the canonical value for the current requested strategy instead of repeatedly going back to the application environment to obtain this value. This prevents the downgraded strategy value from being overwritten and allow the fullsync to complete even in situations of differing `n_val` values among clusters.

### DIFF
--- a/src/riak_repl2_fssource.erl
+++ b/src/riak_repl2_fssource.erl
@@ -55,13 +55,15 @@ legacy_status(Pid, Timeout) ->
 %% gen server
 
 init([Partition, IP]) ->
-    DefaultStrategy = ?DEFAULT_FULLSYNC_STRATEGY,
+    RequestedStrategy = app_helper:get_env(riak_repl,
+                                           fullsync_strategy,
+                                           ?DEFAULT_FULLSYNC_STRATEGY),
 
     %% Determine what kind of fullsync worker strategy we want to start with,
     %% which could change if we talk to the sink and it can't speak AAE. If
     %% AAE is not enabled in KV, then we can't use aae strategy.
-    OurCaps = decide_our_caps(DefaultStrategy),
-    SupportedStrategy = proplists:get_value(strategy, OurCaps, DefaultStrategy),
+    OurCaps = decide_our_caps(RequestedStrategy),
+    SupportedStrategy = proplists:get_value(strategy, OurCaps),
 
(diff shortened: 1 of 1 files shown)


bookshelfdave:
repl_aae_fullsync [passes](https://gist.githubusercontent.com/metadave/9196647/raw/8011762ebe4a65816c04112522997096b5a06678/foo.out)

+1, although you might want @lordnull @cmeiklejohn or @bowrocker's +1 as well.

bookshelfdave:
repl_aae_fullsync_custom_n  test results [here](https://gist.githubusercontent.com/metadave/9197292/raw/eeb31bfaefc304a9e89594fa40db737abb8c1c21/foo.out)

cmeiklejohn:
Did we ever run replication2 against this, considering the change to fssource?

kellymclaughlin:
> Did we ever run replication2 against this, considering the change to fssource?

Yep, it was run and passed prior to merging.

PR #549, opened 2014-03-01
Fullsync path logging cleanup

bsparrow435:
This branch addresses some of the logging issues we've had while testing aae fullsync. Specifically log spam and incorrectly classified log messages. All changes in this branch are to be logging and no functional changes.

### DIFF
--- a/src/riak_core_connection_mgr.erl
+++ b/src/riak_core_connection_mgr.erl
@@ -515,7 +515,7 @@ locate_endpoints({Type, Name}, Strategy, Locators) ->
                     Addrs
             end;
         error ->
-            lager:error("Unknown target: ~p ~p", [Name, Strategy]),
+            lager:warning("Unknown target: ~p ~p", [Name, Strategy]),
             {error, {unknown_target_type, Type}}
     end.
 
--- a/src/riak_core_connection_mgr.erl
+++ b/src/riak_core_connection_mgr.erl
@@ -320,9 +320,6 @@ handle_info({'EXIT', From, Reason}, State = #state{pending = Pending}) ->
     %% Work out which endpoint it was
     case lists:keytake(From, #req.pid, Pending) of
         false ->
-            %% Must have been something we were linked to, or linked to us
-            lager:debug("Connection Manager exiting because linked process ~p exited for reason: ~p",
-                        [From, Reason]),
             lager:error("Connection Manager exiting because linked process ~p exited for reason: ~p",
                         [From, Reason]),
             exit({linked, From, Reason});
@@ -361,8 +358,6 @@ handle_info({'EXIT', From, Reason}, State = #state{pending = Pending}) ->
                 Reason -> % something bad happened to the connection, reuse the request
                     lager:debug("handle_info: EP failed on ~p for ~p. removed Ref ~p",
                                      [Cur, Reason, Ref]),
-                    lager:warning("handle_info: endpoint ~p failed: ~p. removed Ref ~p",
-                                  [Cur, Reason, Ref]),
                     State2 = fail_endpoint(Cur, Reason, ProtocolId, State),
                     %% the connection helper will not retry. It's up the caller.
                     State3 = fail_request(Reason, Req, State2),
@@ -520,7 +515,7 @@ locate_endpoints({Type, Name}, Strategy, Locators) ->
                     Addrs
             end;
         error ->
-            lager:info("Unknown target: ~p ~p", [Name, Strategy]),
+            lager:error("Unknown target: ~p ~p", [Name, Strategy]),
             {error, {unknown_target_type, Type}}
     end.
 
--- a/src/riak_repl2_fscoordinator.erl
+++ b/src/riak_repl2_fscoordinator.erl
@@ -194,7 +194,7 @@ init(Cluster) ->
             riak_repl_util:schedule_cluster_fullsync(Cluster),
             {ok, #state{other_cluster = Cluster, connection_ref = Ref}};
         {error, Error} ->
-            lager:warning("Error connection to remote"),
+            lager:error("Error connecting to remote with message: ~p", [Error]),
             {stop, Error}
     end.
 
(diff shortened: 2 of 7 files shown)


src/riak_core_connection_mgr.erl
```diff
@@ -520,7 +515,7 @@ locate_endpoints({Type, Name}, Strategy, Locators) ->
                     Addrs
             end;
         error ->
-            lager:info("Unknown target: ~p ~p", [Name, Strategy]),
+            lager:error("Unknown target: ~p ~p", [Name, Strategy]),
```
cmeiklejohn: Are we sure about this change?  If a user sees this it might be confusing as it contains very little information.

src/riak_repl2_fscoordinator.erl
```diff
@@ -255,12 +255,12 @@ handle_call({node_clean, Node}, _From, State = #state{dirty_nodes=DirtyNodes}) -
     {reply, ok, NewState};
 
 handle_call(_Request, _From, State) ->
-    lager:info("ignoring ~p", [_Request]),
```
cmeiklejohn, line 258: Why is this not debug level?

cmeiklejohn:
Are we sure about removing the anya lines?  Not sure how valuable this is to debugging.  cc: @jonmeredith

cmeiklejohn:
I'm mostly :+1: on this, however, we should cut down whatever logging we don't absolutely need because we did determine during aae benchmarking that it is introducing overhead, even for lager:debug calls.

cmeiklejohn:
Please be sure that this makes it into the kill bill so it's accounted for and we don't forget to merge it.

jonmeredith:
Debug level is fine for me...

On Sat, Mar 1, 2014 at 6:00 PM, Christopher Meiklejohn <
notifications@github.com> wrote:

> Are we sure about removing the anya lines? Not sure how valuable this is
> to debugging. cc: @jonmeredith https://github.com/jonmeredith
> 
> ## 
> 
> Reply to this email directly or view it on GitHubhttps://github.com/basho/riak_repl/pull/549#issuecomment-36442375
> .

## 

Jon Meredith
VP, Engineering
Basho Technologies, Inc.
jmeredith@basho.com

src/riak_core_connection_mgr.erl
bsparrow435: This is not a code path I have ever hit either in development or in a customers logs so I am not 100% sure of the correct log level. Since i've never hit the path and it is a match on `error` i figured error was a more appropriate level. Maybe `warning` is better.

src/riak_repl2_fscoordinator.erl
bsparrow435, line 258: For all of the modules changed in this PR if I saw an `info` level log for a catch-all handle_call, handle_cast, handle_etc, I changed it to `notice`. My reasoning was to not alarm users but also make them visible in default log configurations. Is there a standard way we handle these catch-all's in other projects that's better? Maybe we should just delete this function and let erlang be erlang.

cmeiklejohn:
What's the deal with this?  Is this targeted for 2.0?

bsparrow435:
Sure! I owe one more run through to eliminate further, not needed, debug logs. I'll do it later today then I think we're good for merge after a rebase.

bsparrow435:
@cmeiklejohn this should be ready for review

cmeiklejohn:
Can you rebase/squash this into one commit?

andrewjstone:
- [ ] code review
- [ ] replication2

src/riak_repl2_fscoordinator.erl
```diff
@@ -331,14 +330,13 @@ handle_cast(stop_fullsync, State) ->
     {noreply, State2};
 
 handle_cast(_Msg, State) ->
-    lager:info("ignoring ~p", [_Msg]),
```
andrewjstone, line 334: I'm not sure removing this is a good idea, unless we expect useless messages. In my experience that usually indicates some bug and it'd be nice to keep it around. But I'm not sure how relevant that is in this code.

src/riak_repl2_fscoordinator.erl
```diff
@@ -391,7 +389,7 @@ handle_info({'EXIT', Pid, Cause},
 
             case dict:fetch(Partition, Retries) of
                 N when N > RetryLimit, is_integer(RetryLimit) ->
-                    lager:warning("fssource dropping partition: ~p, ~p failed"
+                    lager:warning("Fullsync dropping partition: ~p, ~p failed"
                                "retries", [Partition, RetryLimit]),
```
andrewjstone, line 393: Is this valid lagering? The "retries" as second parameter that is.

src/riak_repl2_fscoordinator.erl
```diff
@@ -462,7 +460,6 @@ handle_info(send_next_whereis_req, State) ->
     {noreply, State2};
 
 handle_info(_Info, State) ->
-    lager:info("ignoring ~p", [_Info]),
```
andrewjstone, line 465: I say let it crash or log it. just my 2c.

src/riak_repl2_fscoordinator.erl
```diff
@@ -695,11 +692,11 @@ start_fssource(Partition2={Partition,_,_} = PartitionVal, Ip, Port, State) ->
         {error, Reason} ->
             case Reason of
                 {already_started, OtherPid} ->
-                    lager:notice("A fullsync for partition ~p is already in"
+                    lager:info("A fullsync for partition ~p is already in"
                         " progress for ~p", [Partition,
```
andrewjstone, line 696: Is this a valid 3 param lager call?

src/riak_repl2_fscoordinator.erl
```diff
@@ -695,11 +692,11 @@ start_fssource(Partition2={Partition,_,_} = PartitionVal, Ip, Port, State) ->
...
                             riak_repl2_fssource:cluster_name(OtherPid)]);
                 {{max_concurrency, Lock},_ChildSpec} ->
-                    lager:notice("Fullsync for partition ~p postponed"
+                    lager:debug("Fullsync for partition ~p postponed"
```
andrewjstone, line 699: I'm starting to believe these are all valid. :)

src/riak_repl2_fscoordinator.erl
seancribbs, line 393: Actually that will appear as "failedretries", successive 'strings' are concatenated by the compiler.

cmeiklejohn, line 393: Please fix.

src/riak_repl2_fscoordinator.erl
cmeiklejohn, line 465: Logging is fine, but definitely not at info level.

cmeiklejohn:
Needs comments addressed and clean rebase/squash against master before merge.

cmeiklejohn:
Closing.

PR #551, opened 2014-03-02
Dialyzer fixes.

cmeiklejohn:
Address a series of dialyzer bugs.  Some, but not all.  @seancribbs

### DIFF
--- a/src/riak_core_cluster_mgr.erl
+++ b/src/riak_core_cluster_mgr.erl
@@ -134,8 +134,8 @@ get_is_leader() ->
     gen_server:call(?SERVER, get_is_leader, infinity).
 
 %% @doc Register a function that will get called to get out local riak node
-%% member's IP addrs. MemberFun(inet:addr()) -> [{IP,Port}] were IP is a string
--spec register_member_fun(MemberFun :: fun((inet:addr()) -> [{string(),pos_integer()}])) -> 'ok'.
+%% member's IP addrs. MemberFun(ip_addr()) -> [{IP,Port}] were IP is a string
+-spec register_member_fun(MemberFun :: fun((ip_addr()) -> [{string(),pos_integer()}])) -> 'ok'.
 register_member_fun(MemberFun) ->
     gen_server:cast(?SERVER, {register_member_fun, MemberFun}).
 
--- a/src/riak_core_cluster_mgr.erl
+++ b/src/riak_core_cluster_mgr.erl
@@ -61,7 +61,7 @@
 %% State of a resolved remote cluster
 -record(cluster, {name :: string(),     % obtained from the remote cluster by ask_name()
                   members :: [ip_addr()], % list of suspected ip addresses for cluster
-                  last_conn :: erlang:now() % last time we connected to the remote cluster
+                  last_conn :: erlang:timestamp() % last time we connected to the remote cluster
                  }).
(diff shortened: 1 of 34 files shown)


src/riak_core_cluster_mgr.erl
```diff
@@ -108,7 +108,6 @@
 %%% API
 %%%===================================================================
 %% @doc start the Cluster Manager
--spec(start_link() -> ok).
```
seancribbs, line 111: I realize these specs are wrong, but is there a good reason to remove them from API functions?

src/riak_core_service_mgr.erl
```diff
@@ -117,14 +117,12 @@ start_link({IP,Port}) when is_integer(Port), Port >= 0 ->
 %% for taking ownership of the socket via
 %% Transport:controlling_process(Socket, Pid).  Only the strategy of
 %% `round_robin' is supported; it's arg is ignored.
--spec(register_service(hostspec(), service_scheduler_strategy()) -> ok).
```
seancribbs, line 120: Ditto comment re: removal of specs from API functions

src/riak_repl2_fscoordinator.erl
```diff
@@ -314,8 +314,8 @@ handle_cast(start_fullsync,  State) ->
 
 handle_cast(stop_fullsync, State) ->
     % exit all running, cancel all timers, and reset the state.
-    [erlang:cancel_timer(Tref) || {_, {_, Tref}} <- State#state.whereis_waiting],
-    [begin
+    _ = [erlang:cancel_timer(Tref) || {_, {_, Tref}} <- State#state.whereis_waiting],
+    _ = [begin
```
seancribbs, line 318: I'd make a minor note to extract this into another function for clarity.

src/riak_repl2_ip.erl
```diff
@@ -11,7 +11,6 @@
 %% @doc Given the result of inet:getifaddrs() and an IP a client has
 %%      connected to, attempt to determine the appropriate subnet mask.  If
 %%      the IP the client connected to cannot be found, undefined is returned.
--spec determine_netmask(Ifaddrs :: [{atom(), any()}], SeekIP :: string() | {integer(), integer(), integer(), integer()}) -> 'undefined' | binary().
```
seancribbs, line 14: Again, API function. Might be better to simply use the proper types, e.g. `inet:ip_address()`

seancribbs:
Now that the shoe is on the other foot... I think it would be more helpful to put the warnings that were removed in the extended commit message. Then I can see that your change actually addressed the original warning. On the whole though, this is a vast improvement, I count 89 warnings removed.

src/riak_core_cluster_mgr.erl
cmeiklejohn, line 111: I put in the commit description that the goals was to re-add the signatures with typer once dialyzer was green.

seancribbs, line 111: Cool, thanks.

seancribbs:
:+1: so far. Are you going to continue eliminating warnings or do we want this to merge?

cmeiklejohn:
Going to merge and start a new PR moving forward, does that work?

seancribbs:
Works for me.

PR #560, opened 2014-03-10
Fixes wm stats for multiple client connections

lordnull:
1.3 back port of #556 .


cmeiklejohn:
yolo :+1: on the 1.3 cp

PR #561, opened 2014-03-12
Correctly jsonify mutiple fs source/sink

lordnull:
client_stats/server_starts become an object; each property is the name of
a site, with the value being an object of stats.

Further Fix for #555 .


bowrocker:
- [x] manual verification of code
- [x] eunit: https://gist.github.com/bowrocker/575058c40fab1362b227
- [x] riak_test: https://gist.github.com/bowrocker/9513305

:+1:

PR #562, opened 2014-03-12
Bug/mw/correctly jsonify mutiple fs source sink

lordnull:
2.0 version of #556 .

### DIFF
--- a/src/riak_repl_wm_stats.erl
+++ b/src/riak_repl_wm_stats.erl
@@ -157,16 +157,16 @@ jsonify_stats([{K, V}|T], Acc) when is_atom(K) and is_tuple(V)
     end;
 jsonify_stats([{K,V}|T], Acc) when is_atom(K)
         andalso (K == server_stats orelse K == client_stats), is_list(V) ->
-    StackedStats = lists:foldl(fun
-        ({Pid, Mq, {status, Stats}}, FoldAcc) ->
-            FormattedPid = format_pid(Pid),
-            NewStats = {status, lists:map(fun format_pid_stat/1, Stats)},
-            % could be client or server pid
-            ServerPid = {pid, FormattedPid},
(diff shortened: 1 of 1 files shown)


cmeiklejohn:
Once reviewed, please coordinate documentation update with @metadave and notify product management: cc: @michellep @gburd

andrewjstone:
- [x] make test
- [x] replication2
- [x] dialyzer

andrewjstone:
:+1: 

cc @lordnull @metadave @michellep @gburd

bookshelfdave:
I'd say email docs@basho with any requests so the team is aware of the change. I can help out with writing up docs if necessary.

cmeiklejohn:
@lordnull Can you write up something regarding the change?

lordnull:
Email sent, copied here for posterity:

This is the summary of the change to stats due to https://github.com/basho/riak_repl/pull/562 . This concerns >1.4.8 (including 2.0).

Perviously, client_stats and server_stats for full sync replication stacked all sinks/sources together in one large object:

```
[{pid, pid_one}, {site, <<“site one”>>}, {pid, pid_two}, {site, <<“site two”>>}]
```

Turning this into a json string would have both there, but decoding the string was decoder dependent; if a key exists more than once the decoder may take the first, last, or (in erlang) have multiple copies of the key. For javascript, there’s not multiple copies of the key at all, so the above would become:

```
    {‘pid’:’pid_two’, ‘site’:’site two’}
```

The change takes the site name (which is unique) and splits up the stats based on that, so the final javascript object would look like:

```
    {‘site one’: {‘pid’: ‘pid_one’, ‘site’: ‘site one’}, ‘site two’: {‘pid’: ‘pid_two’, ‘site’: ‘site two’}}
```

This allows one to iterate over the sites (foreach k in object) or lookup a site by a specific name.

PR #564, opened 2014-03-21
Another round of Dialyzer work.

cmeiklejohn:
Removes almost all but the remaining 15 Dialyzer errors.  Each fix is it's own commit and contains the Dialyzer errors removed.

Tagging @seancribbs and @andrewjstone for reviews.

### DIFF
--- a/src/riak_core_cluster_mgr.erl
+++ b/src/riak_core_cluster_mgr.erl
@@ -710,7 +710,7 @@ persist_members_to_ring(State, ClusterName, Members) ->
 %% that were saved in the ring
 cluster_mgr_sites_fun() ->
     %% get cluster names from cluster manager
-    Ring = riak_core_ring_manager:get_my_ring(),
+    {ok, Ring} = riak_core_ring_manager:get_my_ring(),
     Clusters = riak_repl_ring:get_clusters(Ring),
     [{?CLUSTER_NAME_LOCATOR_TYPE, Name} || {Name, _Addrs} <- Clusters].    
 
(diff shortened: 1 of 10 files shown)


src/riak_core_cluster_mgr.erl
```diff
@@ -710,7 +710,7 @@ persist_members_to_ring(State, ClusterName, Members) ->
 %% that were saved in the ring
 cluster_mgr_sites_fun() ->
     %% get cluster names from cluster manager
-    Ring = riak_core_ring_manager:get_my_ring(),
+    {ok, Ring} = riak_core_ring_manager:get_my_ring(),
```
bsparrow435, line 713: wow, good catch. Amazing this hasn't caused issues in the past

cmeiklejohn, line 713: We've flattened a bunch of the locator stuff, so it's probably used in very few cases at this point.

src/riak_repl_fullsync_worker.erl
```diff
@@ -78,7 +79,7 @@ handle_call({get, B, K, Transport, Socket, Pool, Partition, Ver}, From, State) -
 
     Preflist = [{Partition, OwnerNode}],
 
-    ReqID = make_ref(),
+    ReqID = make_req_id(),
```
bsparrow435, line 82: Wow, another great catch.

seancribbs:
Other than the one comment I left on 60987bc, I'm ok with this. Good work. :+1:

andrewjstone:
replication2 passes
 :+1:

PR #566, opened 2014-03-25
Finish him.

cmeiklejohn:
Resolve the remaining dialyzer errors on repl.  Flawless victory.

### DIFF
--- a/src/riak_repl_keylist_client.erl
+++ b/src/riak_repl_keylist_client.erl
@@ -207,20 +207,21 @@ request_partition({skip_partition, Partition}, State) ->
         [State#state.sitename, Partition, State#state.partition]),
     {next_state, request_partition, State}.
 
-send_keylist(Command, #state{kl_fh=FH, sitename=SiteName} = State)
-        when Command == cancel_fullsync ->
+send_keylist(cancel_fullsync,
+             #state{kl_fh=FH, sitename=SiteName} = State) ->
     % stop sending the keylist and delete the file
     _ = file:close(FH),
     _ = file:delete(State#state.kl_fn),
-    NewState = case Command of
-        cancel_fullsync ->
-            application:unset_env(riak_repl, {progress, SiteName}),
-            State#state{partitions=[], partition=undefined};
-        _ ->
-            State
-    end,
-    log_stop(Command, State),
+    application:unset_env(riak_repl, {progress, SiteName}),
+    NewState = State#state{partitions=[], partition=undefined},
+    log_stop(cancel_fullsync, State),
     {next_state, wait_for_fullsync, NewState};
+send_keylist(pause_fullsync,
+             #state{kl_fh=FH} = State) ->
+    _ = file:close(FH),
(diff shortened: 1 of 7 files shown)


seancribbs:
At first glance these seem ok to me, but I'm not a repl domain expert. This leaves only these remaining errors:

```
riak_core_connection_mgr.erl:494: The pattern 'ok' can never match the type {'error',atom()}
riak_repl_aae_source.erl:574: The pattern <MsgType, Msg, State> can never match since previous clauses completely covered the type <7,binary(),#state{index::'undefined' | non_neg_integer(),indexns::'undefined' | [{_,_}],tree_pid::'undefined' | pid(),tree_mref::'undefined' | reference(),built::'undefined' | non_neg_integer(),timeout::'undefined' | pos_integer(),wire_ver::atom(),diff_batch_size::non_neg_integer(),local_lock::boolean(),owner::'undefined' | pid()}>
riak_repl_console.erl:284: The created fun has no local return
riak_repl_console.erl:285: The pattern {'ok', Members} can never match the type [{string(),non_neg_integer()}]
```

cmeiklejohn:
Those will be addressed in a separate PR.

src/riak_repl_keylist_client.erl
```diff
@@ -257,17 +260,13 @@ send_keylist(continue, #state{kl_fh=FH0,transport=Transport,socket=Socket,kl_cou
...
-    NewState = case Command of
-        cancel_fullsync ->
-            application:unset_env(riak_repl, {progress, SiteName}),
-            State#state{partitions=[], partition=undefined};
-        _ ->
-            State
-    end,
-    log_stop(Command, State),
+wait_ack(cancel_fullsync, #state{sitename=SiteName} = State) ->
+    NewState = application:unset_env(riak_repl, {progress, SiteName}),
```
Vagabond: This is not the correct value for NewState.

Vagabond:
I don't understand the reason for this ignore:

```
riak_repl_keylist_client.erl:330: The pattern 'cancel_fullsync' can never match the type 'pause_fullsync'
```

Vagabond:
The rest of this looks fine.

cmeiklejohn:
As best as I could tell via Typer, the unset_env and set_env calls which are producing errors because Dialyzer feels they will never succeed, causes it to identify that the `cancel_fullsync` branch will never be reached.

Vagabond:
You could move the log message before the (un)set_env calls, that should shut things up without doing anything other than logging a few instructions earlier.

src/riak_repl_keylist_client.erl
cmeiklejohn: Nice catch.  This was resolved in 6bb6262.

cmeiklejohn:
Addressed logging issues with e30d78f437555e9543bdca8af9d79cca06759150.

cmeiklejohn:
Depends on https://github.com/basho/riak_kv/pull/882.

Vagabond:
+1 on this.

PR #575, opened 2014-04-17
Refactor rtsink helper.

cmeiklejohn:
Ensure we don't leak monitors remonitoring the same poolboy worker over
and over, and add logging on a process failing abnormally.


cmeiklejohn:
Closing.

PR #574, opened 2014-04-17
Refactor rtsink helper.

cmeiklejohn:
Ensure we don't leak monitors remonitoring the same poolboy worker over
and over, and add logging on a process failing abnormally.

Addresses #204.

Issue #204, opened 2013-02-20
Monitor leak


bookshelfdave:
See: https://github.com/basho/riak_repl/blob/master/src/riak_repl2_rtsink_helper.erl#L63

cmeiklejohn:
Moved to 2.0-RC.

cmeiklejohn:
How is this a leak?  If the process goes down, the monitor automatically is disabled, no?  cc: @vinoski

vinoski:
If `write_objects/4` is called repeatedly, then a new monitor is generated for each call, even for processes already being monitored. If the processes being monitored are long-lived, then those monitors sit around taking up space, essentially leaking until the monitored processes die.

cmeiklejohn:
Agreed, but it looks like the monitor is there by design, so until this entire mechanism is refactored we can't alleviate this "leak", right?

vinoski:
Perhaps the gen_server could keep in its state a table of `{Pid,MonitorRef}` pairs to avoid creating new monitors for processes already being monitored.

BTW this module seems to be missing a `-behavior(gen_server).` attribute.

slfritchie:
I agree with Steve, I think, except that I'm puzzled about what this code is supposed to be doing.  I'm a repl newbie, sorry.  And a poolboy newbie, also.

The gen_server process running riak_repl2_rtsink_helper's code appears to be long-lived and can handle multiple `write_object()` calls which get turned into casts to this one/many/doesn't-matter long-lived gen_server.  If the poolboy worker (which is allocated by riak_repl2_rtsink_helper's code but released by riak_repl_fullsync_worker's code, odd) is also long-lived, then two things happen:
1. We leak monitors by monitoring the same poolboy pid, perhaps up to a zillion times.
2. When there's a poolboy error, we're going to get somewhere between 1 and a tsunami of 'DOWN' messages, one from each monitor.  If we leak for a really long time, e.g. millions of monitors, then the VM will lock up a scheduler thread while it traverses the dead pid's monitor list and creates 'DOWN' message tuples for each one of them and delivers them all.

IIRC #2 won't schedule the recipient PID until all of the pending monitors have generated their 'DOWN' events.  The fact that the recipient will stop after receiving the first one doesn't matter: that pid needs to be runnable before it can receive any of them.

cmeiklejohn:
OK; I'm going to take a stab at rewriting this module.

cmeiklejohn:
Addressed by #574.

PR #576, opened 2014-04-18
Fixes monitor leak for realtime objects (#204).

jonmeredith:
The rtsink_helper uses a pool of riak_repl_fullsync_workers called
riak_repl2_rtsink_pool.  Fullsync punts on checking if the object
was written as it will catch it on the next sync.  Realtime uses
a monitor to make sure the put completes before it acks back to
the rtsource, unfortunately it neglects to demonitor.

Interestingly, the impact for this is mitigated by the poolboy
configuration in use.  The default pool size is 5 processes
in the rtsink_pool bursting to 100.  If the number of concurrent
puts stays below the initial 5, those 5 processes will be reused
and the monitors will grow indefinitely.  If more than 5 concurrent
puts are used, additional processes will be created and when
checked back in, all but the last 5 processes checked in will exit,
cleaning up the monitors.  Even with rtsink_pool_max set to min,
poolboy recycles processes so the monitors are eventually cleared.

When those processes happen to have had a lot of monitors built up
against them, they'll keep the scheduler busy when they die on walking
the proc->monitors structure, locking the remote process and sending
a message per monitor.

Alternative fix to #574, which doesn't try and reuse monitors.

### DIFF
--- a/src/riak_repl2_rtsink_helper.erl
+++ b/src/riak_repl2_rtsink_helper.erl
@@ -15,8 +15,14 @@
          write_objects/3]).
 
 %% gen_server callbacks
--export([init/1, handle_call/3, handle_cast/2, handle_info/2,
-         terminate/2, code_change/3]).
+-export([init/1,
+         handle_call/3,
+         handle_cast/2,
+         handle_info/2,
+         terminate/2,
+         code_change/3]).
+
+-behavior(gen_server).
 
 -record(state, {parent           %% Parent process
                }).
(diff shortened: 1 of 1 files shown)


jonmeredith:
@cmeiklejohn here's the PRs for the monitor issue #576, #577 and #578 for develop.  The way poolboy behaves means I suspect things are less bad than they could be, would like to check on some longer running customer clusters.

cmeiklejohn:
#578 has been :+1:'d, but I haven't had a chance to validate these fixes yet because I don't have these clusters built.  Given how similar they are, I'm :+1:, however.

jonmeredith:
@borshop merge 

Given the state of the 1.2bnw branches, going to merge this so we have a patching option.  Works by hand, suspect build failures are to do with not running R15 builder.

cmeiklejohn:
Manually merging.

PR #578, opened 2014-04-18
Fixes monitor leak for realtime objects (#204).

jonmeredith:
The rtsink_helper uses a pool of riak_repl_fullsync_workers called
riak_repl2_rtsink_pool.  Fullsync punts on checking if the object
was written as it will catch it on the next sync.  Realtime uses
a monitor to make sure the put completes before it acks back to
the rtsource, unfortunately it neglects to demonitor.

Interestingly, the impact for this is mitigated by the poolboy
configuration in use.  The default pool size is 5 processes
in the rtsink_pool bursting to 100.  If the number of concurrent
puts stays below the initial 5, those 5 processes will be reused
and the monitors will grow indefinitely.  If more than 5 concurrent
puts are used, additional processes will be created and when
checked back in, all but the last 5 processes checked in will exit,
cleaning up the monitors.

If those processes happen to have had a lot of montitors built up
against them, they'll keep the scheduler busy they die on walking
the proc->monitors structure, locking the remote process and sending
a message per monitor.

Alternative fix to #574, which doesn't try and reuse monitors.
Companion PR to #576.

### DIFF
--- a/src/riak_repl2_rtsink_helper.erl
+++ b/src/riak_repl2_rtsink_helper.erl
@@ -15,8 +15,14 @@
          write_objects/4]).
 
 %% gen_server callbacks
--export([init/1, handle_call/3, handle_cast/2, handle_info/2,
-         terminate/2, code_change/3]).
+-export([init/1,
+         handle_call/3,
+         handle_cast/2,
+         handle_info/2,
+         terminate/2,
+         code_change/3]).
+
+-behavior(gen_server).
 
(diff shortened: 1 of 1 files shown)


cmeiklejohn:
:+1:, replication2, rt_basic_test, rt_cascading, and replication2_rt_sink_connection, pass.

cmeiklejohn:
:+1: 4936d83

jonmeredith:
@borshop merge please

Issue #585, opened 2014-05-09
rpc:call try/catch rex


joedevivo:
```
/Users/joe/dev/basho/riak_ee/deps/riak_repl/src/riak_repl2_fscoordinator_serv.erl:
  243  
  244  get_node_ip_port(Node, NormIP) ->
  245:     {ok, {_IP, Port}} = rpc:call(Node, application, get_env, [riak_core, cluster_mgr]),
  246      {ok, IfAddrs} = inet:getifaddrs(),
  247      case riak_repl2_ip:determine_netmask(IfAddrs, NormIP) of
  ...
  264  
  265  get_matching_address(Node, NormIP, Masked) ->
  266:     case rpc:call(Node, riak_repl2_ip, get_matching_address, [NormIP, Masked]) of
  267          % this is a clause that will be removed/useless once all nodes on a 
  268          % cluster are >= 1.3
  269          {badrpc, {'EXIT', {undef, _StackTrace}}} ->
  270:             case rpc:call(Node, riak_repl_app, get_matching_address, [NormIP, Masked]) of
  271                  {badrpc, Err} ->
  272                      {error, Err};

/Users/joe/dev/basho/riak_ee/deps/riak_repl/src/riak_repl2_leader.erl:
  160                    try
  161                        %% The default timeouts are fine, failure is fine.
  162:                       LeaderPid = rpc:call(State#state.leader_node,
  163                                             erlang, whereis, [?SERVER]),
  164:                       {_, MboxSize} = rpc:call(State#state.leader_node,
  165                                                 erlang, process_info,
  166                                                 [LeaderPid, message_queue_len]),

/Users/joe/dev/basho/riak_ee/deps/riak_repl/src/riak_repl2_rtq_proxy.erl:
  127  get_peer_meta_support(Nodes) ->
  128      GetSupport = fun(Node) ->
  129:         case rpc:call(Node, riak_core_capability, get, [{riak_repl, rtq_meta}, false]) of
  130              Bool when is_boolean(Bool) ->
  131                  {Node, Bool};

/Users/joe/dev/basho/riak_ee/deps/riak_repl/src/riak_repl_app.erl:
  217      MaybeRetry = fun
  218          ({{badrpc, {'EXIT', {undef, _StrackTrace}}}, Node}) ->
  219:             RPCResult = rpc:call(Node, riak_repl_app, get_matching_address, Args),
  220              lager:debug("rpc to get_matching_address: ~p", [RPCResult]),
  221              RPCResult;

[...]

PR #587, opened 2014-05-21
Bugfix/mw/safer rpc

lordnull:
Creates a helper function that wraps rpc:call/4's in a try/catch. This is to catch the times when the rex process is not running on the remote node.

Addresses #585.

### DIFF
--- a/src/riak_repl_util.erl
+++ b/src/riak_repl_util.erl
@@ -1019,7 +1019,7 @@ maybe_downconvert_binary_objs(BinObjs, SinkVer) ->
 
 %% return the wire format supported by the peer node: w0 | w1
 peer_wire_format(Peer) ->
-    case rpc:call(Peer, riak_core_capability, get, [{riak_kv, object_format}]) of
+    case riak_core_util:safe_rpc(Peer, riak_core_capability, get, [{riak_kv, object_format}]) of
         {unknown_capability,{riak_kv,object_format}} ->
             w0;
         v1 ->
--- a/src/riak_repl2_fscoordinator_serv.erl
+++ b/src/riak_repl2_fscoordinator_serv.erl
@@ -242,7 +242,6 @@ get_partition_node(Partition, Ring) ->
     proplists:get_value(Partition, Owners).
 
 get_node_ip_port(Node, NormIP) ->
-    {ok, {_IP, Port}} = rpc:call(Node, application, get_env, [riak_core, cluster_mgr]),
     {ok, IfAddrs} = inet:getifaddrs(),
     case riak_repl2_ip:determine_netmask(IfAddrs, NormIP) of
         undefined ->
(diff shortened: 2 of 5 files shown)


src/riak_repl_util.erl
```diff
@@ -1053,6 +1054,22 @@ maybe_get_vnode_lock(SrcPartition) ->
...
+%% node, `{badrpc, rpc_process_down}' is returned.
+-spec safe_rpc(Node :: node(), Module :: atom(), Function :: atom(), Args :: [any()]) -> {'ok', any()}.
+safe_rpc(Node, Module, Function, Args) ->
+    try rpc:call(Node, Module, Function, Args) of
+        Out ->
+            lager:error("PANTS PANTS PANTS! ~p", [Out]),
+            Out
+    catch
+        'EXIT':{noproc, _NoProcDetails} ->
+            lager:error("FAIL FAIL FAIL"),
```
cmeiklejohn: It would be nice to have a better message here, given this will be displayed in customer log files.

src/riak_repl_util.erl
```diff
@@ -1053,6 +1054,22 @@ maybe_get_vnode_lock(SrcPartition) ->
...
 
+%% @doc Wraps rpc:call/4 in a try/catch. In cases when the rex process is
+%% not running on a remote node, an exit will occur. Often, this is more
+%% drastic than what is needed. If the rpc server is down on the remote
+%% node, `{badrpc, rpc_process_down}' is returned.
+-spec safe_rpc(Node :: node(), Module :: atom(), Function :: atom(), Args :: [any()]) -> {'ok', any()}.
+safe_rpc(Node, Module, Function, Args) ->
+    try rpc:call(Node, Module, Function, Args) of
+        Out ->
+            lager:error("PANTS PANTS PANTS! ~p", [Out]),
```
cmeiklejohn: It would be nice to have a better message here, given this will be displayed in customer log files.

cmeiklejohn:
Also, please check the buildbot logs.

cmeiklejohn:
I think that this `safe_rpc` function should be added to `riak_core` since we can use this in repl, core, kv, and assorted other locations where we are making rpc calls.  cc: @reiddraper

lordnull:
For the core rpc audit, there is a function that does the same. If the pr https://github.com/basho/riak_core/pull/586 gets merged in, this can be fixed up to point to that.

cmeiklejohn:
@lordnull @reiddraper Any update here?

lordnull:
I did an update to use the core's safe_rpc instead of keeping repls. So it's up to tests passing and updated review passing.

reiddraper:
Maybe worth squashing this down? There are currently 9 commits for 15 additions. Otherwise looking good to me.

cmeiklejohn:
Yes, can we take care of this today and get it merged in?

cmeiklejohn:
@reiddraper Good to merge?

reiddraper:
@borshop merge

PR #589, opened 2014-05-21
Update strip_postcommit to remove all possible repl hooks.

jonmeredith:
Previously only the currently enabled hooks were stripped.
If a custom bucket property contained a hook for a disabled
replication mode, then it would be resurrected whenever fixups ran.

Accidentally pushed fix direct to 1.4 branch.  If any issues with reviewing this we will have to fix there.

### DIFF
--- a/src/riak_repl.erl
+++ b/src/riak_repl.erl
@@ -7,6 +7,9 @@
 -export([install_hook/0, uninstall_hook/0]).
 -export([fixup/2]).
 -export([conditional_hook/3]).
+-ifdef(TEST).
+-include_lib("eunit/include/eunit.hrl").
+-endif.
 
 start() ->
     riak_core_util:start_app_deps(riak_repl),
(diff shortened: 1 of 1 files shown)


src/riak_repl.erl
```diff
@@ -66,5 +69,32 @@ strip_postcommit(BucketProps) ->
...
+
+-define(MY_HOOK1, {struct,
+                    [{<<"mod">>, <<"mymod1">>},
+                     {<<"fun">>, <<"myhook1">>}]}).
+-define(MY_HOOK2, {struct,
+                    [{<<"mod">>, <<"mymod2">>},
+                     {<<"fun">>, <<"myhook2">>}]}).
+strip_postcommit_test_() ->
+    [?_assertEqual(
+        [], % remember, returns the stipped postcommit from bprops
```
engelsanchez, line 86: little typo here because I HAD to comment on something

cmeiklejohn:
Should this be tagged for the 2.0-RC milestone?

engelsanchez:
Tagged now. I am verifying and this should be merged pretty soon.

cmeiklejohn:
OK.

engelsanchez:
I've tested this fix and its counterpart already in 1.4 branch (VP of oopsies).  In both cases, you can see how even with everything disabled, removing mode_repl12 from the modes list makes the v2 postcommit hook _reappear_ in the bucket properties, both default and in buckets with custom properties.   With this fix, removing the mode_repl12 mode always ends up removing the v2 hook.

When I say remove, I mean effectively removed by the fixups on local rings. The raw ring still has the hooks, of course. We do not really remove them.

:+1: :boat: :dancer:

engelsanchez:
:+1: 572517e

jonmeredith:
@borshop merge

PR #591, opened 2014-05-26
Ensure that cancel_fullsync works.

cmeiklejohn:
Ensure that the cancel_fullsync functionality actually cancels the
fullsync process.  Abort immediately, with a normal shutdown, to trigger
the helper function to stop as well.


andrewjstone:
:+1:  440a59

Issue #323, opened 2013-07-04
Can't cancel, pause fullsync (both default and advanced)


kuenishi:
see ZD #5028 , there're logs.

`riak_repl_keylist_server:bloom_fold/3` calls `gen_fsm:sync_send_event/3` to keylist server. When it is in `wait_for_partition` state, this server crashes because there's no `wait_for_partition/3`.

Its solution would be adding `wait_for_partition/3` or stop using `sync_send_event` or anything else.

cmeiklejohn:
Moving to 2.0-RC.

PR #592, opened 2014-05-27
No longer use sync_send_event.

cmeiklejohn:
Given that messages from the bloom_fold can arrive when the fullsync is
paused or cancelled, and we don't have handlers for sync_send_event in
any of the states except the diff_bloom state, as well as we don't
require the return value, move to use async message sending.

Addresses #323.  
Please merge before #591.  

cc: @jonmeredith

### DIFF
--- a/src/riak_repl_keylist_server.erl
+++ b/src/riak_repl_keylist_server.erl
@@ -62,8 +62,7 @@
          build_keylist/2,
          wait_keylist/2,
          diff_keylist/2,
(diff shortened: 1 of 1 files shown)


cmeiklejohn:
Can you get the review of this one, @andrewjstone?

andrewjstone:
Yep. I'll grab this one.

andrewjstone:
:+1: add831d

PR #595, opened 2014-05-30
Disable background manager by default.

cmeiklejohn:
@jrwest @kellymclaughlin

### DIFF
--- a/priv/riak_repl.schema
+++ b/priv/riak_repl.schema
@@ -132,7 +132,7 @@
 {mapping, "mdc.fullsync.use_bg_manager", "riak_repl.fullsync_use_background_manager", [
     {datatype, {enum, [true, false]}},
     {level, advanced},
-    {default, true}
+    {default, false}
 ]}.
 
 %% @doc How frequently the stats for fullsync source processes should be


kellymclaughlin:
Did a quick (hahaha) run of `replication2` and it is still passing. Eunit, eqc, dialyzer, and xref look good.

:+1:

kellymclaughlin:
@borshop merge

priv/riak_repl.schema
```diff
@@ -132,7 +132,7 @@
 {mapping, "mdc.fullsync.use_bg_manager", "riak_repl.fullsync_use_background_manager", [
     {datatype, {enum, [true, false]}},
     {level, advanced},
-    {default, true}
+    {default, false}
```
jrwest, line 135: fwiw, I noticed that the other bg manager cuttlefish mappings are using `{dataype, flag}`.

cmeiklejohn, line 135: @joedevivo Thoughts?

seancribbs, line 135: enum is still valid, no need to change it for RC

cmeiklejohn, line 135: Perfect, thanks.

jrwest, line 135: another thing i noticed -- also probably not needed before RC -- https://github.com/basho/riak_repl/blob/develop/test/riak_repl_schema_tests.erl does not test this mapping.

Issue #579, opened 2014-05-01
Memory leak on Sink during Realtime Replication


jcapricebasho:
We've discovered a reproducible memory leak on the SINK that occurs organically in conjunction with heartbeat timeouts on the SOURCE, and inorganically (reproducible) when the `rt_heartbeat_interval` is set to an incorrect value (such as `disabled` instead of `undefined` when trying to disable heartbeat.)  The leak manifests itself in binary memory, and the SOURCE must be taking load.  Otherwise, binary memory does not change.

To reproduce:
1. On the SOURCE, set `rt_heartbeat_interval` to `disabled` in the `riak_repl` section of the `app.config`
2. Start the SOURCE and SINK
3. Enable and start Realtime replication from SOURCE -> SINK

Once Realtime is enabled and started, you'll begin seeing these errors in the SOURCE (this is a result of the incorrectly configured `rt_heartbeat_interval` parameter.)

```
2014-05-01 16:16:45.503 [info] <0.24841.33>@riak_core_connection:try_ssl:208 "SOURCE" and "SINK" agreed to not use SSL
2014-05-01 16:16:45.503 [info] <0.24840.33>@riak_repl2_rtsource_conn:handle_call:227 Established realtime connection to site "SINK" address 127.0.0.1:10046
2014-05-01 16:16:45.504 [error] <0.24840.33> gen_server <0.24840.33> terminated with reason: bad argument in call to erlang:send_after(disabled, <0.24840.33>, send_heartbeat) in riak_repl2_rtsource_conn:schedule_heartbeat/1 line 404
2014-05-01 16:16:45.504 [error] <0.24840.33> CRASH REPORT Process <0.24840.33> with 0 neighbours exited with reason: bad argument in call to erlang:send_after(disabled, <0.24840.33>, send_heartbeat) in riak_repl2_rtsource_conn:schedule_heartbeat/1 line 404 in gen_server:terminate/6 line 747
2014-05-01 16:16:45.504 [error] <0.24823.33> Supervisor riak_repl2_rtsource_conn_sup had child "SINK" started with riak_repl2_rtsource_conn:start_link("SINK") at <0.24840.33> exit with reason bad argument in call to erlang:send_after(disabled, <0.24840.33>, send_heartbeat) in riak_repl2_rtsource_conn:schedule_heartbeat/1 line 404 in context child_terminated
```

When this issue is seen organically, heartbeat timeouts will be visible on the SOURCE.

On the SINK, the below messages are visible:

```
[...]

PR #596, opened 2014-05-31
Don't attempt to send heartbeat when invalid.

andrewjstone:
Don't attempt to send the heartbeat message when it's misconfigured as
a value which is not an integer, which causes the process to crash,
triggering a memory leak and unbounded growth in the binary heap.

A cmeik and andrew stone joint.

cc: @jcapricebasho @kellymclaughlin

### DIFF
--- a/src/riak_repl2_rtsource_conn.erl
+++ b/src/riak_repl2_rtsource_conn.erl
@@ -371,7 +371,7 @@ send_heartbeat(State = #state{hb_interval = undefined}) ->
 %% well as connection issues - either way we want to re-establish.
 send_heartbeat(State = #state{hb_timeout = HBTimeout,
                               hb_sent_q = SentQ,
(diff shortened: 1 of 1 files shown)


andrewjstone:
Resolves #579.

andrewjstone:
We may want to add a validator in cuttlefish to ensure that this value is an integer or undefined.  cc: @joedevivo

kellymclaughlin:
Nice job. Changes look good, just running through some testing. Is it worth also opening a PR for this change against the `1.4` branch just in case we have to do another 1.4.X release?

kellymclaughlin:
I verified that the binary memory holds steady when using different sized key ranges with the `repl_rt_heartbeat_interval` test. Also verified `replication2` passes with this change.

:+1:

cmeiklejohn:
@borshop merge

PR #598, opened 2014-06-02
Add logic to clear cancelled connection requests

bsparrow435:
This branch adds logic to clear cancelled connections from the
riak_core_connection_manager Pending state. I have also added a few
helper functions to give a better view into request
states(get_request_states) and connection
errors(get_connection_errrors).

Connections are cancelled only when a user runs `riak-repl disconnect
NAME/[IP,PORT]`. I’ve added logic to this path to  call
remove_cancelled_connections after waiting cm_cancellation_interval.
This interval must either be an integer or the atom never.

This logic only makes one attempt to remove the cancelled connection.
It might be nice to have a helper function that we can call to manually
clear the connections but currently this can be done by re-trying
`riak-repl disconnect`.

The riak_test showing this issue is `replication2_connections.erl` and
the issue can be demonstrated with this code by setting the
`cm_cancellation_interval` to `never`.

### DIFF
--- a/src/riak_core_connection_mgr.erl
+++ b/src/riak_core_connection_mgr.erl
@@ -418,9 +418,9 @@ disconnect_from_target(Target, State = #state{pending = Pending}) ->
             State;
         Req ->
             case get_cancellation_interval() of
-                never ->
+                undefined ->
                     lager:warning("Cancelled connection ~p will not be removed because
-                        cm_cancellation_interval is set to never", [Req#req.target]);
+                        cm_cancellation_interval is set to undefined", [Req#req.target]);
                 Interval when is_integer(Interval) ->
                     erlang:send_after(Interval,self(),{remove_cancelled_connection,Req#req.ref});
                 Error ->
@@ -432,15 +432,17 @@ disconnect_from_target(Target, State = #state{pending = Pending}) ->
                                                  Req#req{state = cancelled})}
     end.
 
-%% remove pending connection from state for supplied Ref.
+%% @doc remove pending connection from state for supplied Ref.
 remove_cancelled_connection(Ref, State = #state{pending = Pending}) ->
     case lists:keytake(Ref, #req.ref, Pending) of
         false ->
             State;
         {value, Req, Pending2} ->
-            if Req#req.state == cancelled ->
-                      State#state{pending=Pending2};
-               true -> State
+            case Req#req.state of
+                cancelled ->
+                    State#state{pending=Pending2};
+                _ ->
+                    State
             end
     end.
 
--- a/src/riak_core_connection_mgr.erl
+++ b/src/riak_core_connection_mgr.erl
@@ -234,6 +234,9 @@ handle_call({connect, Target, ClientSpec, Strategy}, _From, State) ->
                                                   State#state.pending,
                                                   Request)},
     lager:debug("Starting connect request to ~p, ref is ~p", [Target, Reference]),
+    %% reset backoff for all endpoints to expedite connection against
+    %% existing endpoint
+    ok = reset_backoff(),
     {reply, {ok, Reference}, start_request(Request, State2)};
 
 handle_call({get_endpoint_backoff, Addr}, _From, State) ->
--- a/src/riak_core_connection_mgr.erl
+++ b/src/riak_core_connection_mgr.erl
@@ -45,6 +45,7 @@
 
 -define(SERVER, riak_core_connection_manager).
 -define(MAX_LISTENERS, 100).
+-define(CM_CANCELLATION_INTERVAL, 5 * 60 * 1000). %% 5 minutes
 
 -type(counter() :: non_neg_integer()).
 
@@ -110,6 +111,8 @@
          register_locator/2,
          apply_locator/2,
          reset_backoff/0,
+         get_request_states/0,
+         get_connection_errors/1,
          stop/0
          ]).
 
@@ -193,6 +196,14 @@ connect(Target, ClientSpec) ->
 disconnect(Target) ->
     gen_server:cast(?SERVER, {disconnect, Target}).
 
+%% @doc Get the #req.target and #req.state for all connections
+get_request_states() ->
+    gen_server:call(?SERVER, get_request_states).
+
+%% @doc Return the #ep.addr and #ep.failures for all connections
+get_connection_errors(Addr) ->
+    gen_server:call(?SERVER, {get_connection_errors, Addr}).
+
 %% doc Stop the server and sever all connections.
 stop() ->
     gen_server:call(?SERVER, stop).
@@ -251,7 +262,6 @@ handle_call({should_try_endpoint, Ref, Addr}, _From, State = #state{pending=Pend
             {Answer, ReqState} =
                 case Req#req.state of
                     cancelled ->
-                        %% helper process hasn't cancelled itself yet.
                         {false, cancelled};
                     _ ->
                         {true, connecting}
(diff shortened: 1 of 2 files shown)


cmeiklejohn:
Dialyzer is returning the following problems:

```
riak_core_connection_mgr.erl:541: The pattern 'ok' can never match the type {'error',atom()}
```

bsparrow435:
Oh god, the `connection_helper`, i hate this stupid function. It's supposed to start and then when finished hit the 'EXIT' handle_info in the riak_core_connection_mgr to iterate stats and do some other logic. However, i've seen it totally skip this when `riak_core_connection:sync_connect` returns an `{error, Reason}` which is one reason the stats module is useless. This helper function needs to be totally reworked.

With that said, the call at that line is `riak_core_connection:sync_connect` who's spec is `-spec sync_connect(ip_addr(), clientspec()) -> ok | {error, atom()}.`

I don't see the issue here.

cmeiklejohn:
Test for this behavior: https://github.com/basho/riak_test/pull/626

cmeiklejohn:
@bsparrow435 How are we looking here?

src/riak_core_connection_mgr.erl
```diff
@@ -391,12 +417,33 @@ disconnect_from_target(Target, State = #state{pending = Pending}) ->
...
+                Error ->
+                    lager:error("Unsupported cm_cancellation_interval: ~p", [Error])
+            end,
             %% The helper process will discover the cancellation when it asks if it
             %% should connect to an endpoint.
             State#state{pending = lists:keystore(Req#req.ref, #req.ref, Pending,
                                                  Req#req{state = cancelled})}
     end.
 
+%% remove pending connection from state for supplied Ref.
```
cmeiklejohn: Missing `@doc`.

src/riak_core_connection_mgr.erl
```diff
@@ -391,12 +417,33 @@ disconnect_from_target(Target, State = #state{pending = Pending}) ->
...
+remove_cancelled_connection(Ref, State = #state{pending = Pending}) ->
+    case lists:keytake(Ref, #req.ref, Pending) of
+        false ->
+            State;
+        {value, Req, Pending2} ->
+            if Req#req.state == cancelled ->
```
cmeiklejohn: Rewrite as case, if statements are terrible to read.

src/riak_core_connection_mgr.erl
```diff
@@ -391,12 +417,33 @@ disconnect_from_target(Target, State = #state{pending = Pending}) ->
...
+            end,
             %% The helper process will discover the cancellation when it asks if it
             %% should connect to an endpoint.
             State#state{pending = lists:keystore(Req#req.ref, #req.ref, Pending,
                                                  Req#req{state = cancelled})}
     end.
 
+%% remove pending connection from state for supplied Ref.
+remove_cancelled_connection(Ref, State = #state{pending = Pending}) ->
+    case lists:keytake(Ref, #req.ref, Pending) of
```
cmeiklejohn, line 440: Is this a typo `#req.ref`?  What record are you matching against here?

src/riak_core_connection_mgr.erl
```diff
@@ -391,12 +417,33 @@ disconnect_from_target(Target, State = #state{pending = Pending}) ->
             %% already gone!
             State;
         Req ->
+            case get_cancellation_interval() of
+                never ->
```
cmeiklejohn: Wouldn't `undefined` be better here?

src/riak_core_connection_mgr.erl
bsparrow435, line 440: No, it's searching the `Pending` list which is a list of `#req`'s and saying extract the entry who's `#req.ref` matches `Ref`.

src/riak_core_connection_mgr.erl
bsparrow435: I wanted it to be explicit since unset app env's return `undefined`. With that said, the default argument in the app_helper:get_env should handle that case so i'm fine with whatever atom we want to use.

src/riak_core_connection_mgr.erl
bsparrow435: Looking at the rest of the module, most functions are missing `@doc` and type specifications. I'll add these to my new functions but would it be ok to go ahead and add the rest in this PR?

bsparrow435:
Just pushed commit addressing comments. Working on associated `riak_test` now.

src/riak_core_connection_mgr.erl
```diff
@@ -391,12 +417,35 @@ disconnect_from_target(Target, State = #state{pending = Pending}) ->
...
+            end,
             %% The helper process will discover the cancellation when it asks if it
             %% should connect to an endpoint.
             State#state{pending = lists:keystore(Req#req.ref, #req.ref, Pending,
                                                  Req#req{state = cancelled})}
     end.
 
+%% @doc remove pending connection from state for supplied Ref.
+remove_cancelled_connection(Ref, State = #state{pending = Pending}) ->
+    case lists:keytake(Ref, #req.ref, Pending) of
```
cmeiklejohn, line 440: I still don't see how this will ever match, given that the second parameter is supposed to be an integer.  Does this grab the default value from the record?

bsparrow435, line 440: I did it this way because a few lines above we did the same thing in lists:keystore. My understanding is that `#req.ref` will return an integer representing it's place in the record. I'll try testing that now to confirm. 

Looking at the rest of the module and other modules in `riak_repl` this is fairly common in all `lists:key*` operations.

bsparrow435, line 440: I just confirmed this behavior with the following immediately below line 431 of this file:

```
lager:notice("BRIAN RETURN: ~p",[#req.ref]),
```

Result:

```
2014-06-04 18:50:57.294 [notice] <0.2467.0>@riak_core_connection_mgr:disconnect_from_target:431 BRIAN RETURN: 2
```

cmeiklejohn, line 440: Ah, very interesting.  Thanks for taking the time to explain why it works.

cmeiklejohn:
I'm still concerned given this code introduced a new Dialyzer error which is not on master which should be fixed prior to merge.  Otherwise, once the Dialyzer warning is fixed, and the test refactored, we're good to go here.

src/riak_core_connection_mgr.erl
reiddraper, line 440: No big deal now, but that is definitely something I've never seen before. It's clever, but I think too much so.

bsparrow435:
I just found an issue where, if connected/disconnected in the correct order within 5 minutes, the endpoint(ex. "127.0.0.1":"10046") will be blacklisted forever. I need to add some logic to reset the retry_backoff. Will work on it tonight.

bsparrow435:
Also will be adding another riak_test to show issue

bsparrow435:
False alarm, i wasn't waiting long enough for the endpoint backoff to finish. Still, going to add one more step to the riak_test to prove reconnection is possible as well as adding logic to reset the backoff timer when the cancelled connection is removed.

src/riak_core_connection_mgr.erl
bsparrow435, line 440: The lists and orddicts of records within the State of this module are a nightmare. It should all be simplified and requests and endpoints broken out into separate state in different modules so we can use normal methods of interacting with them. This entire module is one big list/orddict searching/editing library for the gigantic state.

cmeiklejohn, line 440: I agree with @reiddraper.  It's a bit obtuse.  Can we just be explicit about the placement in the record?

bsparrow435, line 440: I went to the docs, from the bottom of the Records page: http://www.erlang.org/doc/reference_manual/records.html

```
In addition, #Record.Name returns the index in the tuple representation of Name of the record Record. Name must be an atom.
```

So this is an explicit, by design behavior.

bsparrow435, line 440: Seems the reason this works is because people were tired of defining indexes for elements in a record.

cmeiklejohn, line 440: Yes, I know it's by design.  I was arguing it was obtuse, given two Erlang programmers who have been doing it for 2+ years didn't know what it was doing, so we were making an argument to just put the integer value in (or macro).

reiddraper, line 440: Yep, and I know this is code you inherited, @bsparrow435, so not asking for it to be changed now. Might as well be consistent within the module until we fix the whole mess.

cmeiklejohn, line 440: Just leave it as is and we'll address all of it later.

bsparrow435:
Dialyzer issue is the last open item:

```
riak_core_connection_mgr.erl:546: The pattern 'ok' can never match the type {'error',atom()}
```

Myself and @cmeiklejohn are looking into it now.

cmeiklejohn:
@reiddraper found it; warning was is in the ignore file, but line number changed.  Just push a commit to update the line number.

bsparrow435:
Nice find, thanks.

cmeiklejohn:
Pushed https://github.com/basho/riak_repl/commit/d61ff7ec039ac1729a1bcb985380dcc87f522a0f with updated line numbers.

cmeiklejohn:
:+1: d61ff7e

cmeiklejohn:
@borshop merge

Issue #600, opened 2014-06-10
Mixed-version cluster fullsync can fail during bloom fold 


kellymclaughlin:
I discovered this problem investigating why the `replication_upgrade` test failed on the beta6 giddyup scorecard. It appears to be a result of the changes in #592. The 1.4 nodes are only able to handle async `diff_obj` messages and the 2.0 nodes are only able to handle sync `diff_obj` messages. This leads cases where the fullsync fails. It seems like there are two possibilities to resolve this: 
1. Add a means to detect the version of the node to which a `diff_obj` message is being sent.
2. Revert back to synchronous messaging for `diff_obj` messages and address the concerns of #592 in another way.

The crash log shows error reports similar to the following when this problem occurs:

```
2014-06-10 15:00:55 =ERROR REPORT====
** State machine <0.4639.1> terminating
** Last event in was {diff_obj,{r_object,<<117,92,246,72,6,36,170,156,171,64,114,117,71,238,84,169,45,114,101,97,108,116,105,109,101,95,111,110,108,121>>,<<0,0,0,18>>,[{r_content,{dict,3,16,16,8,80,48,{[],[],[],[],[],[],[],[],[],[],[],[],[],[],[],[]},{{[],[],[],[],[],[],[],[],[],[],[[<<"X-Riak-VTag">>,50,82,70,101,88,113,74,78,67,57,121,120,57,74,85,56,89,101,49,74,76,50]],[[<<"index">>]],[],[[<<"X-Riak-Last-Modified">>|{1402,432983,438605}]],[],[]}}},<<0,0,0,18>>}],[{<<92,176,215,10,83,151,109,34>>,{1,63569652183}}],{dict,1,16,16,8,80,48,{[],[],[],[],[],[],[],[],[],[],[],[],[],[],[],[]},{{[],[],[],[],[],[],[],[],[],[],[],[],[],[],[[clean|true]],[]}}},undefined}}
** When State == diff_bloom
[...]

PR #601, opened 2014-06-11
Handle synchronous events for rolling upgrades.

cmeiklejohn:
Add handling for synchronous events, related to #600.   However, I can't get replication_upgrade to get past the following:

```
19:35:42.272 [info] Wait until ring converged on ['dev4@127.0.0.1','dev5@127.0.0.1','dev6@127.0.0.1']
```

@kellymclaughlin do you mind running?

### DIFF
--- a/src/riak_repl_keylist_server.erl
+++ b/src/riak_repl_keylist_server.erl
@@ -59,9 +59,13 @@
 
 %% states
 -export([wait_for_partition/2,
+         wait_for_partition/3,
          build_keylist/2,
+         build_keylist/3,
          wait_keylist/2,
+         wait_keylist/3,
          diff_keylist/2,
+         diff_keylist/3,
          diff_bloom/2,
          diff_bloom/3]).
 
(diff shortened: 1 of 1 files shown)


kellymclaughlin:
I have determined the issue plaguing the `replication_upgrade` test was the [cancellation of fullsync](https://github.com/basho/riak_test/blob/master/tests/replication.erl#L531) when an attempt to start and wait for fullsync completion failed. This step is actually unnecessary because the only reason this condition occurs is that fullsync simply fails to start on the initial call to `riak_repl_console:start_fullsync/1` due to a race condition. So it is never the case that fullsync needs to be canceled. [This PR](https://github.com/basho/riak_test/pull/639) is open to remove that cancellation ad using the changes represented the `replication_upgrade` test passes for me. The PR contains an additional change to replace the wait function used to determine when a replication connection has been established after a node upgrade. It is not necessary for the test to pass, but should decrease the test execution time.

We should investigate why attempting to cancel fullsync when it is not running causes this problem, but in my view that seems like a separate and less severe issue.

src/riak_repl_keylist_server.erl
```diff
@@ -156,64 +161,26 @@ wait_for_partition({start_fullsync, _} = Command, State) ->
...
+    wait_for_individual_partition(Partition, State);
 %% Unknown event (ignored)
 wait_for_partition(Event, State) ->
     lager:debug("Full-sync with site ~p; ignoring event ~p",
         [State#state.sitename, Event]),
     {next_state, wait_for_partition, State}.
 
-build_keylist(Command, State)
-        when Command == cancel_fullsync ->
+build_keylist(Command, State) when Command == cancel_fullsync ->
```
kellymclaughlin, line 174: I know this is preexisting cruft, but the guards here could be eliminated in favor an explicit pattern match on the `Command` atoms and would make the code a bit more clear and concise.

kellymclaughlin, line 174: Looking closer I see that a similar pattern is used throughout this module so perhaps it is not worth worrying about at this time.

kellymclaughlin:
These changes look good and in concert with the riak_test PR mentioned above it resolves the `replication_upgrade` test failure.

:+1:

cmeiklejohn:
@borshop merge

Issue #611, opened 2014-08-19
Mismatched bucket types cause realtime replication to fill RTQ


jonmeredith:
If the bucket types are different between two clusters, realtime replication does
not acknowledge the incompatible object from the realtime sink to the realtime source,
causing the realtime queue to grow to maximum capacity and drop objects.

Root cause is the DoneFun isn't called immediately during the error case.

https://github.com/basho/riak_repl/pull/490/files#diff-4ca8b1cb0696e26e1d5024bc67e155d2R295

With the log level set at info, on a loaded cluster many log entries will be written
clearing it out.  The log message does not provide details on which bucket/bucket type
has issues.

Recommend

a) Restrict frequency of log message and perhaps keep a counter of occurrence (by bucket).
(Perhaps start keeping count with dict:add_counter, fire a timer after 10s or 60s to report totals and clear it/log changes)

b) Modify the r_t to verify that the RTQ is drained when testing bucket types.

bowrocker:
Fixes and riak test merged, closing.

PR #612, opened 2014-08-20
Address issue 611: acknowledge dropped puts due to bucket type mismatches

bowrocker:
The repl bucket types implementation is failing to acknowledge puts that are dropped on the sink, which leaves the dropped item in the RTQ, which will eventually fill. 
- Make sure to call DoneFun() in the mismatch case, so the drop will be acknowledged and cleared.
- Make logging of the mismatch a debug.
- Make the logging report the offending bucket type.
- Add a counter of mismatch events and fire a timer to report it. 
- Add a configuration parameter to control the interval of reporting: bucket_type_drop_report_interval, default 10s.

Issue: https://github.com/basho/riak_repl/issues/611
Riak test: https://github.com/basho/riak_test/pull/676
### Review checklist
- [ ] manual verification of code
- [ ] eunit (w/ gist of output)
- [ ] EQC (w/ gist of output)
- [ ] riak_test (w/ gist of output)
- [ ] Dialyzer
- [ ] XRef
- [ ] Coverage reports
- [ ] internal docs (design docs)
- [ ] man pages

### DIFF
--- a/src/riak_repl2_rtsink_conn.erl
+++ b/src/riak_repl2_rtsink_conn.erl
@@ -34,7 +34,7 @@
 
 % how long to wait to reschedule socket reactivate.
 -define(REACTIVATE_SOCK_INT_MILLIS, 10).
--define(DEFAULT_INTERVAL_MILLIS, 2000).
+-define(DEFAULT_INTERVAL_MILLIS, 10000).
 -define(DEFAULT_BUCKET_TYPE, <<"default">>).
 
 -record(state, {remote,           %% Remote site name
--- a/src/riak_repl2_rtsink_conn.erl
+++ b/src/riak_repl2_rtsink_conn.erl
@@ -34,7 +34,7 @@
 
 % how long to wait to reschedule socket reactivate.
 -define(REACTIVATE_SOCK_INT_MILLIS, 10).
--define(DEFAULT_INTERVAL_MILLIS, 10000).
+-define(DEFAULT_INTERVAL_MILLIS, 2000).
 -define(DEFAULT_BUCKET_TYPE, <<"default">>).
 
 -record(state, {remote,           %% Remote site name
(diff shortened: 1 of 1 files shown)


src/riak_repl2_rtsink_conn.erl
```diff
@@ -217,7 +229,12 @@ handle_info(reactivate_socket, State = #state{remote = Remote, transport = T, so
                                 [Remote, Reason]),
                     {stop, normal, State}
             end
-    end.
+    end;
+handle_info(report_bt_drops, State=#state{bt_drops = DropDict}) ->
+    _ = [ lager:debug("drops due to missing or mismatched type ~p:~p",
```
jonmeredith: Now that we're rate limited, this can be an error message. Would be good to give a count so operators can know how many updates were lost to help plan any mitigations after discovered.  `dict:fold` would give you both easily.

src/riak_repl2_rtsink_conn.erl
```diff
@@ -217,7 +229,12 @@ handle_info(reactivate_socket, State = #state{remote = Remote, transport = T, so
                                 [Remote, Reason]),
                     {stop, normal, State}
             end
-    end.
+    end;
+handle_info(report_bt_drops, State=#state{bt_drops = DropDict}) ->
+    _ = [ lager:debug("drops due to missing or mismatched type ~p:~p", 
+                  [BucketType, dict:fetch(BucketType, DropDict)]) ||
+        BucketType <- dict:fetch_keys(DropDict) ],
+    {noreply, State#state{bt_drops = dict:new(), bt_timer = undefined}}.
```
jonmeredith: May be nice to recheck interval from the appenv here so you can adjust without having to restart the connection.

jonmeredith:
Very quick scan of the code looks good.  Looks like calling the DoneFun on drops is valid, and that cascaded realtime would still work if you had A <-> B <-> C configured with A and C having matching types and B misconfigured, data would still arrive at B (although still would not fullsync).

Over time we may need to support other reasons for drops than bucket type misconfigurations (perhaps security restrictions), but we can refactor the reporting/state names then as it is all process-local.

I won't have time to complete review/running tests, so please could somebody else take the official review.

kellymclaughlin:
Shouldn't this be opened against the `2.0` branch?

bowrocker:
Closing in favor of https://github.com/basho/riak_repl/pull/613.

PR #616, opened 2014-09-18
Improved, pipelined AAE fullsync

jtuple:
This pull-request includes multiple improves to AAE fullsync, including:
1. Making AAE exchange pipelined. This is the most significant change, since non-pipelined AAE is sensitive to network latency. And in the worst case was shown to take weeks to complete, even for a trivial exchange. The new pipelined approach is designed to handle high latency links.
2. Ensuring AAE fullsync only does at most 1 bloom fold per partition. Previously, the implementation did a bloom fold per AAE tree per partition.
3. Directly sending up to a configurable threshold of key differences, thus avoiding the bloom fold entirely when clusters are mostly in sync.
4. Using the internal `n_val=1` get option when retrieving objects to send. This sends only a single request to a single vnode, therefore reducing cluster load during a fullsync.

Preliminary test result can be found [here](https://gist.github.com/jtuple/92e0fe300579ed24ef80). More testing is underway/planned.

Sibling pull-requests: basho/riak_core#628, basho/riak_kv#1023

### DIFF
--- a/src/riak_repl_aae_sink.erl
+++ b/src/riak_repl_aae_sink.erl
@@ -15,6 +15,7 @@
 
 %% API
 -export([start_link/4, init_sync/1]).
+-export([sender_init/2, sender_loop/1]).
 
 -record(state, {
           clustername,
@@ -22,6 +23,7 @@
           transport,
           tree_pid      :: pid(),  %% pid of the AAE tree
           partition,
+          sender        :: pid(),
           owner         :: pid()   %% our fssource owner
          }).
 
@@ -51,7 +53,8 @@ handle_call(init_sync, _From, State=#state{transport=Transport, socket=Socket}) 
                   {nodelay, true},
                   {header, 1}],
     ok = Transport:setopts(Socket, TcpOptions),
-    {reply, ok, State};
+    Sender = spawn_sender(Transport, Socket),
+    {reply, ok, State#state{sender=Sender}};
 
 handle_call(status, _From, State) ->
     Reply = [{partition_syncing, State#state.partition}],
@@ -73,7 +76,7 @@ handle_cast(_Msg, State) ->
 handle_info({Proto, _Socket, Data}, State=#state{transport=Transport,
                                                  socket=Socket}) when Proto==tcp; Proto==ssl ->
     TcpOptions = [{active, once}], %% reset to receive next tcp message
-    ok = Transport:setopts(Socket, TcpOptions),
+    Transport:setopts(Socket, TcpOptions),
     case Data of
         [MsgType] ->
             process_msg(MsgType, State);
@@ -164,8 +167,41 @@ process_msg(?MSG_COMPLETE, State=#state{owner=Owner}) ->
 
 %% Send a response back to the aae_source worker
 
-send_reply(Msg, State=#state{socket=Socket, transport=Transport}) ->
+send_reply(Msg, State=#state{sender=Sender}) ->
     Data = term_to_binary(Msg),
-    ok = Transport:send(Socket, <<?MSG_REPLY:8, Data/binary>>),
+    Sender ! <<?MSG_REPLY:8, Data/binary>>,
     {noreply, State}.
 
+%%%===================================================================
+%%% TCP send helper
+%%%===================================================================
+
+%% Sending is performed in a separate process to ensure that the sink
+%% never blocks and can therefore always receive messages + reset the
+%% active once flag. If a separate process is not used, it is possible
+%% for both the source and sink to deadlock on a TCP send that each
+%% stall waiting for the remote side to receive.
+%%
+%% There is currently no backpressure between the sink and this helper
+%% process. Adding backpressure runs the risk of once again hitting
+%% the aforementioned deadlock scenario.
+%%
+%% The correct approach is to add backpressure to the actual fullsync
+%% protocol itself. This remains as future work.
+%%
+%% NOTE: the SSL transport wraps the underlying socket in a gen_server,
+%% and all socket functions route through this server. As such, splitting
+%% sending and receiving between two processes does not help. The SSL
+%% transport is thus known to *not* be safe and *can* deadlock.
+
+spawn_sender(Transport, Socket) ->
+    spawn_link(?MODULE, sender_init, [Transport, Socket]).
+
+sender_init(Transport, Socket) ->
+    sender_loop({Transport, Socket}).
+
+sender_loop(State={Transport, Socket}) ->
+    receive Msg ->
+            ok = Transport:send(Socket, Msg)
+    end,
+    ?MODULE:sender_loop(State).
--- a/src/riak_repl_aae_source.erl
+++ b/src/riak_repl_aae_source.erl
@@ -345,10 +345,11 @@ key_exchange(start_key_exchange, State=#state{cluster=Cluster,
     %% always starts as the empty list. KeyDiffs is a list of hashtree::keydiff()
     Bloom = State#state.bloom,
     Count0 = State#state.diff_cnt,
+    Limit = app_helper:get_env(riak_repl, fullsync_direct, ?GET_OBJECT_LIMIT),
     AccFun = fun(KeyDiffs, Acc0) ->
             Count = case Acc0 of [C] when is_integer(C) -> C; _ -> 0 end,
             FoldFun =
-              case (Count0+Count) > ?GET_OBJECT_LIMIT of
+              case (Count0+Count) > Limit of
                   %% Gather diff keys into a bloom filter
                   true  -> fun(KeyDiff, AccIn) ->
                                    accumulate_diff(KeyDiff, Bloom, AccIn, State)
--- a/src/riak_repl_aae_source.erl
+++ b/src/riak_repl_aae_source.erl
@@ -45,11 +45,16 @@
                 local_lock = false :: boolean(),
                 owner       :: pid(),
                 proto       :: term(),
-                bloom       :: reference() %% ebloom
+                bloom       :: reference(), %% ebloom
+                diff_cnt=0  :: non_neg_integer()
                }).
 
 %% Per state transition timeout used by certain transitions
 -define(DEFAULT_ACTION_TIMEOUT, 300000). %% 5 minutes
+
+%% the first this many differences are not put in the bloom
+%% filter, but simply sent to the remote site directly.
+-define(GET_OBJECT_LIMIT, 1000).
 
 %%%===================================================================
 %%% API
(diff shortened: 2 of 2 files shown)


krestenkrab:
Maybe we should use `riak_kv_vnode:local_get(PartitionIndex,BKey)` rather than `Client:get(..., [{n_val,1}...`.  This would even further reduce the cluster load, as we don't risk going over the network to get the value.  We are, after all, running in the local vnode that is supposed to have the value.  The only obvious downside I can see is that `local_get`doesn't let you specify a timeout.

krestenkrab:
I'd like to propose that we rename `fullsync_direct` to be `fullsync_direct_limit`.  In our next round of changes we have a sibling to this config option which is the cutoff percentage (of #keys in partition); and that one I'd like to call `fullsync_direct_percentage_limit`.

src/riak_repl_aae_source.erl
```diff
@@ -30,6 +30,15 @@
 -type index() :: non_neg_integer().
 -type index_n() :: {index(), pos_integer()}.
 
+-record(exchange, {mode   :: inline | buffered,
+                   buffer :: ets:tid(),
```
jonmeredith, line 34: buffer can also be undefined.

src/riak_repl_aae_sink.erl
```diff
@@ -164,8 +167,41 @@ process_msg(?MSG_COMPLETE, State=#state{owner=Owner}) ->
...
+%% NOTE: the SSL transport wraps the underlying socket in a gen_server,
+%% and all socket functions route through this server. As such, splitting
+%% sending and receiving between two processes does not help. The SSL
+%% transport is thus known to *not* be safe and *can* deadlock.
+
+spawn_sender(Transport, Socket) ->
+    spawn_link(?MODULE, sender_init, [Transport, Socket]).
+
+sender_init(Transport, Socket) ->
+    sender_loop({Transport, Socket}).
```
jonmeredith, line 201: Be kind to your fellow debuggers and set something in the process dictionary we can postmortem what this process is and why it existed.

src/riak_repl_aae_source.erl
```diff
@@ -305,29 +353,60 @@ key_exchange(start_key_exchange, State=#state{cluster=Cluster,
...
     %% wait for differences from bloom_folder or to be done
     {next_state, compute_differences, State}.
 
+%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+
+async_get_bucket(Level, Bucket, IndexN, State) ->
+    send_asynchronous_msg(?MSG_GET_AAE_BUCKET, {Level,Bucket,IndexN}, State).
+
+wait_get_bucket(_Level, _Bucket, _IndexN, State) ->
+    Reply = get_reply(State),
```
jonmeredith, line 371: Any need for Reply variable?

src/riak_repl_aae_source.erl
```diff
@@ -305,29 +353,60 @@ key_exchange(start_key_exchange, State=#state{cluster=Cluster,
...
+    Reply.
+
+async_get_segment(Segment, IndexN, State) ->
+    send_asynchronous_msg(?MSG_GET_AAE_SEGMENT, {Segment,IndexN}, State).
+
+wait_get_segment(_Segment, _IndexN, State) ->
+    Reply = get_reply(State),
```
jonmeredith, line 378: unneeded var

src/riak_repl_aae_sink.erl
krestenkrab, line 201: I suppose this means: use `proc_lib`.

src/riak_repl_aae_source.erl
```diff
@@ -494,7 +597,7 @@ send_missing(RObj, State=#state{client=Client, wire_ver=Ver, proto=Proto}) ->
     end.
 
 send_missing(Bucket, Key, State=#state{client=Client, wire_ver=Ver, proto=Proto}) ->
-    case Client:get(Bucket, Key, 1, ?REPL_FSM_TIMEOUT) of
+    case Client:get(Bucket, Key, [{r, 1}, {timeout, ?REPL_FSM_TIMEOUT}, {n_val, 1}]) of
         {ok, RObj} ->
```
krestenkrab, line 626: I was reading around some old documentation for full sync, and came across the `max_get_workers` full sync configuration property.  This seems to have fallen through the cracks in the `v3` full sync scheme; perhaps this may also be a bottle neck for performance.

src/riak_repl_aae_sink.erl
jonmeredith, line 201: That's what I used to do until the inestimable @Vagabond pointed out that anything started with `proclib` should comply with OTP Design principles and handle ``sys` messages etc.
Maybe a poor mans version would set the the procdict entries so that etop etc showed them nicely.

src/riak_repl_aae_source.erl
jonmeredith, line 626: I think that setting controlled how many workers the fullsync source used for retrieving vclocks and objects to send.

src/riak_repl_aae_source.erl
```diff
@@ -443,36 +526,56 @@ replicate_diff(KeyDiff, Acc, State=#state{index=Partition}) ->
...
+                                        bloom=Bloom,
+                                        count=Count,
+                                        limit=Limit},
+                     State) ->
+    if Count < Limit ->
+            %% Below threshold, handle directly
+            Exchange2 = handle_direct(Mode, Bucket, Key, Exchange, State),
+            Exchange2#exchange{count=Count+1};
+       true ->
+            %% Past threshold, add to bloom filter for future bloom fold
```
jonmeredith, line 565: Could build the bloom filter here on demand instead of in `maybe_accumulate_key/4` - would save allocating around 100Mb on top of the potential stack used by compare2 of 16mb.

```
(riak@127.0.0.1)84>     NumKeys = 10000000.
(riak@127.0.0.1)84>     {ok, Bloom} = ebloom:new(NumKeys, 0.01, random:uniform(1000)).
(riak@127.0.0.1)86> ebloom:size(Bloom).
95929552
```

Would be better for the frequent-small-syncs case.

krestenkrab, line 565: Right. This is also part of the 'estimate keys' extension to this.  Makes good sense to add that optimisation here as well, now that we have the 'fullsync_direct_limit' config option.

The 1M fixed limit is pretty bad for our primary use case w/ 2B keys, N=3, Q=128. That leaves ~50M keys in each partition; meaning that a 1M bloom has 50x chance of a false positive, i.e. a false positive probability of 0.5.  Fortunately, the actual number of diffs we'll see is very small, as we also run realtime sync.

jonmeredith:
Couldn't work out how to inline respond to Kresten's comment, but yes, I think it does need to use `riak_kv_vnode:local_get/2` as there are no guarantees you will hit the local vnode if you use a reduced n_val get (which is probably something we should also fix for CS).

jonmeredith:
Unrelated to the AAE changes, the mechanics for changing the controlling process are totally unnecessary for the `Remote` function defined in `key_exchange/2` as there are no messages delivered by an `{active, false}` socket.

src/riak_repl_aae_source.erl
```diff
@@ -305,29 +353,60 @@ key_exchange(start_key_exchange, State=#state{cluster=Cluster,
...
+    update_trees(continue, State2).
+
+maybe_send_direct(#exchange{mode=inline, count=Count, limit=Limit},
+                  #state{index=Partition}) ->
+    %% Inline sends already occured inline
+    Sent = erlang:min(Count, Limit),
+    lager:info("Directly sent ~b differences inline for partition ~p",
+               [Sent, Partition]),
+    ok;
+maybe_send_direct(#exchange{buffer=Buffer}, State=#state{index=Partition}) ->
```
jonmeredith, line 402: If you get here and you have a bloom filter, you do the direct gets (at the moment using get FSMs) for all of the entries.  Is there a good reason not to detect you will need to bloom fold anyway and just set the elements?

src/riak_repl_aae_source.erl
```diff
@@ -305,29 +353,60 @@ key_exchange(start_key_exchange, State=#state{cluster=Cluster,
...
+    %% Just move on to the next tree exchange since we accumulate
+    %% differences across all trees.
+    State2 = State#state{exchange=Exchange},
+    update_trees(continue, State2).
+
+maybe_send_direct(#exchange{mode=inline, count=Count, limit=Limit},
+                  #state{index=Partition}) ->
+    %% Inline sends already occured inline
+    Sent = erlang:min(Count, Limit),
+    lager:info("Directly sent ~b differences inline for partition ~p",
```
jonmeredith, line 399: Having multiple sent counts per partition is confusing for users and monitoring.  You will potentially have log messages that say

`Directly sent 1000 differences inline for partition 123456789
No differences, skipping bloom fold for partition 123456789`

`Directly sending 1000 differences inline for partition 123456789
No differences, skipping bloom fold for partition 123456789`

`Directly sent 1000 differences inline for partition 123456789
Bloom folding over ~54321 differences for partition 123456789`

`Directly sending 1000 differences inline for partition 123456789
Bloom folding over ~54321 differences for partition 123456789`

So… as a user, I would like just one log message that succinctly tells me what you completed sending (not intended sending) and if you like, you can add some details about the way you did it.

```
Fullsync sent 55321 differences for partition 123456789 in 13.4 seconds using [direct,bloom]
```

jonmeredith:
Spoke with @jtuple and @matthewvon on the path to merging.  This is relatively minimal change PR to a point release to address a significant pain point, so I'm accepting that we won't clean up some things I would like to.

Code review looks logically good.  The only major thing that needs to be fixed is the switch to `riak_kv_vnode:local_get/2`.  The reporting improvements and general refactoring can be done on the 2.1 schedule.  Depending on how much improvement is needed to the protocol the module may be significantly refactored anyway.  Joe said the AAE estimation branch moved the bloom filter creation, so if we get that merged it will take care of that.

Will approve once I've run some of the r_t AAE tests.

src/riak_repl_aae_source.erl
krestenkrab, line 399: Rune and I also talked about reporting. It would make sense to report something that can help guide selecting configuration values. In particular direct/bloom cutoff limit, because if often overshoot by a small number - say the average # diffs is 1200 and the limit is configured to 1000 then it makes sense to set the cutoff at ~1400 perhaps if that means that you avoid a huge fold in most cases.  

One heuristic to apply would be to - if folding - count the actual number of objects folded over, and if the ratio of diffs/fold-count is say below 1% issue a warning that it may make sense to raise the full sync_direct_limit.

jonmeredith:
Compared riak_test output to 2.0.0 and get similar results for the intermittent cases. A little work to do there, but ok.

Like the other pull requests, this is targeted at develop, not 2.0.0

Plus one to merge once target is fixed.

engelsanchez:
I believe this is not needed anymore. All changes made it to 2.0, and today 2.0 has been merged to develop.

PR #617, opened 2014-09-18
Feature - fssink pool

jonmeredith:
Adds pool of fullsync workers for storing incoming objects.

Both the key list and AAE strategies block the keylist/aae sink process while writing objects.  This adds a pool of workers to absorb the work.


jonmeredith:
Pulled in too much other crap.  Reopening.

PR #618, opened 2014-09-18
Added a worker pool for fullsync sinks.

jonmeredith:
Although keylisting has a pool for issuing gets, both the
AAE and the Keylist strategies serialize writing objects
from the sink process.  This adds a pool with a default size
of 100 shared across all fullsync sinks that make the writes async.

No difference in safety properties, neither version checks
the response from the put FSM.


jonmeredith:
Assigned to Joe.

jtuple:
+1 merge. More testing underway, but current tests show worker pool to always be a win: often times making things 150-160% faster.

Note: will be ported to 2.0 branch and merged against 2.0

jonmeredith:
Included in #620

PR #621, opened 2014-09-24
Refactor cluster manager and connection manager to use otp behaviors.

lordnull:
Splits out responsibilities a bit and uses otp behaviors (mainly gen_fsm) to help code-readability and introspection of running systems.

Riak_test is basho/riak_test/pull/684 .

### DIFF
--- a/src/riak_core_service_conn.erl
+++ b/src/riak_core_service_conn.erl
@@ -15,7 +15,7 @@
 -export([code_change/4, terminate/3]).
 
 -record(state, {
-    transport, socket, remote_rev, remote_caps, init_args
+    transport, transport_msgs, socket, remote_rev, remote_caps, init_args
 }).
 
 %% ===============
@@ -41,7 +41,8 @@ init({Listener, Socket, Transport, InArgs}) ->
     ok = ranch:accept_ack(Listener),
     ok = Transport:setopts(Socket, ?CONNECT_OPTIONS),
     ok = Transport:setopts(Socket, [{active, once}]),
-    State = #state{transport = Transport, socket = Socket, init_args = InArgs},
+    TransportMsgs = Transport:messages(),
+    State = #state{transport = Transport, transport_msgs = TransportMsgs, socket = Socket, init_args = InArgs},
     gen_fsm:enter_loop(?MODULE, [], wait_for_hello, State).
 
 %% ===============
@@ -78,7 +79,7 @@ handle_event(_Req, StateName, State) ->
 %% handle_info
 %% ===============
 
-handle_info({_TransTag, Socket, Data}, wait_for_hello, State = #state{socket = Socket}) when is_binary(Data) ->
+handle_info({TransOk, Socket, Data}, wait_for_hello, State = #state{socket = Socket, transport_msgs = {TransOk, _, _}}) when is_binary(Data) ->
     case binary_to_term(Data) of
         {?CTRL_HELLO, TheirRev, ThierCaps} ->
             SSLEnabled = app_helper:get_env(riak_core, ssl_enabled, false),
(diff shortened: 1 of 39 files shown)


src/riak_core_service_conn.erl
```diff
@@ -0,0 +1,236 @@
...
+
+handle_event(_Req, StateName, State) ->
+    {next_state, StateName, State}.
+
+%% ===============
+%% handle_info
+%% ===============
+
+handle_info({_TransTag, Socket, Data}, wait_for_hello, State = #state{socket = Socket}) when is_binary(Data) ->
+    case binary_to_term(Data) of
```
krestenkrab, line 83: Urgh! We need to play it more safe here, because we're decoding random data coming off the internet.  I suggest limiting the size of `Data`, and also using `binary_to_term(Data, [safe])`, so that someone posing as a remote cluster cannot inject random new atoms into this VM.  Remember this all happens _before_ SSL upgrade.

It's a little tricky, because we want to be able to handle the case when they have capabilities we don't know of, i.e. atoms that are not necessarily present in this VM.  So maybe just limiting `byte_size(Data)` to something like a few K should suffice...

src/riak_core_service_conn.erl
```diff
@@ -0,0 +1,236 @@
...
+                    State2 = State#state{transport = NewTransport, socket = NewSocket, remote_caps = ThierCaps, remote_rev = TheirRev},
+                    {next_state, wait_for_protocol_versions, State2}
+            end;
+        _Else ->
+            lager:debug("Invalid hello: ~p", [Data]),
+            {stop, {error, invalid_hello}, State}
+    end;
+
+handle_info({_TransTag, Socket, Data}, wait_for_protocol_versions, State = #state{socket = Socket}) when is_binary(Data) ->
+    case binary_to_term(Data) of
```
krestenkrab, line 106: Same applies here regarding `binary_to_term(..., [safe])` and/or size limiting. If we're not on SSL, then this might well be data coming from a random client sitting somewhere on the net.

src/riak_repl_console.erl
```diff
@@ -321,7 +320,8 @@ showClusterConn({Remote,Pid}) ->
             CAddr = choose_best_addr(Remote, ClientAddr),
             io:format("~-20s ~-20s ~-15s ~p (via ~s)~n",
                       [ConnName, Name, PidStr,IPs, CAddr])
-    after 2 ->
```
krestenkrab, line 324: From reading `riak_core_cluster_conn.erl`, it looks like the status could also be `initiating_connection`, `waiting_for_cluster_name`, and `waiting_for_cluster_members`.  Am I missing something?

krestenkrab:
All in all, minor things that need fixing, except for the indiscriminate decoding of `binary_to_term` on the open internet.

src/riak_core_service_conn.erl
lordnull, line 83: I agree that this is bad thing™ and needs to be fixed. Both sides use 'term_to_binary', so I'll need to inspect the code closely to ensure that we never send an atom that the remote side may not already have. My particular concern is the capabilities exchange step. If that's the case, I'm not sure anything short of a complete revamp of the handshake protocol would suffice.

krestenkrab:
After running (on code derived from this in branch `enhance/krab/cluster-realtime-rebalance`), we ran into two problems fixed with the following commits: 2c1744256c59e3804d50c1022beda4b3c4365aca and f41234feab6d202b355d3435d4488705b722dc82.

lordnull:
I've cherry-picked those two commits in on my local machine, though I'm not sure how that will effect your branch; whether you want to rebase it or keep the commits there (and me ditching the cherry-pick).

krestenkrab:
No problem, @rsltrifork will work it out.

jonmeredith:
Looks like unit tests passed, but were marked as failure on the builder because it had to kill gold rush.

```
=======================================================
  All 189 tests passed.
Cover analysis: /usr/home/buildbot/masters/riak/test-riak_repl/build/.eunit/index.html
{"Kernel pid terminated",application_controller,"{application_terminated,goldrush,killed}"}

Crash dump was written to: erl_crash.dump
Kernel pid terminated (application_controller) ({application_terminated,goldrush,killed})
```

It also gave two dialyzer errors

```
==> build (dialyzer)
riak_core_connection.erl:171: The call ranch_tcp:send(Socket::port(),Hello::binary()) breaks the contract (inet:socket(),iolist()) -> 'ok' | {'error',atom()}
riak_core_connection.erl:263: Expression produces a value of type 'ok' | {'error',_}, but this value is unmatched
Unknown types:
  proplist:proplist/0
make: *** [dialyzer] Error 1
```

Can you investigate please @lordnull so we can merge this.

jonmeredith:
Passes Dialyzer now as of http://buildbot.bos1/builders/test-riak_repl/builds/1254, checking why GoldRush has problems on eunit shutdown.

jonmeredith:
@borshop: retry

lordnull:
Mistargeted pr, use #631 instead.

PR #626, opened 2014-10-10
Bugfix mw chatty transiet aae fs failures

lordnull:
There are cases where the aae full sync strategy will fail, but will be able to recover later, such as when the aae trees are not yet built. Until the cause of the transient failure is rectified (or the partition is restarted too many times), the crash log will be filled with useless messages. This PR introduces the use of a message 'soft_exit' to an owner process (passed into the start functions), and then exits the aae worker normally. This allows the owner process (fscoordinator) to retry the partition, but does not spam the crash logs.

~~The PR works, but an automated test is not yet complete.~~

The aae full sync test has been updated: basho/riak_test/pull/691
- [X] Edit/create riak_test to ensure feature works.


src/riak_repl2_fscoordinator.erl
```diff
@@ -324,7 +331,7 @@ handle_cast(stop_fullsync, State) ->
         largest_n = undefined,
         owners = [],
         partition_queue = queue:new(),
-        retries = dict:new(),
+        %retries = dict:new(),
```
jonmeredith: Cruft

jonmeredith:
I don't think this will work in a mixed cluster.  Not sure if you saw the approach I outlined in the comments for https://bashoeng.atlassian.net/browse/RIAK-1235

If fscoordinator is running on a 1.3.X/1.4.10 or below node, it will try and start up processes on remote nodes by calling start_link/3 which will not exist on the older node and fail.

https://github.com/basho/riak_repl/pull/626/files#diff-5075ea65b0560166a4e4684a839326b2R40

If fscoordinator is running on pre-1.4.11 nodes, and fssource is running on 1.4.11+ then it will send unhandled messages to undefined, and exit normally making the fscoordinator think it has succeeded when it has not.

Why not add a step to the fscoordinator in `start_fssource` to call something like

```
    case riak_repl2_fssource_sup:enable(LocalNode, Partition, {Ip, Port}) of
        {ok, Pid} ->
            link(Pid),
            riak_repl2_fssource:report_soft_exits(Pid),
            Running = orddict:store(Pid, PartitionVal, State#state.running_sources),
            State#state{running_sources = Running};
```

And have

```
report_soft_exits(Pid) ->
    try gen_server:call(Pid, {report_soft_exits, self()}, timer:seconds(15)) of
       ok -> %% Hit catchall
          false;
       {reporting_to, Pid} ->
          true
    catch
      _:Reason -> %% Probably a timeout if overloaded
          {error, Reason}
    end.
```

On upgraded nodes, you could store it in the state and make a function that reports back
  {exit, soft_exit(Why, State), State}

```
soft_exit(Why, #state{soft_exits_to=undefined} = State) ->
  {stop, Why, State};
soft_exit(Why, #state{soft_exits_to=Pid} = State) ->
  Pid ! {soft_exit, self(), Why},
  {stop, normal, State}
```

Within a node between fssource/riak_repl_aae_source I think it is safe to just modify the start call, but you should verify there are not any external dependencies.

jonmeredith:
# riak_repl2_fscoordinator.erl

## Soft error counting

It would have been a smaller change to introduce a soft_retries dictionary to count
those and check for going over. That would have only meant changing in the new
handle_abnormal_exit function, rather than in multiple places in the module.

Instead the factor to use the partition_info record is spread all over the module.
Given the switch to keytake in so many places, it would be cleaner to flatten
it as a list of records instead of an orddict.

```
-record(partition_info, {
    pid, index, retries = 0, soft_retries = 0, node
}).
```

then accessor code could be simplified to

```
    PartitionEntry = lists:keytake(Pid, #partition_info.pid, State#state.running_sources),
    case PartitionEntry of
        false ->
            % late exit or otherwise non-existant
            {noreply, State};
        {value, PartitionInfo, Running} ->
             do_something(PartitionInfo)
    end
```

or even cleaner given how much the pattern is used, break out the accessor into a function

```
   case take_source(Pid, State) of
     {false, _RunningSources} ->
         {noreply, State}
     {PartitionInfo, RunningSources} ->
         do_something(PartitionInfo)
   end

```

There is at least one place where the new format is not handled correctly
https://github.com/basho/riak_repl/pull/626/files#diff-ff7718aa59949a972fa2d2fae40b11d0R329

```
    [begin
        unlink(Pid),
        riak_repl2_fssource:stop_fullsync(Pid),
        riak_repl2_fssource_sup:disable(node(Pid), Part)
    end || {Pid, {Part, _PartN}} <- State#state.running_sources],
```

Presumably `{Pid, {Part, _PartN}}` should be `{Pid, #partition_info(index = Part}}`
or if you flattened running_sources, `#partition_info{pid= Pid, index = Part}`.

# Soft error behavior

When soft errors are exceeded the partition is silently dropped from all lists.
Although the soft_retry count is updated, it should increment error_exits once
exceeded so that a status call can tell fullsync has not synchronized completely.

The log entry printed on completion does not include any information on
whether the current fullsync completed with error.

Existing behavior is for the rtdirty flag to be reset regardless of exits.
Should file a separate issue for that, outside the scope of this fix.

# Purgatory

What is the purpose of the purgatory queue, why does it need any different
behavior from the partition_queue?

# Exit limits

The existing setting to control how many times a partition can be retried is
`max_fssource_retries`, would recommend using something similar like `max_fssource_soft_retries`.

The `if` statements, seem ugly, how about writing a small helper like

```
hit_retry_limit(Count, infinity) ->
    false;
hit_retry_limit(Count, Limit) ->
    Count >= Limit.
```

That way the duplicate queing cases could be collapsed into a pair of cases.

If purgatory is not needed, you could break it into

```
{Count, Limit, State} = case ExitType of
    soft_exit ->
         %% update soft counts and get limits
    'EXIT'
         %% update hard exit counts and limits
    end
case hit_retry_limit(Count, Limit) of
  true ->
     %% Log failed message
  false ->
     %% Log retry message and add to partition_queue
```

src/riak_repl2_fssource.erl
```diff
@@ -137,9 +145,22 @@ handle_call(stop_fullsync, _From, State=#state{fullsync_worker=FSW,
     {reply, ok, State};
 handle_call(legacy_status, _From, State=#state{fullsync_worker=FSW,
                                                socket=Socket}) ->
-    Res = case is_pid(FSW) andalso is_process_alive(FSW) of
-        true -> gen_fsm:sync_send_all_state_event(FSW, status, infinity);
-        false -> []
+    lager:debug("Sending status to ~p", [FSW]),
+    Res = if
```
jonmeredith: Why was the check for `is_process_alive/1` removed before making the legacy_status request? Unnecessary as the sync_send_all_state checks? Why the change to an ugly `if`?

Is this change related to soft error handling?

jonmeredith:
@lordnull done reviewing for now.  -1 until we resolve the rolling upgrade issue, lets talk about the best path forward in the morning.

src/riak_repl2_fssource.erl
lordnull: While testing the initial soft_exit change, I found that it was possible for the is_process_alive/1 to return true, but by the time it got to the sync_send_all_state_event/3, the process was no longer alive due to a soft_exit. The is_process_alive/1 was not doing it's job (ensure we only try to get stats from a living worker), so I removed it in favor of try/catch.

The more I've worked with erlang, the less 'if' has seemed ugly. It's a guard with a boolean output. Why is a case statement superior?

jonmeredith: Thanks for the explanation.

if is ugly to me because the fall you have to specify the fall through case
as true. I know others (CMJ, SeanC, Kresten) don't like it based on other
review feedback. In the rest of Erlang you'd specify that as an _Var on
something you didn't care about.  As they aren't used much by others I
think it risks tripping people up.

On Tue, Oct 14, 2014 at 12:32 PM, Micah notifications@github.com wrote:

> In src/riak_repl2_fssource.erl:
> 
> > @@ -137,9 +145,22 @@ handle_call(stop_fullsync, _From, State=#state{fullsync_worker=FSW,
> >      {reply, ok, State};
> >  handle_call(legacy_status, _From, State=#state{fullsync_worker=FSW,
> >                                                 socket=Socket}) ->
> > -    Res = case is_pid(FSW) andalso is_process_alive(FSW) of
> > -        true -> gen_fsm:sync_send_all_state_event(FSW, status, infinity);
> > -        false -> []
> > -    lager:debug("Sending status to ~p", [FSW]),
> > -    Res = if
> 
> While testing the initial soft_exit change, I found that it was possible
> for the is_process_alive/1 to return true, but by the time it got to the
> sync_send_all_state_event/3, the process was no longer alive due to a
> soft_exit. The is_process_alive/1 was not doing it's job (ensure we only
> try to get stats from a living worker), so I removed it in favor of
> try/catch.
> 
> The more I've worked with erlang, the less 'if' has seemed ugly. It's a
> guard with a boolean output. Why is a case statement superior?
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/basho/riak_repl/pull/626/files#r18848806.

## 

Jon Meredith
Chief Architect
Basho Technologies, Inc.
jmeredith@basho.com

lordnull:
## Soft Error Counting

The primary urge to move to using a record was for locality of information. Rather than needing to track data about a partition spread through out the state of the fscoordinator (retries in one dict, soft_retries in another, which node it's on in yet another data structure), it's all kept and updated in one record. That record gets updated and moved to various waiting areas and queues.

Not to tied to the idea, but it seemed a cleaner method going forward than just tacking on another bit of state in the coordinator to reset.

## Purgatory

While testing, I found that the soft_exit_retries was increasing very rapidly. It was retrying partitions faster than the cause for the retry was resolved since the failed partition was put back into the partition_queue immediately. By putting the failed partition in a purgatory queue and feeding the partitions back in more slowly, this lowers the amount of useless work done.

src/riak_repl2_fscoordinator.erl
```diff
@@ -465,6 +425,85 @@ handle_info(_Info, State) ->
...
+            SoftRetryLimit = app_helper:get_env(riak_repl, max_fssource_soft_retries, ?DEFAULT_SOURCE_SOFT_RETRIES),
+            SoftRetryCount = State#state.soft_retry_exits + 1,
+            if
+                SoftRetryLimit =:= infinity ->
+                    Purgatory = queue:in(Partition, State3#state.purgatory),
+                    {noreply, State3#state{purgatory = Purgatory, soft_retry_exits = SoftRetryCount}};
+
+                SoftRetryLimit >= Partition#partition_info.soft_retries ->
+                    lager:info("Discaring partition ~p since it has reached the soft exit retry limit of ~p", [Partition#partition_info.index, SoftRetryLimit]),
+                    {noreply, State3#state{soft_retry_exits = SoftRetryCount}};
```
jonmeredith: This should count as an error_exit as the fullsync did not complete for all vnodes.

lordnull:
I've done further testing with the r_t linked above and eunit. The occasionally failure I will get in r_t is related to transfers, not repl; when it gets past transfers the test passes.

I'm not getting the same errors on eunit as on the build machine; the errors I've gotten are related to rtq, while the build machine seems to have errors related to a test not properly cleaning itself up.

Where do we go from here?

lordnull:
I've tested this on a local rebase off of 1.4, and it still passes r_t still. Do we want to merge as is, see if the eunit tests can be fixed, or push the rebase and merge that?

cmeiklejohn:
Pushed some minor cleanup changes, please review, @lordnull and/or @jonmeredith:

https://github.com/basho/riak_repl/commit/d6a987b389ed71ab76e9e8a9790b1a56cd7a7a60
https://github.com/basho/riak_repl/commit/ec300dc3f8b018dc9b3dcfc067e4d4f263d70db3
https://github.com/basho/riak_repl/commit/9ec4c7d1511c9f74afdfc0ffe5917d9c04d8ff25
https://github.com/basho/riak_repl/commit/137bee8abd92094803be4826d1e24c674dabfb2e
https://github.com/basho/riak_repl/commit/94232f0f1f251ba2adb180e6b422ac54c58e13b9
https://github.com/basho/riak_repl/commit/9ec4c7d1511c9f74afdfc0ffe5917d9c04d8ff25

Let me know if you have any questions.

lordnull:
Looks good to me.

engelsanchez:
I will be finishing the review for this PR

src/riak_repl2_fscoordinator.erl
```diff
@@ -518,6 +489,83 @@ handle_info({'DOWN', Mon, process, Pid, Why}, #state{stat_cache = #stat_cache{wo
...
+    % stats
+    #state{partition_queue = PQueue} = State,
+
+    State2 = State#state{busy_nodes = NewBusies, running_sources = Running},
+    {ErrorCount, State3} = increment_error_dict(Partition, ExitType, State2),
+
+    case ExitType of
+        soft_exit ->
+            lager:debug("putting partition ~p in purgatory due to soft exit of ~p", [Index, Pid]),
+            receive
```
engelsanchez: Here we are blocking for an 'EXIT' message from the fssource, which we should be about to get. But there is a chance of the disterl reconnection happening at the wrong time and this message getting lost forever. That would make this block forever too.

lordnull: The intent of the line is to avoid the extra log spam the secondary redundant exit message would cause. Given that, what do you think would be a reasonable wait time for the exit message?

src/riak_repl2_fscoordinator.erl
```diff
@@ -518,6 +489,83 @@ handle_info({'DOWN', Mon, process, Pid, Why}, #state{stat_cache = #stat_cache{wo
...
+            SoftRetryLimit = app_helper:get_env(riak_repl, max_fssource_soft_retries, ?DEFAULT_SOURCE_SOFT_RETRIES),
+            SoftRetryCount = State#state.soft_retry_exits + 1,
+            if
+                SoftRetryLimit =:= infinity ->
+                    Purgatory = queue:in(Partition, State4#state.purgatory),
+                    {noreply, State4#state{purgatory = Purgatory, soft_retry_exits = SoftRetryCount}};
+
+                SoftRetryLimit < ErrorCount ->
+                    lager:info("Discaring partition ~p since it has reached the soft exit retry limit of ~p", [Partition#partition_info.index, SoftRetryLimit]),
+                    ErrorCount = State#state.error_exits + 1,
```
engelsanchez: @corbaguy spotted this: You really want ErrorCount1 here. It is matching against the already bound ErrorCount and will fail at runtime.

src/riak_repl2_fscoordinator.erl
```diff
@@ -518,6 +489,83 @@ handle_info({'DOWN', Mon, process, Pid, Why}, #state{stat_cache = #stat_cache{wo
...
+                    Dropped = [Partition#partition_info.index | State#state.dropped],
```
engelsanchez: Should be better to use State4 here instead of State, even though dropped has not changed for future proofing this. The same for lines 536 and 528.  Don't you love Erlang's single assignment :)

src/riak_repl2_fscoordinator.erl
engelsanchez: I suppose any timeout number is a guess, and that path will be very rarely touched in real life. So something like 1 min, etc would probably be OK. And normally these two messages will come at pretty much the same time.

Another alternative is to set up a monitor to the process here. Then in a function wait for EXIT optionally, and then DOWN. When you get the DOWN message, which is guaranteed, you know no more messages for that process will arrive. Since the process is likely already dead at this point, that should not take long normally.

lordnull: I prefer waiting on a known event than guessing at a timeout. Since we would be trying to monitor a dead process, the monitor option seems the most reliable and clean option.

src/riak_repl2_fscoordinator.erl
```diff
@@ -536,108 +581,81 @@ code_change(_OldVsn, State, _Extra) ->
...
+            erlang:cancel_timer(Tref),
             % we don't know for sure it's no longer busy until we get a busy reply
             NewBusies = sets:del_element(Node, State#state.busy_nodes),
             State2 = State#state{whereis_waiting = Waiting2, busy_nodes = NewBusies},
-            Partition2 = {Partition, N, Node},
+            Partition2 = PartitionInfo#partition_info{node = Node, whereis_tref = undefined},
             State3 = start_fssource(Partition2, Ip, Port, State2),
             start_up_reqs(State3)
     end;
 handle_socket_msg({location_busy, Partition}, #state{whereis_waiting = Waiting} = State) ->
```
engelsanchez, line 597: So this code path should only be touched if interacting with the special anya branches, right? It looks like dead code, isn't it? Shall we remove, please?

/cc @jonmeredith

src/riak_repl2_fscoordinator.erl
```diff
@@ -536,108 +581,81 @@ code_change(_OldVsn, State, _Extra) ->
...
             PQueue2 = queue:in(Partition2, PQueue),
-            NewBusies = sets:add_element(OldNode, State#state.busy_nodes),
+            NewBusies = sets:add_element(Partition2#partition_info.node, State#state.busy_nodes),
             State3 = State2#state{partition_queue = PQueue2, busy_nodes = NewBusies},
             start_up_reqs(State3)
     end;
 handle_socket_msg({location_busy, Partition, Node}, #state{whereis_waiting = Waiting} = State) ->
-    case proplists:get_value(Partition, Waiting) of
-        undefined ->
+    case lists:keytake(Partition, 1, Waiting) of
```
engelsanchez: This should be #partition_info.index! 1 seems wrong (it will be the record tag).

src/riak_repl2_fscoordinator.erl
```diff
@@ -401,68 +418,22 @@ handle_info({'EXIT', Pid, Cause},
...
-                                         retries = Retries,
-                                         running_sources = Running,
-                                         error_exits = ErrorExits},
-                    maybe_complete_fullsync(Running, State2);
-                _ -> %% have not run out of retries yet
-                    % reset for retry later
-                    lager:debug("Fssource rescheduling partition: ~p",
-                               [Partition]),
-                    PQueue2 = queue:in(Partition, PQueue),
-                    RetryExits = State#state.retry_exits + 1,
```
engelsanchez, line 445: Question: we used to increment the retry exits stats whenever we got a location_down message before. It seems that now we don't, for this case. Is that intentional? What is the meaning of this stat, exactly?

lordnull, line 445: It was an oversight. The state is how many times we've tried to get the node for a given partition, but the sink side claims that node is down. This could be due to handoff, or other transient condition. So the more correct thing to do is requeue the partition and try it again later.

src/riak_repl2_fscoordinator.erl
```diff
@@ -518,6 +489,80 @@ handle_info({'DOWN', Mon, process, Pid, Why}, #state{stat_cache = #stat_cache{wo
...
+            SoftRetryLimit = app_helper:get_env(riak_repl, max_fssource_soft_retries, ?DEFAULT_SOURCE_SOFT_RETRIES),
+            SoftRetryCount = State4#state.soft_retry_exits + 1,
+            if
+                SoftRetryLimit =:= infinity ->
+                    Purgatory = queue:in(Partition, State4#state.purgatory),
+                    {noreply, State4#state{purgatory = Purgatory, soft_retry_exits = SoftRetryCount}};
+
+                SoftRetryLimit < ErrorCount ->
+                    lager:info("Discaring partition ~p since it has reached the soft exit retry limit of ~p", [Partition#partition_info.index, SoftRetryLimit]),
+                    ErrorCount1 = State4#state.error_exits + 1,
```
engelsanchez: Micah, would you mind changing the name to ErrorExits. It would be less confusing, as this variable is not related to the original ErrorCount. Ted and I got a bit confused while we were going around reviewing the changes. ErrorExits is the name used for the same purpose on line 550.

src/riak_repl2_fssource.erl
```diff
@@ -51,9 +56,34 @@ cluster_name(Pid) ->
...
+            false;
+        true ->
+            true
+    catch
+        _What:Reason ->
+            lager:debug("Could not create soft link to ~p from ~p due to ~p", [Pid, self(), Reason]),
+            {error, Reason}
+    end.
+
+%% @doc Remote a soft link created
```
engelsanchez: While you are here, I suppose you could remove this line. I guess it's edoc for a function that was added, then removed.

engelsanchez:
+1 0bdb1ba

We've finished reviewing this PR. All problems found during the review have been addressed. All eunit tests pass for me locally (the builder seems to have a setup problem though) and the aae fullsync riak_test passes too.  I've also manually tested by interfering with AAE trees manually while AAE fullsyncs are running. As expected, the errors that used to spam the logs are only info entries now and the soft retries are incremented instead.  This is good to go!

PR #636, opened 2014-11-18
Implement soft_exit, primarily for aae_fullsyn.

lordnull:
Problem: transient failures of aae, such as trees not yet built or locks not
being aquired, would cause an aae fullsync process to exit abnormally. This
could happen several times in a row, creating log spam.

Resolution: the concept of soft_exit. A soft_exit is a message sent from a soon
to be exiting process to a soft_linked process. The exiting process would then
exit normally, while any soft_linked processes could handle the soft_exit
message in a similar fashion as an exit message. This would indicate an exit
reason that should be handled, but not bad enough to have the system logger
know about it.

The soft_exit message sent from the aae worker to the fscoordinator is
as simple as `{soft_exit, pid(), term()}'.

The current implementation is not generic. There can only one soft_link to
the aae, and there's no general mechanism to use soft_link's or soft_exits
elsewhere in the code base. Sorry.

Another change rolled into this is consistent use of a #partition_info record
in the fscoordinator, and error tracking the fscoordinator's state. By swapping
to useing a single data structure in the partition queue, whereis waiting list,
and purgatory queues it makes it easier to understand the fscordinator (as
there is less code modify structures).

This is a forward port of the fix done for 1.4 (#626). A riak_test exists at https://github.com/basho/riak_test/pull/703. Conflicts favor existing code
where it does not directly effect the fix.

Conflicts:
    Makefile
    rebar.config
    src/riak_repl2_fssource.erl
    src/riak_repl2_rtq_proxy.erl
    src/riak_repl_aae_source.erl
    test/riak_core_cluster_mgr_tests.erl

### DIFF
--- a/include/riak_repl.hrl
+++ b/include/riak_repl.hrl
@@ -28,6 +28,9 @@
 -define(DEFAULT_SOURCE_RETRIES, infinity).
 %% How many times we should retry when failing a reservation
 -define(DEFAULT_RESERVE_RETRIES, 0).
+%% How many times during a fullsync we should retry a partion that has sent
+%% a 'soft_exit' message to the coordinator
+-define(DEFAULT_SOURCE_SOFT_RETRIES, infinity).
 %% 20 seconds. sources should claim within 5 seconds, but give them a little more time
 -define(RESERVATION_TIMEOUT, (20 * 1000)).
 -define(DEFAULT_MAX_FS_BUSIES_TOLERATED, 10).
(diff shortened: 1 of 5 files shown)


engelsanchez:
Cherry picked commits f565ea2 and 7da94e8 from the 1.4 branch that fix an issue that crash the process due to incorrect syntax and the regression with the error_exits/retry_exits stats. The stats can be verified with the repl_location_failures riak test, which will fail without that commit.

engelsanchez:
Added another commit that fixes dialyzer warnings, one of which was an actual bug. @lordnull can you take a look at it? Initializing the process with an owner actually never put the owner in the state. Not sure if that was used yet.

lordnull:
:+1: new changes look good to me and don't explode the world.

engelsanchez:
@borshop merge

@lordnull Unfortunately we should have done this in the 2.0 branch first :(. I'll open a PR for that branch now. So. Many. Branches.

PR #638, opened 2014-11-21
Fixed invalid call to update error count track.

lordnull:
Increment_error_dict expects the partition, elementN of error dict, and the
state. It pulls the dict out of the state so it put it back in place, thus just
returning the state. So this call that passed the dict in was wrong.


engelsanchez:
The change looks correct (and the previous version obviously broken). All eunit tests still pass locally. The riak_test where this error was found has other failures unfortunately, but not the crash related to this clause.

:+1:

PR #639, opened 2014-11-26
Fix error/retry exit counts on location down msgs

engelsanchez:
When a partition is not available, perhaps after a number of retries,
the error exits stat should be incremented. Also, the retry exits stat
should be incremented on each retry.  This was discovered when
backporting the repl_location_failures riak_test.

To verify this fix, use https://github.com/basho/riak_test/pull/706

The test fails without this change complaining about the error exits, passes with it.


tburghart:
+1

engelsanchez:
Thanks Ted!

PR #389, opened 2013-08-27
Cleanup rt_dirty behavior

bookshelfdave:
originally https://github.com/basho/riak_repl/pull/333

PR #643, opened 2014-12-09
Bugfix/mw/better rt dirty

lordnull:
A rebase with a couple of fixes of https://github.com/basho/riak_repl/pull/389. Requires updates to r_t replication2_dirty found in https://github.com/basho/riak_test/pull/709.

### DIFF
--- a/src/riak_repl_stats.erl
+++ b/src/riak_repl_stats.erl
@@ -140,8 +140,10 @@ rt_dirty() ->
                         riak_repl2_fscoordinator:node_dirty(node())
                       catch
                         _:_ ->
-                         lager:debug("Failed to notify coordinator of rt_dirty status")
-                      end
+                         %% This could be triggered on startup if the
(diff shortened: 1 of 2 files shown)


src/riak_repl_stats.erl
```diff
@@ -140,8 +140,10 @@ rt_dirty() ->
                         riak_repl2_fscoordinator:node_dirty(node())
                       catch
                         _:_ ->
-                         lager:debug("Failed to notify coordinator of rt_dirty status")
-                      end
+                         %% This could be triggered on startup if the
+                         %% fscoordinator isn't running
+                         lager:warning("Failed to notify coordinator of rt_dirty status.")
```
engelsanchez: Since we are logging a warning, I think we should also include the exact error that triggered it.

bsparrow435:
Unsure what problems this solves. Going to look more at it later.

Issue #615, opened 2014-09-10
SNMP stats crash if no repl2 leader


engelsanchez:
The SNMP stats poller process was observed to crash on a bad match on this line:

https://github.com/basho/riak_repl/blob/1.4.10/src/riak_repl_console.erl#L956

The stat for fullsync_coordinator_srv there was `{[], []}` instead of a list, so the foldl function head doesn't match.  That value should always be a list.  This is the culprit. It should return an empty list if no leader is found instead:

https://github.com/basho/riak_repl/blob/1.4.10/src/riak_repl2_fscoordinator_serv.erl#L55

via

https://github.com/basho/riak_repl/blob/1.4.10/src/riak_repl_console.erl#L872
https://github.com/basho/riak_repl/blob/1.4.10/src/riak_repl_console.erl#L137

Zendesk ticket: zd://8685

PR #645, opened 2014-12-17
Added test and fix to coord_serv not giving list for status.

lordnull:
Fix for #615. For those reviewing, this fix may not actually be enough.

While looking through the code, I noticed that the leader node isn't actually used, it's just checked for. After that, the stats gathered are just for the local node. So, is the current behavior correct at all?


engelsanchez:
Yes, you are right. It is not used. That seems to have been left behind from the original commit: https://github.com/basho/riak_repl/commit/989d79a30c1fb9e60b0c1136cc976cbaf616f988

I vote for dropping the check entirely. Thanks for looking at this!

lordnull:
The second part to the question is do we want to alter the behavior so it asks all the nodes for stats, or keep it node-local?

lordnull:
For posterity: final decision was to defer determining if local only data was valid to later. Fix the error here first (leader existence causing crash). Open a new issue to determine if we should be trying to gather all node's stats.

engelsanchez:
:+1: 39fa59f

I hit rebuild on buildbot as the eunit tests run was incomplete. I'm not sure why.

engelsanchez:
@borshop merge

PR #648, opened 2014-12-23
Remove partition from purgatory when giving up

engelsanchez:
When a partition has hit the soft exit limit, we add it to the dropped
list, but forgot to remove it from the purgatory list. So it may
actually be retried later.

Also, using a big hammer to silence noisy lager output during EQC tests.

This should fix the cause of this riak_test failure: http://giddyup.basho.com/#/projects/riak_ee/scorecards/115/115-1668-repl_aae_fullsync-centos-6-64/65935/artifacts/1226179

Where you can see partitions being "discared" when they hit the soft limit, but then tried again later. btw, that typo is also fixed by this PR.

/cc @lordnull

### DIFF
--- a/src/riak_repl2_fscoordinator.erl
+++ b/src/riak_repl2_fscoordinator.erl
@@ -543,10 +543,14 @@ handle_abnormal_exit(ExitType, Pid, _Cause, {value, PartitionWithSource, Running
                     {noreply, State4#state{purgatory = Purgatory, soft_retry_exits = SoftRetryCount}};
 
                 SoftRetryLimit < ErrorCount ->
-                    lager:info("Discaring partition ~p since it has reached the soft exit retry limit of ~p", [Partition#partition_info.index, SoftRetryLimit]),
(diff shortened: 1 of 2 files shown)


lordnull:
Code looks good, but repl_aee_fullsync test hasn't passed the last couple of times I've tried. currently trying to figure out if it's a problem with my set-up or these changes.

```
================ repl_aae_fullsync failure stack trace =====================
{{badmatch,{error,notfound}},
 [{repl_aae_fullsync,difference_test,0,
                     [{file,"tests/repl_aae_fullsync.erl"},{line,370}]},
  {repl_aae_fullsync,confirm,0,
                     [{file,"tests/repl_aae_fullsync.erl"},{line,41}]},
  {riak_test_runner,return_to_exit,3,
                    [{file,"src/riak_test_runner.erl"},{line,159}]}]}
============================================================================
```

lordnull:
:+1: 31a1657

Got the r_t to pass, as well as dialyzer and eunit.

engelsanchez:
Thanks @lordnull ! Now I guess we need to wait for the builder problem to be fixed to rerun the tests and allow the merge.

engelsanchez:
@borshop merge

PR #653, opened 2015-01-13
Issue Riak-1338

tburghart:
Cleaned up the care and feeding of the fake_sink test server to use a dynamically allocated port.  Also attempt to make sure it gets shut down cleanly so that the reported address-in-use error would be less likely, though the port could be in use otherwise so a dynamic one is superior all around.

Revised the riak_repl2_rtsource_conn eunit test to throw an error on failure, which is what eunit is looking for to determine pass/fail.

Removed a block of "receive" code at the end of the riak_repl2_rtsource_conn test that always returned a timeout error, as nothing was sending it a message at that point.

Refactored a teensy bit of test code so the names make sense.

### DIFF
--- a/src/riak_repl2_rtsource_conn.erl
+++ b/src/riak_repl2_rtsource_conn.erl
@@ -399,88 +399,104 @@ schedule_heartbeat(State) ->
 %% ===================================================================
 -ifdef(TEST).
 
--define(SINK_PORT, 5007).
-
 riak_repl2_rtsource_conn_test_() ->
-    {spawn,
-     [
-      {setup,
-       fun setup/0,
-       fun cleanup/1,
-       [
-        fun cache_peername_test_case/0
-       ]
-      }]}.
+    {spawn, [{
+        setup,
+        fun setup/0,
+        fun cleanup/1,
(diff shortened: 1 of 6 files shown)


lordnull:
:+1: e1c2ccf

Code looks better, and all tests pass now.

Basho-JIRA:
[~tburghart] is this intended to be included in 2.0.5 or 2.0.6? 

**_[posted via JIRA by Derek Somogyi]**_

cmeiklejohn:
I don't know what was _intended_, but this is most certainly in 2.0.5, given it's already been merged to the 2.0 branch.

```
[cmeiklejohn@Christophers-MacBook-Pro riak_ee ((riak_ee-2.0.5))]$ cd deps/riak_repl; git log | grep 329c766
Merge: 329c766 5573645
commit 329c76626ffc49b4414ed72ce4e34a5e54fc3e7e

```

Basho-JIRA:
This is already included in 2.0.5.

To be clear, this affects test code only, it doesn't touch production code. 

**_[posted via JIRA by Ted Burghart]**_

PR #655, opened 2015-01-22
AAE FS Automatic creation of hashtrees 

lixen:
Fullsync_strategy is configured to aae
AAE is set to off

1) AAE trees are completely built.
2) The AAE FS completes successully
3) AAE trees are removed as we shouldn't let the soon outdated AAE trees linger.
https://bashoeng.atlassian.net/browse/RIAK-1134

Some performance numbers:
https://docs.google.com/a/basho.com/spreadsheets/d/1YW3yqhX_tJd_PbgLiUFJnT585No50-mWtnfbhwSdKwg/edit#gid=0

### DIFF
--- a/src/riak_repl_aae_sink.erl
+++ b/src/riak_repl_aae_sink.erl
@@ -123,7 +123,7 @@ code_change(_OldVsn, State, _Extra) ->
 
 %% replies: ok
 process_msg(?MSG_INIT, Partition, State) ->
-    case riak_kv_vnode:hashtree_pid(Partition) of
+    case riak_kv_vnode:hashtree_pid(Partition, riak_kv_entropy_manager:enabled()) of
         {ok, TreePid} ->
             %% monitor the tree and crash if the tree goes away
             monitor(process, TreePid),
--- a/src/riak_repl_aae_sink.erl
+++ b/src/riak_repl_aae_sink.erl
@@ -131,7 +131,7 @@ process_msg(?MSG_INIT, Partition, State) ->
             riak_repl2_fs_node_reserver:claim_reservation(Partition),
             case riak_kv_entropy_manager:enabled() of
                 false ->
-                    riak_kv_index_hashtree:poke(TreePid);
+                    riak_kv_index_hashtree:build(TreePid);
                _ ->
                     ok
             end,
@@ -164,7 +164,7 @@ process_msg(?MSG_UPDATE_TREE, IndexN, State=#state{tree_pid=TreePid}) ->
 process_msg(?MSG_LOCK_TREE, State=#state{tree_pid=TreePid}) ->
     %% NOTE: be sure to die if tcp connection dies, to give back lock
     case riak_kv_index_hashtree:wait_for_lock(TreePid, fullsync_source) of
-        not_built ->
+        building ->
             %% Wait for lock
             {noreply, State};
         Response ->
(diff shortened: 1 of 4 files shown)


src/riak_repl_aae_source.erl
```diff
@@ -145,8 +145,9 @@ handle_info({ssl_error, Socket, Reason}, _StateName, State) ->
     lager:error("AAE source ssl connection to ~p closed unexpectedly with: ~p",
                 [State#state.cluster, Reason]),
     {stop, {ssl_error, Socket, Reason}, State};
-handle_info(_Info, StateName, State) ->
-    {next_state, StateName, State}.
+handle_info({hashtree_lock,_}, prepare_exchange, State) ->
+    gen_fsm:send_event(self(), start_exchange),
+    {next_state, prepare_exchange, State#state{local_lock=true}}.
```
engelsanchez: I hate catch all callbacks too, but restore it for now just in case it breaks something.

src/riak_repl_aae_sink.erl
```diff
@@ -157,12 +163,23 @@ process_msg(?MSG_UPDATE_TREE, IndexN, State=#state{tree_pid=TreePid}) ->
...
+        Response ->
+            send_reply(Response, State)
+    end;
 
 %% no reply
 process_msg(?MSG_COMPLETE, State=#state{owner=Owner}) ->
     riak_repl2_fssink:fullsync_complete(Owner),
+    case riak_kv_entropy_manager:enabled() of
+        false ->
+            riak_kv_index_hashtree:destroy(State#state.tree_pid);
```
engelsanchez, line 186: We destroy the tree upon normal termination, but should we create a cast version of destroy to be used in the terminate method to try to do it when dying unexpectedly?

src/riak_repl_aae_source.erl
```diff
@@ -187,25 +188,42 @@ prepare_exchange(start_exchange, State0=#state{transport=Transport,
...
-                Error ->
-                    lager:info("AAE source failed get_lock for partition ~p, got ~p",
-                               [Partition, Error]),
-                    State#state.owner ! {soft_exit, self(), Error},
-                    {stop, normal, State}
-            end;
-        {error, wrong_node} ->
-            {stop, wrong_node, State0}
-    end;
+case riak_kv_vnode:hashtree_pid(Partition, riak_kv_entropy_manager:enabled()) of
```
engelsanchez: Let's restore the proper indentation to this block.

src/riak_repl_aae_sink.erl
```diff
@@ -108,8 +108,8 @@ handle_info({Error, Socket, Reason},
     lager:info("AAE sink connection to ~p closed unexpectedly: ~p",
                [State#state.clustername, {Socket, Error, Reason}]),
     {stop, {Socket, Error, Reason}, State};
-handle_info(_Info, State) ->
-    {noreply, State}.
+handle_info({hashtree_lock,_}, State) ->
+    send_reply(ok, State).
```
engelsanchez: Same thing with the catch all clause here. Let's add it back.

src/riak_repl_aae_source.erl
```diff
@@ -299,6 +317,12 @@ key_exchange(finish_fullsync, State=#state{owner=Owner}) ->
     lager:debug("AAE fullsync source completed partition ~p",
                 [State#state.index]),
     riak_repl2_fssource:fullsync_complete(Owner),
+    case riak_kv_entropy_manager:enabled() of
```
engelsanchez: Should we destroy the tree also in the case of a canceled fullsync?

PR #657, opened 2015-02-12
core_connection exit abnormally on socket closed.

lordnull:
The connection manager uses a helper process to wrap the call to
riak_core_connection:sync_connect/2. that call will start the gen_fsm that is
the connection, and then block until that gen_fsm exits (using a monitor).
On a normal exit, ‘ok’ is returned. On anything else, an error is returned.

In the case of the socket being closed while the gen_fsm is negotiating the
protocol, the exit cause was ‘normal’. this meant the riak_core_connection
process never called the method to tell a process wanting to connect (such as
a proxy get provider) it has a socket since it’s exited, and the connection
manager never called the method to the process wanting the connection the
connection failed since all it saw was a success.

this fix resolves a failing riak_test replication2_pg:test_pg_proxy

### DIFF
--- a/src/riak_core_connection.erl
+++ b/src/riak_core_connection.erl
@@ -249,7 +249,7 @@ handle_info({_TransTag, Socket, Data}, wait_for_protocol, State = #state{socket 
 
 handle_info({_LikelyClose, Socket}, _StateName, State = #state{socket = Socket}) ->
     lager:debug("Socket closed"),
-    {stop, normal, State}.
+    {stop, socket_closed, State}.
 
 terminate(_Why, _StateName, _State) ->
     ok.


jadeallenx:
Haven't reviewed this at all, but great find.

cmeiklejohn:
:+1: f05d871

:+1: confirmed resolved replication2_pg:test_pg_proxy

_edit sdc: reordered approvals so @borshop picks it up_

cmeiklejohn:
Resolves replication2_upgrade as well:

```
Test Results:
replication2_upgrade-bitcask: pass
replication2_upgrade-bitcask: pass
---------------------------------------------
0 Tests Failed
2 Tests Passed
That's 100.0% for those keeping score
```

cmeiklejohn:
@borshop merge

Issue #637, opened 2014-11-20
"Heartbeat is misconfigured" log by race condition


shino:
ZD ticket: https://basho.zendesk.com/agent/tickets/9508
(Sorry this is writtin in Japanese, not much information except the
first comment, bottom in Web UI.)

Summary of the ticket:

Customer reported the following log when load testing on Riak Enterprise 2.0.2.
The log happended occasionally, almost once in 1 to 10 min.
Load was 500 - 23,000 PUTs/sec

```
2014-11-12 16:48:46.240 [warning] <0.2416.0>@riak_repl2_rtsource_conn:schedule_heartbeat:394 Heartbeat is misconfigured and is not a valid integer.
```

---

It seems that `riak_repl2_rtsource_conn` has race bug.
https://github.com/basho/riak_repl/blob/2.0/src/riak_repl2_rtsource_conn.erl#L345
- At `riak_repl2_rtsource_conn:schedule_heartbeat/1`,
  `hb_interval_tref` is assumed to be `undefined`.
- But there is race condition that it is not true.
- Then the log is output, even heartbeat itself is working correctly.

```
               rtsource_conn/helper  rtsink_conn/helper
                          |                  |
    heartbeat req         |--------+         |
    [set timeout timer]   |        |         |
                          |        |         |
                          |        |         |
    send objects          |--------|-------->|
                          |        |         |
    receive ack           |<-------|---------| ack for objects
    [set interval timer]  |        +-------->|
                          |                  |
    receive heartbeat     |<-----------------|
    ->schedule_heartbeat  |                  |
                          |                  |
                          |                  |
                          |                  |
```

shino:
By adding some sleeps in source and sink, the log sometimes occured.
Diff and settings are described in  https://gist.github.com/shino/9e6be1ece105e1d0070b .

But the step seems to be non-deterministic (not every time, the log happens).
So, please don't take the above steps too seriously :see_no_evil:

Basho-JIRA:
Jon M confirmed code merged to 2.0.6 on 6/25/15.  Dev is done. Now just waiting to go into additional releases.  

**_[posted via JIRA by Patricia Brewer]**_

PR #661, opened 2015-03-16
heartbeat scheduling bug fix

jenafermiller:
In the case the `hb_interval` is already set, the pattern match to get into the first case will not succeed and thus an error message will be thrown.  However, the default error message is incorrect in the case that the heartbeat interval is already set.  We are seeing this issue in [Customer Ticket #10506](https://basho.zendesk.com/agent/tickets/10506) and we believe the root cause is similar to the issue Shino has raised - Ticket #637 (RIAK-1381).  This code change will allow for better logging statements in the situation the heartbeat is already set and we should just leave it alone.


jonmeredith:
@jenagraham is this fixed by #662?

jenafermiller:
@jonmeredith it is fixed!  Thanks!  Closing this one out.

PR #670, opened 2015-04-24
Fix deadlocks in fullsync (aae and keylist) when handoff occurs

JeetKunDoug:
Before this fix, neither source would handle a 'DOWN' message with a normal reason code, which would cause them to deadlock waiting for the vnode to respond if it was handed off to another node.
Tests in riak_repl (will update comments with PR).


JeetKunDoug:
Going to kill this and retarget to 2.1 branch

PR #671, opened 2015-04-24
Fix deadlocks when handoff occurs during fullsync, and function clause mismatch in wait_keylist

JeetKunDoug:
Deadlock fix: Before this fix, neither source would handle a 'DOWN' message with a normal reason code, which would cause them to deadlock waiting for the vnode to respond if it was handed off to another node.
wait_keylist: Was coded as a sync call, but is called async.
Tests in riak_repl. Also requires a fix in riak_kv:

https://github.com/basho/riak_kv/pull/1106
https://github.com/basho/riak_test/pull/789

### DIFF
--- a/src/riak_repl2_fscoordinator.erl
+++ b/src/riak_repl2_fscoordinator.erl
@@ -56,7 +56,6 @@
     socket,
     transport,
     largest_n,
-    owners = [],
     connection_ref,
     partition_queue = queue:new(),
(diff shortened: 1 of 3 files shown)


seancribbs:
:+1: aa2f29c

Ditto on the backport to 2.0 on this one.

JeetKunDoug:
@borshop merge

Issue #654, opened 2015-01-22
Realtime replication for blocks, and other redesign [JIRA: RCS-110]


Basho-JIRA:
Currently, only buckets, users and manifests are replicated in realtime.

It would be good to start discussing on:
- add realtime replication on blocks
- whether do we add switch on replication usage/storage stats
- whether do we add switch on GC bucket

This should involve code change in repl, so linking to riak_repl repo (not CS repo).

**[Created in JIRA by Kota Uenishi]**

Basho-JIRA:
Actually the work has already been merged to 2.1 branch, so it'll be included in Riak 2.1.2. 

**_[posted via JIRA by Kota Uenishi]**_

shino:
cross reference : riak_test in riak_cs repo https://github.com/basho/riak_cs/pull/1146

Basho-JIRA:
Release note: Add configuration items to replicate blocks in realtime replication ( `replicate_cs_blocks_realtime` ) and to replicate block tombstones in realtime replication ( `replicate_cs_block_tombstone` ). The former is to relieve load in fullsync replication (whose price is longer RTQ in realtime replication and risk to drop manifests or other important keys), which is `false` by default. The latter is to relieve load generated by GC in sink cluster of duplicate efforts to reclaim data, which is `true` by default. Both configuration are to be set in `advanced.config` of Riak node, not Riak CS configuration. Note that they are available in Riak 2.1.2 or newer and default behaviour on transferring block tombstones has changed from previous Riak releases ( Riak KV 2.1.2 or older).

c.f. description in https://github.com/basho/riak_repl/pull/678 

**_[posted via JIRA by Kota Uenishi]**_

kuenishi:
Followed by https://github.com/basho/riak_repl/pull/709

PR #678, opened 2015-05-11
Improve CS replication on blocks

kuenishi:
This pull request includes:
- Add eqc tests on skip/send logic combination, which used to be confusing, and to avoid regression
- Refactor existing `send/2` and `send_realtime/2` logic
- Add config knob to switch block tombstone replication (both fs and rt, on by default)
- Add config knob to switch block replication (only rt, off by default)

This does not change replication semantics for CS, but add tombstone of blocks are replicated by default to improve disk space reclaim performance, and deletion performance.

This addressed #654.

### DIFF
--- a/src/riak_repl_cs.erl
+++ b/src/riak_repl_cs.erl
@@ -1,5 +1,5 @@
 %% Copyright (c) 2012-2015 Basho Technologies, Inc.
-%% This repl hook skips Riak CS block buckets
+%% This repl hook skips some objects in Riak CS
 
 %% @doc Handle filters on replicating Riak CS specific data. See also
 %% test/riak_repl_cs_eqc.erl to know which is replicated or not, in
@@ -12,8 +12,8 @@
 %% cross-replicated configuration won't be any problem because they
 %% are just deletion.
 %%
-%% You should not replicate neither in fullsync or realtime as there
-%% are a race condition related to CS garbage collection.
+%% You should not replicate blocks neither in fullsync or realtime as
+%% there are a race condition related to CS garbage collection.
 
 -module(riak_repl_cs).
 
@@ -30,9 +30,14 @@
 -define(STORAGE_BUCKET, <<"moss.storage">>).
 -define(BUCKETS_BUCKET, <<"moss.buckets">>).
 
+-define(CONFIG_REPL_BLOCKS, replicate_cs_blocks_realtime).
+-define(CONFIG_REPL_BLOCK_TOMBSTONE, replicate_cs_block_tombstone).
+-define(CONFIG_REPL_USERS, replicate_cs_user_objects).
+-define(CONFIG_REPL_BUCKETS, replicate_cs_bucket_objects).
+
 %% For fullsync, we don't want to ever replicate tombstones or blocks
 %% or storage or access.  Depending on app.config, we may or may not
-%% want to replicate user and bucket objects. An exceptionis
+%% want to replicate user and bucket objects. An exception is
 %% tombstones of blocks.
 -spec send(riak_object:riak_object(), riak_client:riak_client()) ->
     ok | cancel.
@@ -68,17 +73,17 @@ replicate_object(<<?BLOCK_BUCKET_PREFIX, _Rest/binary>>, IsTombstone, FSorRT) ->
         {false, fullsync} ->
             true;
         {false, realtime} ->
-            app_helper:get_env(riak_repl, replicate_blocks, false);
+            app_helper:get_env(riak_repl, ?CONFIG_REPL_BLOCKS, false);
         {true, _} ->
-            app_helper:get_env(riak_repl, replicate_block_tombstone, true)
+            app_helper:get_env(riak_repl, ?CONFIG_REPL_BLOCK_TOMBSTONE, true)
     end;
 replicate_object(_, true, _) -> false;
 replicate_object(?STORAGE_BUCKET, _, _) -> false;
 replicate_object(?ACCESS_BUCKET, _, _) -> false;
 replicate_object(?USER_BUCKET, _, _) ->
-    app_helper:get_env(riak_repl, replicate_cs_user_objects, true);
+    app_helper:get_env(riak_repl, ?CONFIG_REPL_USERS, true);
 replicate_object(?BUCKETS_BUCKET, _, _) ->
-    app_helper:get_env(riak_repl, replicate_cs_bucket_objects, true);
+    app_helper:get_env(riak_repl, ?CONFIG_REPL_BUCKETS, true);
 replicate_object(_, _, _) -> true.
 
 
@@ -94,10 +99,10 @@ bool_to_ok_or_cancel(false) ->
 -ifdef(TEST).
 
 reset_app_env() ->
-    ok = application:unset_env(riak_repl, replicate_block_tombstone),
-    ok = application:unset_env(riak_repl, replicate_blocks),
-    ok = application:unset_env(riak_repl, replicate_cs_bucket_objects),
-    ok = application:unset_env(riak_repl, replicate_cs_user_objects).
+    ok = application:unset_env(riak_repl, ?CONFIG_REPL_BLOCKS),
+    ok = application:unset_env(riak_repl, ?CONFIG_REPL_BLOCK_TOMBSTONE),
+    ok = application:unset_env(riak_repl, ?CONFIG_REPL_USERS),
+    ok = application:unset_env(riak_repl, ?CONFIG_REPL_BLOCKS).
 
 repl_blocks_test() ->
     reset_app_env(),
--- a/src/riak_repl_cs.erl
+++ b/src/riak_repl_cs.erl
@@ -128,7 +128,7 @@ dont_repl_storage_objects_test() ->
 
 dont_repl_tombstoned_object_test() ->
     reset_app_env(),
-    ok = application:set_env(riak_repl, replicate_block_tombstone, false),
+    ok = application:set_env(riak_repl, replicate_cs_block_tombstone, false),
     %% the riak client isn't even used
     Client = fake_client,
     Bucket = <<"anything">>,
--- a/src/riak_repl_cs.erl
+++ b/src/riak_repl_cs.erl
@@ -48,6 +48,13 @@ send_realtime(Object, _RiakClient) ->
 %%% Internal functions
 %%%===================================================================
 
+replicate_object(<<?BLOCK_BUCKET_PREFIX, _Rest/binary>>, IsTombstone, FSorRT) ->
+    case {IsTombstone, FSorRT} of
+        {false, fullsync} -> true;
+        {false, realtime} -> false;
+        {true, _} ->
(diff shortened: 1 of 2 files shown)


shino:
I want to discuss about the default value of block tombstone propagation.
It's important to reclaim blocks as early as possible. On the other hand,
dropping GC keys or manifests from RTQ may be critical (in some case.)

RTQ has limit for item count, default value is 2,000. If RTQ is filled by block tombstones,
manifests and gc keys (from DELETE Object API calls) may drop. Then, GC key dropping
may leads to orphaned blocks.

kuenishi:
We have a knob to control gc block tombstone generation - `gc_max_workers`. Controlling this with RTQ max size is enough to prevent gc key drops, I think.

What if we dropped GC keys? Maybe we must have a way to observe it's drop and compensate with fullsync?

shino:
> Controlling this with RTQ max size is enough to prevent gc key drops, I think.

I agree it is true, in most cases. One not-so-good point we have now is W=all put at
sink side. So single sloooow (v)node can make RTQ longer.

> What if we dropped GC keys?

One sequence for orphaned blocks:
1. PUT Object on cluster A
2. fullsync replicates the object (manifest and blocks) from cluster A to cluster B
3. DELETE Object API on cluster A
4. No fullsync A->B for a while
5. GC reclaims deleted object at step 3 on cluster A
6. Cluster B have blocks but does not have corresponding GC key

src/riak_repl_cs.erl
```diff
@@ -16,113 +30,57 @@
...
-    true;
-block_bucket(_Bucket) ->
-    false.
+-spec replicate_object(binary(), boolean(), fullsync|realtime) -> boolean().
+replicate_object(<<?BLOCK_BUCKET_PREFIX, _Rest/binary>>, IsTombstone, FSorRT) ->
+    case {IsTombstone, FSorRT} of
+        {false, fullsync} ->
+            true;
+        {false, realtime} ->
+            app_helper:get_env(riak_repl, replicate_blocks, false);
```
shino: Is it better to align with existing prefix `replicate_cs_*` (e.g. `replicate_cs_user_objects`)?

kuenishi: ah, yes, you're right.

src/riak_repl_cs.erl
```diff
@@ -1,6 +1,20 @@
-%% Copyright (c) 2012 Basho Technologies, Inc.
+%% Copyright (c) 2012-2015 Basho Technologies, Inc.
 %% This repl hook skips Riak CS block buckets
```
shino: This line contradicts with descriptions below.

src/riak_repl_cs.erl
```diff
@@ -1,6 +1,20 @@
...
+%% fullsync or realtime.
+%%
+%% For blocks, all tombstones are replicated by default, which is
+%% exception for tombstones. This is to reclaim data space faster.  CS
+%% wants to delete blocks as fast as possible, because keys of blocks
+%% consumes memory space, disk space and disk IO. Deletion conflict in
+%% cross-replicated configuration won't be any problem because they
+%% are just deletion.
+%%
+%% You should not replicate neither in fullsync or realtime as there
```
shino: What shouldn't we replicate?

src/riak_repl_cs.erl
```diff
@@ -16,113 +30,57 @@
 -define(STORAGE_BUCKET, <<"moss.storage">>).
 -define(BUCKETS_BUCKET, <<"moss.buckets">>).
 
-%% For fullsync, we don't want to ever replicate tombstones or blocks or
-%% storage or access.
-%% Depending on app.config, we may or may not want to replicate
-%% user and bucket objects.
+%% For fullsync, we don't want to ever replicate tombstones or blocks
+%% or storage or access.  Depending on app.config, we may or may not
+%% want to replicate user and bucket objects. An exceptionis
```
shino: "exception is"?

src/riak_repl_cs.erl
shino: `replicate_blocks` only affects realtime but other config items like `replicate_block_tombstone` does both rt and fs. I want some additional token that represents realtime for `replicate_blocks`.

kuenishi: Will change to `replicate_cs_blocks_realtime`.

kuenishi:
Updated ^^;

src/riak_repl_cs.erl
```diff
@@ -16,113 +30,62 @@
...
-%% For fullsync, we don't want to ever replicate tombstones or blocks or
-%% storage or access.
-%% Depending on app.config, we may or may not want to replicate
-%% user and bucket objects.
+-define(CONFIG_REPL_BLOCKS, replicate_cs_blocks_realtime).
+-define(CONFIG_REPL_BLOCK_TOMBSTONE, replicate_cs_block_tombstone).
+-define(CONFIG_REPL_USERS, replicate_cs_user_objects).
+-define(CONFIG_REPL_BUCKETS, replicate_cs_bucket_objects).
+
+%% For fullsync, we don't want to ever replicate tombstones or blocks
```
shino: Blocks will be replicated by fs regardless of configuration.

kuenishi: That's old comment ...

src/riak_repl_cs.erl
```diff
@@ -195,71 +141,52 @@ dont_repl_tombstoned_object_realtime_test() ->
...
+
+do_repl_gc_object_test() ->
+    reset_app_env(),
     Client = fake_client,
-    Bucket = <<"moss.buckets">>,
+    Bucket = <<"riak-cs-gc">>,
     Object = riak_object:new(Bucket, <<"key">>, <<"val">>),
-    ?assert(ok_or_cancel_to_bool(send_realtime(Object, Client))).
+    ?assert(ok_or_cancel_to_bool(send_realtime(Object, Client))),
+    ?assert(ok_or_cancel_to_bool(send(Object, Client))).
```
shino: I want a test for tombstoned gc objects are NOT replicated. The property is _very much_ important to avoid block leak but currently the comment [1] only mentions about it (in somewhat different way). Explicit and executable test case helps much. 

[1] https://github.com/basho/riak_cs/blob/develop/src/riak_cs_gc_worker.erl#L251-L257

src/riak_repl_cs.erl
```diff
@@ -135,59 +98,42 @@ bool_to_ok_or_cancel(false) ->
...
-    Object = riak_object:new(Bucket, <<"key">>, <<"val">>),
-    M = dict:from_list([{<<"X-Riak-Deleted">>, true}]),
-    Object2 = riak_object:update_metadata(Object, M),
-    Object3 = riak_object:apply_updates(Object2),
-    ?assertNot(ok_or_cancel_to_bool(send(Object3, Client))).
-
-dont_repl_tombstoned_object_realtime_test() ->
+dont_repl_tombstoned_object_test() ->
+    reset_app_env(),
+    ok = application:set_env(riak_repl, replicate_block_tombstone, false),
```
shino: `replicate_cs_block_tombstone`

src/riak_repl_cs.erl
shino: No need for this. EQC test case covers it.

test/riak_repl_cs_eqc.erl
```diff
@@ -0,0 +1,153 @@
...
+do_send(RTorFS, Object, DecisionTable) ->
+    Type = bucket_name_to_type(riak_object:bucket(Object),
+                               riak_kv_util:is_x_deleted(Object)),
+    {Type, RT, FS} = lists:keyfind(Type, 1, DecisionTable),
+    case RTorFS of
+        rt -> RT;
+        fs -> FS
+    end.
+
+cs_bucket() ->
```
shino: This generates riak buckets.

src/riak_repl_cs.erl
```diff
@@ -16,113 +30,62 @@
...
 recv(_Object) ->
     ok.
 
-%% For realtime, we don't want to ever replicate tombstones or
-%% storage or access.
-%% Depending on app.config, we may or may not want to replicate
-%% user and bucket objects.
+%% For realtime, we don't want to ever replicate tombstones or storage
+%% or access. Depending on app.config, we may or may not want to
+%% replicate user and bucket objects. An exception is tombstones of
```
shino: and blocks

src/riak_repl_cs.erl
```diff
@@ -135,59 +98,42 @@ bool_to_ok_or_cancel(false) ->
...
-    %% the riak client isn't even used
-    Client = fake_client,
-    Bucket = <<"0b:foo">>,
-    Object = riak_object:new(Bucket, <<"key">>, <<"val">>),
-    ?assert(ok_or_cancel_to_bool(send(Object, Client))).
+reset_app_env() ->
+    ok = application:unset_env(riak_repl, ?CONFIG_REPL_BLOCKS),
+    ok = application:unset_env(riak_repl, ?CONFIG_REPL_BLOCK_TOMBSTONE),
+    ok = application:unset_env(riak_repl, ?CONFIG_REPL_USERS),
+    ok = application:unset_env(riak_repl, ?CONFIG_REPL_BLOCKS).
```
shino: duplicated with three lines above. or may be intented to unset `?CONFIG_REPL_BUCKETS`?

shino:
Much better readability by refactoring and clarity by EQC case. Cool PR!

eunit passed, cs's repl_v3_test passed. Please tell me any other requirements/policy before +1 in riak_repl.

There remians non-essential and/or comments for comments. I will run eunit and r_t after they are addressed.

src/riak_repl_cs.erl
kuenishi: I think we should explicitly cover this.

test/riak_repl_cs_eqc.erl
kuenishi: How'bout `cs_raw_bucket()` ? Or `raw_s_bucket()`

src/riak_repl_cs.erl
kuenishi: Yes, this is meant for REPL_BUCKETS

test/riak_repl_cs_eqc.erl
shino: Both seems good. Or it may be good to name it simply `riak_bucket()` because `riak_object()` is used for representing riak object for cs data.

kuenishi:
Updated.

src/riak_repl_cs.erl
shino: This is typo.

test/riak_repl_cs_eqc.erl
shino: Haven't changed?

kuenishi:
Updated (2).

shino:
eunit, xref, dialyzer repl_v3_test (in cs) ran successfully.
:+1: :exclamation: :sake:

kuenishi:
Thanks, man. Will merge manually because of @borshop unavailability.

PR #680, opened 2015-05-18
Changed keep alive to erlang:send_after/3 style. [JIRA: RIAK-1887]

### LINKED ISSUE #679: The proxy get riak_repl2_pg_block_provider.erl timer handling looks too odd [JIRA: RIAK-1807]
See the end of this commentary for suggestions for changing the timer handling code in riak_repl2_pg_block_provider.erl

---

From: https://basho.zendesk.com/agent/tickets/10882

Wow, something is really wrong with that https://basho.zendesk.com/agent/#/tickets/10793 cluster dump. That is 8.3 million 'keepalive' messages. If I used some Perl to replace all of the 'keepalive,' strings to make it easier to read, then I see only this:

Message queue: [KEEPALIVE-REPEATS,{tcp,#Port&lt;0.340692>,&lt;&lt;174 bytes>>},KEEPALIVE-REPEATS,keepalive]
So, there's only one non-keepalive message out of over 8.3 million.

If the timer service sends one message/second, then that is over 96 days of keepalive messages. If the timer process was correct and sends only 1 msg/sec, and if the process was over 96 days old, then the gen_server was stuck for over 96 days and there was only 1 PDU of TCP data in the same 96 days. That scenario seems .... very unlikely?

If the OS system time went backward for 96 days, then perhaps the timer service had a "catch up" of sending 8M messages very quickly? I don't know, that sounds .... very unlikely also?

HRM, this code looks weird, it should be using erlang:send_interval() or erlang:start_timer() or perhaps timer:send_after() as a one-time thing, rather than as a repeated timer.

https://github.com/basho/riak_repl/blob/develop/src/riak_repl2_pg_block_provider.erl#L163-L164

The handle_info() callbacks cancels the repeated timer and then re-created here:

https://github.com/basho/riak_repl/blob/develop/src/riak_repl2_pg_block_provider.erl#L131-L134

If keepalive_timer() is changed to use a 1-time timer, then there are some places where new calls to keepalive_timer() must be added, e.g. https://github.com/basho/riak_repl/blob/develop/src/riak_repl2_pg_block_provider.erl#L109-L112

lordnull:
This resolves an issue where the provider process could become blocked or
stuck for an extended period of time, only to receive many pointless
messages from the timer system.

First change is to up the keepalive from 1 second to 60. This should be
enough for the remote end to realize if the socket connection has died, but
not so much as to be spammy.

The second, more invasive change, is to use the erlang:send_after/3 reset
pattern rather than timer:send_interval/3. By using the erlang:send_after,
messages will not be sent if the process becomes stuck, as no code capable
of resetting the timer will run. Secondly, the return values of the
erlang:send_after and erlang:cancel_timer don't need to be pattern matched
to ensure correctness (since the former returns the reference itself, and
the latter returns an integer or false).

Resolves #679 (RIAK-1807) .

### DIFF
--- a/src/riak_repl2_pg_block_provider.erl
+++ b/src/riak_repl2_pg_block_provider.erl
@@ -109,7 +109,7 @@ handle_cast(_Msg, State) ->
 handle_info(keepalive, State=#state{socket=Socket, transport=Transport}) ->
     Data = term_to_binary(stay_awake),
     Transport:send(Socket, Data),
-    State2 = keepalive_timer(State),
+    State2 = keepalive_timer(State#state{keepalive_timer = undefined}),
     {noreply, State2};
 handle_info({tcp_closed, Socket}, State=#state{socket=Socket}) ->
     lager:info("Connection for proxy_get ~p closed", [State#state.other_cluster]),
@@ -130,9 +130,7 @@ handle_info({Proto, Socket, Data},
         when Proto==tcp; Proto==ssl ->
     Transport:setopts(Socket, [{active, once}]),
     Msg = binary_to_term(Data),
-    %% restart the timer after each message has been processed
-    State = keepalive_timer(State0),
-    handle_msg(Msg, State);
+    handle_msg(Msg, State0);
 handle_info(_Msg, State) ->
     {noreply, State}.
 
@@ -161,7 +159,7 @@ code_change(_OldVsn, State, _Extra) ->
     {ok, State}.
 
 keepalive_timer(State = #state{keepalive_timer = undefined}) ->
-    NewRef = erlang:send_after(?KEEPALIVE, keepalive, self()),
+    NewRef = erlang:send_after(?KEEPALIVE, self(), keepalive),
     State#state{keepalive_timer = NewRef};
 
 keepalive_timer(State) ->
--- a/src/riak_repl2_pg_block_provider.erl
+++ b/src/riak_repl2_pg_block_provider.erl
@@ -16,7 +16,7 @@
     terminate/2, code_change/3, status/1, status/2]).
 
 %% send a message every KEEPALIVE milliseconds to make sure the service is running on the sink
--define(KEEPALIVE, 1000).
+-define(KEEPALIVE, 60000).
 
 -record(state,
         {
@@ -75,16 +75,16 @@ handle_call({connected, Socket, Transport, _Endpoint, _Proto, Props}, _From,
     lager:debug("Keeping stats for " ++ SocketTag),
     riak_core_tcp_mon:monitor(Socket, {?TCP_MON_PROXYGET_APP, source,
                                        SocketTag}, Transport),
-    TRef = keepalive_timer(),
     Transport:setopts(Socket, [{active, once}]),
     {ok, Client} = riak:local_client(),
-    {reply, ok, State#state{
+    State2 = State#state{
                   transport=Transport,
                   socket=Socket,
                   other_cluster=Cluster,
-                  client=Client,
-                  keepalive_timer=TRef
-                 }};
+                  client=Client
+    },
+    State3 = keepalive_timer(State2),
+    {reply, ok, State3};
 handle_call(status, _From, State=#state{socket=Socket,
                                         proxy_gets_provided=PGCount}) ->
     SocketStats = riak_core_tcp_mon:socket_status(Socket),
@@ -109,7 +109,8 @@ handle_cast(_Msg, State) ->
 handle_info(keepalive, State=#state{socket=Socket, transport=Transport}) ->
     Data = term_to_binary(stay_awake),
     Transport:send(Socket, Data),
-    {noreply, State};
+    State2 = keepalive_timer(State),
+    {noreply, State2};
 handle_info({tcp_closed, Socket}, State=#state{socket=Socket}) ->
     lager:info("Connection for proxy_get ~p closed", [State#state.other_cluster]),
     {stop, socket_closed, State};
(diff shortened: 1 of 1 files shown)


src/riak_repl2_pg_block_provider.erl
```diff
@@ -109,7 +109,8 @@ handle_cast(_Msg, State) ->
 handle_info(keepalive, State=#state{socket=Socket, transport=Transport}) ->
     Data = term_to_binary(stay_awake),
     Transport:send(Socket, Data),
-    {noreply, State};
+    State2 = keepalive_timer(State),
```
jonmeredith: Could modify state to set timer ref undefined here and avoid the cancel_timer.

src/riak_repl2_pg_block_provider.erl
```diff
@@ -160,6 +160,11 @@ terminate(_Reason, #state{}) ->
 code_change(_OldVsn, State, _Extra) ->
     {ok, State}.
 
-keepalive_timer() ->
-    timer:send_interval(?KEEPALIVE, keepalive).
+keepalive_timer(State = #state{keepalive_timer = undefined}) ->
+    NewRef = erlang:send_after(?KEEPALIVE, keepalive, self()),
```
jonmeredith: Does this work?  Manpage says

```
 erlang:send_after(Time, Dest, Msg) -> TimerRef
```

jonmeredith: Doesn't error as as Dest can be a registered named.

```
Eshell V5.10.3  (abort with ^G)
1> erlang:send_after(60000, keepalive, self()).    
#Ref<0.0.0.28>
2> erlang:send_after(60000, self(), keepalive).
#Ref<0.0.0.33>
```

jonmeredith:
-1 timer interval not set correctly.

jonmeredith:
So, thinking about it, I'm not even sure if we should cancel and restart the timer on receiving TCP traffic.

Divining history based on other support incidents we've had, I think this timeout was added to handle the case of remote node death.  As the block provider connects _to_ the block requester and just listens for requests, if the requestor side dies, the provider does notice until the TCP keepalive mechanism discovers that (2 hours, 2 days, who knows, kernel/tuning specific), so this periodically generates traffic from provider to requestor to trigger TCP timeout logic.

I think it's safe just to periodically send the stay_awake message every 60s or so.

src/riak_repl2_pg_block_provider.erl
```diff
@@ -109,7 +109,8 @@ handle_cast(_Msg, State) ->
 handle_info(keepalive, State=#state{socket=Socket, transport=Transport}) ->
     Data = term_to_binary(stay_awake),
     Transport:send(Socket, Data),
-    {noreply, State};
+    State2 = keepalive_timer(State#state{keepalive_timer = undefined}),
```
jonmeredith: Can clean this up - no longer cancel timers, so no need to do undefined distinction.

src/riak_repl2_pg_block_provider.erl
```diff
@@ -160,6 +158,11 @@ terminate(_Reason, #state{}) ->
 code_change(_OldVsn, State, _Extra) ->
     {ok, State}.
 
-keepalive_timer() ->
-    timer:send_interval(?KEEPALIVE, keepalive).
+keepalive_timer(State = #state{keepalive_timer = undefined}) ->
+    NewRef = erlang:send_after(?KEEPALIVE, self(), keepalive),
+    State#state{keepalive_timer = NewRef};
 
+keepalive_timer(State) ->
```
jonmeredith: This path is never hit.

src/riak_repl2_pg_block_provider.erl
```diff
@@ -160,6 +158,11 @@ terminate(_Reason, #state{}) ->
 code_change(_OldVsn, State, _Extra) ->
     {ok, State}.
 
-keepalive_timer() ->
-    timer:send_interval(?KEEPALIVE, keepalive).
+keepalive_timer(State = #state{keepalive_timer = undefined}) ->
```
jonmeredith: No longer need pattern match.

jonmeredith:
Confirmed existence of problem on existing code - new interval timer added for every served proxy get request

Set up two single node clusters c1 -> c2, enabled and started pg (provider on c1, requester c2)

```
{_,Pid,_,_}] = supervisor:which_children(riak_repl2_pg_block_requester_sup).
[riak_repl2_pg_block_requester:proxy_get(Pid, <<"b">>,<<"k">>,[]) || X <- lists:seq(1,1000)].
```

Verified timer increase with `ets:i(timer_interval_tab)`.

Verified keepalive message being reset with redbug

```
(dev1@127.0.0.1)27> redbug:start("riak_repl2_pg_block_provider:handle_info", [{msgs, 600},{time,300000}]).
ok

09:58:29 <{riak_repl2_pg_block_provider,init,1}> {riak_repl2_pg_block_provider,
                                                  handle_info,
                                                  [keepalive,
                                                   {state,ranch_tcp,
                                                    #Port<0.28119>,undefined,
                                                    "c2",#Ref<0.0.0.183149>,
                                                    undefined,
                                                    {riak_client,
                                                     'dev1@127.0.0.1',
                                                     undefined},
                                                    #Ref<0.0.0.207114>,3004}]}
09:59:29 <{riak_repl2_pg_block_provider,init,1}> {riak_repl2_pg_block_provider,
                                                  handle_info,
                                                  [keepalive,
                                                   {state,ranch_tcp,
                                                    #Port<0.28119>,undefined,
                                                    "c2",#Ref<0.0.0.183149>,
                                                    undefined,
                                                    {riak_client,
                                                     'dev1@127.0.0.1',
                                                     undefined},
                                                    #Ref<0.0.0.211602>,3004}]}
```

Will retest then approve after killing the dead code.

Basho-JIRA:
Jon M confirmed code merged to 2.0.6 on 6/25/15.  Dev is done. Now just waiting to go into additional releases.  

**_[posted via JIRA by Patricia Brewer]**_

PR #665, opened 2015-04-02
Update AAE fullsync to support bucket types

ian-mi:
Fixes issue #663 (RIAK-1292) (RIAK-1294) (RIAK-1295) by updating `encode_obj_msg` to support the `w2` wire protocol, and by changing the AAE sink to use an updated `decode_obj_msg` function which can now properly handle bucket type property hashes.

`encode_obj_msg` and `decode_obj_msg` are not currently proper inverses of each other, but I kept the current behavior as not all messages passed to `decode_obj_msg` are guaranteed to be encoded using `encode_obj_msg`.

I have manually verified that this allows keys with a bucket type to be synced using AAE fullsync, but a test combining bucket types and AAE fullsync should probably be added to `riak_test`.

Issue #663, opened 2015-03-30
AAE Fullsync Incompatible with Bucket Types [JIRA: RIAK-1655]


ian-mi:
When using AAE fullsync, the fullsync source will crash when attempting to send a key with a bucket type with an error such as

```
2015-03-30 13:47:10.371 [error] <0.2216.0> Supervisor riak_repl2_fssource_sup had child 433883298582611803841718934712646521460354973696 started with {riak_repl2_fssou
rce,start_link,undefined} at <0.2163.1> exit with reason {{function_clause,[{riak_repl_util,new_w1,[{<<"type">>,<<"test">>},<<"test">>,<<53,1,0,0,0,38,131,108,0,0,0,1,
104,2,109,0,0,0,12,197,82,177,11,248,33,144,157,0,0,78,34,104,2,97,1,110,5,0,16,49,142,206,14,106,0,0,0,1,0,0,0,6,1,118,97,108,117,101,0,0,0,174,0,0,5,147,0,11,106,80,
0,9,163,133,22,55,75,103,66,118,54,102,66,121,102,103,116,104,54,84,80,68,98,56,77,79,70,0,0,0,0,12,1,88,45,82,105,97,107,45,77,101,116,97,0,0,0,3,0,131,106,0,0,0,6,1,
105,110,100,101,120,0,0,0,3,0,131,106,0,0,0,13,1,...>>],...},...]},...} in context child_terminated
```

It appears that this is happening because fullsync has not been updated to use the new replication wire protocol.

ian-mi:
This issue does not occur with keylist fullsync, because `riak_repl_util:encode_obj_msg` is called with the `w0` wire protocol version. In AAE fullsync it is being called with `w2` which is not being handled correctly.

Basho-JIRA:
There is a fix for this via [~bsparrow435]. We should schedule it for 2.0.6 and 2.1.1. 

**_[posted via JIRA by Sean Cribbs]**_

Basho-JIRA:
https://github.com/basho/riak_repl/pull/692 reverts the original fix - if we decide to move forward with it for 2.1.2, it's here for reference. 

**_[posted via JIRA by Douglas Rohrer]**_

Basho-JIRA:
Also, duplicate of riak-1987 

**_[posted via JIRA by Douglas Rohrer]**_

Basho-JIRA:
Ian is working on the binary vs term object as it dances through the realtime processes.

[...]

PR #682, opened 2015-05-20
Update AAE fullsync to support bucket types V2

ian-mi:
This reverts commit 078f16315c494ea06bd1c1b5f79aece43b09ff74. See pull request #665.

### DIFF
--- a/src/riak_repl2_fssink.erl
+++ b/src/riak_repl2_fssink.erl
@@ -148,7 +148,7 @@ handle_info({Proto, Socket, Data},
         State=#state{socket=Socket,transport=Transport}) when Proto==tcp; Proto==ssl ->
     %% aae strategy will not receive messages here
     Transport:setopts(Socket, [{active, once}]),
-    case decode_obj_msg(Data) of
+    case riak_repl_util:decode_obj_msg(Data) of
         {fs_diff_obj, RObj} ->
             riak_repl_util:do_repl_put(RObj);
         Other ->
@@ -184,18 +184,6 @@ handle_info(init_ack, State=#state{socket=Socket,
     end;
 handle_info(_Msg, State) ->
     {noreply, State}.
-
-decode_obj_msg(Data) ->
-    Msg = binary_to_term(Data),
-    case Msg of
-        {fs_diff_obj, BObj} when is_binary(BObj) ->
-            RObj = riak_repl_util:from_wire(BObj),
-            {fs_diff_obj, RObj};
-        {fs_diff_obj, _RObj} ->
-            Msg;
-        Other ->
-            Other
-    end.
 
 terminate(_Reason, #state{fullsync_worker=FSW, work_dir=WorkDir, strategy=Strategy}) ->
     %% TODO: define a fullsync worker behavior and call it's stop function
--- a/src/riak_repl2_fssink.erl
+++ b/src/riak_repl2_fssink.erl
@@ -148,8 +148,9 @@ handle_info({Proto, Socket, Data},
         State=#state{socket=Socket,transport=Transport}) when Proto==tcp; Proto==ssl ->
     %% aae strategy will not receive messages here
     Transport:setopts(Socket, [{active, once}]),
-    case riak_repl_util:decode_obj_msg(Data) of
-        {fs_diff_obj, RObj} ->
+    case binary_to_term(Data) of
+        {fs_diff_obj, BinObj} ->
+            RObj = riak_repl_util:decode_bin_obj(BinObj),
             riak_repl_util:do_repl_put(RObj);
         Other ->
             gen_fsm:send_event(State#state.fullsync_worker, Other)
--- a/src/riak_repl_fullsync_worker.erl
+++ b/src/riak_repl_fullsync_worker.erl
@@ -140,7 +140,7 @@ handle_cast({put, RObj, Pool}, State) ->
     poolboy:checkin(Pool, self()),
     {noreply, State};
 handle_cast({bin_put, BinObj, Pool}, State) ->
-    RObj = riak_repl_util:from_wire(BinObj),
+    RObj = riak_repl_util:decode_bin_obj(BinObj),
     riak_repl_util:do_repl_put(RObj),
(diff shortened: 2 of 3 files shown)


cmeiklejohn:
:+1:

Seems straightforward.

One question: does this assume the bucket type exists on the remote side and has the same configuration?  It seems to, unless I'm missing something obvious...

ian-mi:
No, the put path uses riak_repl_util:do_repl_put which checks that the bucket type exists and matches the remote hash. This was why the code needed to be moved from using bin_put to put, as bin_put assumes that its input is an object while put just passes its input to do_repl_put. I believe Sparrow's test also checks that mismatched types are rejected.

bsparrow435:
It does indeed, thanks for the review @cmeiklejohn

cmeiklejohn:
Is this not a bors-enabled repo?

cmeiklejohn:
:+1: 00040f9

jonmeredith:
bors out all the way... http://bors.bos1:8000/bors.html?repo=riak_repl

On Fri, May 22, 2015 at 11:57 AM, Christopher S. Meiklejohn <
notifications@github.com> wrote:

> Is this not a bors-enabled repo?
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/basho/riak_repl/pull/682#issuecomment-104729415.

## 

Jon Meredith
Chief Architect
Basho Technologies, Inc.
jmeredith@basho.com

cmeiklejohn:
I realized I forgot to approve the commit in the right way.

cmeiklejohn:
This will merge this change into develop; what other branches do we want to target with this patch?

jonmeredith:
I would argue for 2.0 branch and 2.1 branch as it's a significant bug for
adoption of 2.0 features.

On Fri, May 22, 2015 at 12:15 PM, Christopher S. Meiklejohn <
notifications@github.com> wrote:

> This will merge this change into develop; what other branches do we want
> to target with this patch?
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/basho/riak_repl/pull/682#issuecomment-104735375.

## 

Jon Meredith
Chief Architect
Basho Technologies, Inc.
jmeredith@basho.com

cmeiklejohn:
Actually, @ian-mi, before we merge, can you modify the commit message to be a bit more descriptive?

jonmeredith:
Is this PR complete?  A quick scan of the code didn't show me where the BTHash is being generated for send_missing in riak_repl_aae_source (did not check keylisting).

What is the impact of running this on a cluster with mixed versions on both sides?  Unclear to me if the behavior has been defined.

Given original issue #663 (RIAK-1292) (RIAK-1294) (RIAK-1295) (RIAK-1736), it seems like we cannot send objects with bucket types over fullsync as that will crash the receiver and we will make no progress at all.

Options:
1) Increment the protocol version (need to investigate if it's fullsync, or the individual AAE/keylist methods) to determine if the BT object can be sent.  If it can, great. If not, do not send the object so that default objects can be sent.  LOUDLY log a count of the objects not sent and recommend an upgrade to a version that fixes the issue.

2) Continue to let it crash and make that be the forcing function for upgrades.  It's broken already, so this does fix it.

Thoughts @ian-mi @bsparrow435 @cmeiklejohn ?

ian-mi:
Keylist fullsync should already work with bucket types. The BTHash was already being generated in enocde_obj_msg, which keylist fullsync uses. This PR adds support to encode_obj_msg for the w2 wire format which AAE fullsync uses (but keylist does not), and adds support for the BTHash to the receiver by calling decode_obj_msg and using do_repl_put instead of bin_put.

lordnull:
+1 00040f9d61c3d9371ea2f0c95a4778f8e0b3ef41

As expected, this passes the [test meant for it](https://github.com/basho/riak_test/pull/797). There was mention of concerns about speed/efficiency. While there is a speed penalty here, after running the repl_aae_fullysnc_benchmark test multiple times both with and without this patch, I have two posibilities. I either do not have a reliable benchmark setup, or any slow down introduced by this patch is small enough that it is within the speed variance of a normal run (e.g., no detectible change).

For posterity, the results in seconds:

| with? | run 1 | run 2 | run 3 | avg |
| --- | --- | --- | --- | --- |
| Without | 50.75 | 40.70 | 34.68 | 42.04 |
| With | 46.74 | 47.74 | 30.65 | 41.71 |

JeetKunDoug:
@ian-mi at some point you suggested you could update this (or at least investigate) doing the decoding in worker process rather than where it is now - see https://github.com/basho/riak_repl/pull/665#issuecomment-120999465 for context if necessary (as I had to). If possible, we'd like to get that done before we merge this. If you don't have time right now, please let @jonmeredith know and we'll try to find someone else to work on it.

cmeiklejohn:
Are we really considering merging this?  I don't believe the placement of the decoding is the problematic part of this code -- it's the general approach and not considering all of the bucket properties that should be considered for replication.

I've expressed my concern with merging this, as well as the associated bucket type fixes without performing the proper engineering work to support this correctly.

ian-mi:
First draft of rework to decode objects in the fullsync worker instead of AAE sink process has been pushed. Should reduce some overhead. Will finish and test on Monday.

ian-mi:
Confirmed that the new version passes Sparrow's new AAE fullsync / bucket types test as well as the existing fullsync tests. New version should be ready for review / merge.

bsparrow435:
`repl_bucket_types` passes with code, fails without. This test also confirms keylist strategy and mixed versions. :+1:

jonmeredith:
+1 5f28530

Reviewed changes.
Manually recreated bug and confirmed fixed by just upgrading the sink side.
Checked realtime replication and keylist fullsync work with change applied.

Will let bors test the merged code, then decide whether to merge against 2.0 and merge forward or merge into develop and backport.

We still need a permanent solution to the bucket properties used for creating the hash, but will resolve in a different PR.

jonmeredith:
Replaced by #706 (RIAK-1602) (RIAK-1802)

PR #686, opened 2015-05-27
Ensure that we hash write_once bucket.

cmeiklejohn:
Ensure that when computing a hash for repl on the bucket type
properties, that we also consider write_once buckets as well.


cmeiklejohn:
@borshop: retry

cmeiklejohn:
Reopening.

PR #687, opened 2015-05-27
Update bucket properties hash for write_once.

cmeiklejohn:
Update the bucket properties hash to take the write_once bucket property
into consideration; additionally, remove the allow_mult and
last_write_wins restrictions.

### DIFF
--- a/src/riak_repl_bucket_type_util.erl
+++ b/src/riak_repl_bucket_type_util.erl
@@ -14,7 +14,7 @@
 
 -define(DEFAULT_BUCKET_TYPE, <<"default">>).
 
--define(DEFAULT_HASH_BUCKET_PROPS, [consistent, datatype, n_val, allow_mult, last_write_wins]).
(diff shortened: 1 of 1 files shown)


cmeiklejohn:
If this looks good; will open PRs for 2.0.6 and 2.1.2.

jonmeredith:
+1 37ad44d

cmeiklejohn:
@borshop merge

jonmeredith:
Thinking a bit harder about this, I hadn't realized that the `riak_repl_bucket_type_util:property_hash` function was already around for realtime.

If we change the way the hash is generated, this is going to break realtime replication between mixed version clusters.

Lazy question - do we have any r_t's that validate matching configs work and bad configs are rejected during rolling upgrades?

cmeiklejohn:
I'm reverting this commit and will re-open a version that introduces capabilities this week.

Additionally, to answer your question: no, we don't have those tests.

PR #691, opened 2015-05-28
Add bucket type support for AAE fullsync.

cmeiklejohn:
Add bucket type support for AAE fullsync; ensure that the message is
properly decoded into the two tuple, and use this to ensure that the
bucket properties hash matches.

This reverts commit 078f16315c494ea06bd1c1b5f79aece43b09ff74.


ian-mi:
Already merged to 2.1

PR #692, opened 2015-05-28
Add bucket type support for AAE fullsync.

cmeiklejohn:
Add bucket type support for AAE fullsync; ensure that the message is
properly decoded into the two tuple, and use this to ensure that the
bucket properties hash matches.

This reverts commit 078f16315c494ea06bd1c1b5f79aece43b09ff74.

### DIFF
--- a/src/riak_repl2_fssink.erl
+++ b/src/riak_repl2_fssink.erl
@@ -148,7 +148,7 @@ handle_info({Proto, Socket, Data},
         State=#state{socket=Socket,transport=Transport}) when Proto==tcp; Proto==ssl ->
     %% aae strategy will not receive messages here
     Transport:setopts(Socket, [{active, once}]),
-    case decode_obj_msg(Data) of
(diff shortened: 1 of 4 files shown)


jonmeredith:
(Repeat of comment on #665)

Although this is reverted and we're looking for a more general solution to replicated bucket properties, I wanted to add a note here in case we use something like this in the interim.

I can definitely see why it's logically sound to put the decoding stuff into riak_repl_util, but that means that the objects will travel between the aae_sink and the worker process as decoded objects rather than as binaries and I'm concerned this will cause a minor performance regressions - generating more garbage on the aae_sink process and increased term copying.

Basho-JIRA:
Per Doug's request, marked this as a duplicate of 1736 

**_[posted via JIRA by Patricia Brewer]**_

Basho-JIRA:
Closing this ticket since it is a dup of 1736.  Ticket has been linked. Doug R approved.  

**_[posted via JIRA by Patricia Brewer]**_

PR #696, opened 2015-06-01
Add capability for default hash.

cmeiklejohn:
Use a capability to determine how to derive the bucket type hash so the
cluster can support mixed versions.


bsparrow435:
Tested manually, confirmed it works. Thanks.

PR #728, opened 2016-02-29
Fix riak_repl2_rtq_eqc test cleanup

nickelization:
Leaving the rtq_max_bytes setting as-is could sometimes clobber other
tests that didn't expect random values to be present in that setting.
(Specifically the riak_repl2_rtsource_conn test, in the example we found
on the builder.)

Unsetting this value on cleanup fixes the problem.

### DIFF
--- a/test/riak_repl2_rtq_eqc.erl
+++ b/test/riak_repl2_rtq_eqc.erl
@@ -67,6 +67,7 @@ setup() ->
 
 cleanup(_) ->
     kill_and_wait(riak_repl2_rtq),
+    application:unset_env(riak_repl, rtq_max_bytes),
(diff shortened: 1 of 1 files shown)


JeetKunDoug:
:+1: 573bfa6

JeetKunDoug:
@borshop: retry

JeetKunDoug:
Re-kicked the build manually on Buildbot, but it failed to properly update status here - so asking @borshop nicely to retry.

nickelization:
After discussing with @JeetKunDoug and @jonmeredith, decided to manually merge this fix. It sucks to have to bypass bors, but the fix is just for a test, and so the fact that other tests are still broken shouldn't stop us from at least partially improving the situation.

PR #738, opened 2016-04-04
Timeseries support

macintux:
Introduces a new wire protocol and support for timeseries data.

Differences between traditional KV values and timeseries that impact this PR:
- Timeseries uses the `w1c` put path which currently lacks postcommit support.
- Timeseries data uses two different keys: a local key which is a hash of column values, and a partition key which is a hash of those same columns but with the time value mapped to a quantum.
- Timeseries data can be uploaded in batches which will in turn be converted into 1 or more batches, each targeted at a single partition. Those batches can be dropped into the realtime queue with the partition key for remote distribution.

Without this PR, there is no realtime sync, and fullsync is crippled because the objects will be pushed into the standard KV put path at the remote cluster and will not be delivered to the proper location (and thus cannot be found).

Among the problems not solved by this PR:
- As has historically been the case, there is very little sharing of metadata between clusters, so the timeseries bucket type must be constructed on each cluster independently.
- If the sink cluster does not support timeseries data at all, bad things will happen. This is a problem to be investigated during the next phase, the great merge of TS and KV.

Requires corresponding changes to `riak_kv` (PR will be linked) and there is a new test in `riak_test` (ditto).


macintux:
This now depends on the inclusion of https://github.com/basho/riak_ql/pull/122 by @gordonguthrie for DDL comparisons on the sink cluster.

Updated riak_test test to reflect the DDL comparison requirement.

All PRs ready for review.

src/riak_repl2_rtsink_conn.erl
```diff
@@ -291,12 +294,19 @@ maybe_push(Binary, Meta) ->
...
-          riak_repl2_rtq:push(length(List), Binary, Meta2)
+            lager:debug("app env either set to always, or in default; doing cascade"),
+            Meta2 = orddict:erase(skip_count, Meta),
+            rtq_push(Binary, riak_repl_util:from_wire(Binary), Meta2)
     end.
 
+%% KV data is a list of to_wire objects
+rtq_push(Binary, Items, Meta) when is_list(Items) ->
+    riak_repl2_rtq:push(length(Items), Binary, Meta);
+%% Timeseries data is an encoded tagged tuple
```
nickelization, line 305: Minor detail, but wondering if this comment is now redundant enough that we should just remove it. It's only a tagged tuple now, and no longer encoded like in the first draft implementation; plus it seems obvious enough what it is now from looking at the function header since the tag is `ts`.

hazen:
Testing it not looking very good: https://gist.github.com/javajolt/e456843bfac11f15eaf1e2dada4364f2

I'm rebuilding the whole thing and will rerun over night.  Hopefully I just screwed something up.

macintux:
Ugh. Will look.

macintux:
Some of the failures appear to be related to the reformatting of messages to include a tagged ‘ok’ tuple. Making certain my riak_test branch is merged with riak_ts-develop should have handled those.

Some are because of a typo “aggregration” which exists in the tests and I failed to clean up when I restructured ‘ts_util:get_ddl’.

I do see some replication failures that aren’t so simple to explain, so digging into those. Probably some real bugs lurking there.

-John

hazen:
Looks like deletes are having issues:

```
{{assertEqual_failed,[{module,ts_simple_single_key_ops},
                      {line,44},
                      {expression,"riakc_ts : delete ( Pid , << \"table1\" >> , [ 1 , 2 , 4 ] , [ ] )"},
                      {expected,ok},
                      {value,{error,{1009,
                                     <<"Failed to delete record: riak_kv_w1c_server_crashed">>}}}]},
```

and

```
ts_simple_http_api_SUITE:delete_data_existing_row_test failed on line 236
Reason: {badmatch,{ok,"500",
    [{"Vary","Accept-Encoding"},
     {...}
```

macintux:
replication_object_reformat failed locally until I bumped the max timeout for retries; transfers take forever. I’m surprised we don’t bump up the max concurrency value in all test clusters.

-John

macintux:
I think I completely missed the relationship between deletes and realtime replication. Thanks, will chase this down.

-John

macintux:
Deletes are fixed in the corresponding KV branch. Will look into a test for replication + deletes.

src/riak_repl_cs.erl
```diff
@@ -16,113 +31,68 @@
...
--spec skip_buckets_bucket(binary()) -> boolean().
-skip_buckets_bucket(Bucket) ->
-    ReplicateBuckets = app_helper:get_env(riak_repl, replicate_cs_bucket_objects,
-                       true),
-    handle_should_replicate_buckets_bucket(ReplicateBuckets, Bucket).
+%% @doc decides replicate or not, according to the combination of
+%% bucket name, whether tombstone or not, and fullsync or realtime.
+-spec replicate_object(binary(), boolean(), fullsync|realtime) -> boolean().
+replicate_object(<<?BLOCK_BUCKET_PREFIX, _Rest/binary>>, IsTombstone, FSorRT) ->
+    case {IsTombstone, FSorRT} of
```
JeetKunDoug, line 68: So why does this one function head have a nested case? Would be easy to break down into a series of function heads and be more consistent with the rest of the replicate_object implementation.

macintux, line 68: This is part of the `develop-2.2` merge.

JeetKunDoug, line 68: Awesome - so I probably missed it when it came in from CS - should get it cleaned up but can be done post-1.3

src/riak_repl2_ts.erl
```diff
@@ -0,0 +1,71 @@
...
+postcommit(PartitionBatch, Bucket, BucketProps) ->
+    maybe_postcommit(PartitionBatch, Bucket, proplists:get_value(repl, BucketProps, both)).
+
+%% If `repl' is `false' or `fullsync', we skip realtime
+maybe_postcommit(_PartitionBatch, _Bucket, false) ->
+    ok;
+maybe_postcommit(_PartitionBatch, _Bucket, fullsync) ->
+    ok;
+maybe_postcommit({PartIdx, Vals}=PartitionBatch, Bucket, _ReplProp) ->
+    lager:debug("Timeseries batch sent to repl~n    PartIdx~p => ~p...", [PartIdx, hd(Vals)]),
```
JeetKunDoug: Is this providing any useful information at this point? Seems like we should remove the lager call here.

macintux: Given how hard it is to troubleshoot MDC, and the low cost of lager debug statements when the loglevel is at normal levels, I'd rather leave it, but open to discussion.

src/riak_repl2_ts.erl
```diff
@@ -0,0 +1,71 @@
...
+
+get_identity_hashes(Table) ->
+    Mod = riak_ql_ddl:make_module_name(Table),
+    case catch Mod:get_identity_hashes() of
+        {_, {undef, _}} ->
+            %% Module doesn't know about identity hashes, doesn't
+            %% exist, etc
+            lager:warning("DDL module for ~p does not exist or is not "
+                          "compiled with identity hashes", [Table]),
+            [-1];
```
JeetKunDoug: So why send the data across the RTQ instead of just dropping it here? It appears we simply filter out the [-1] hashes on the other end of the Repl connection in the sink, so why not either crash or just drop them here?

JeetKunDoug: Ok, so crashing would probably be bad since this is called from the w1c worker, but dropping them rather than forwarding on to the RTQ seems reasonable.

src/riak_repl_bucket_type_util.erl
```diff
@@ -21,13 +21,37 @@ bucket_props_match(Props) ->
...
+            RemoteHash = prop_get(?BT_META_PROPS_HASH, undefined, Props),
+            Extra = prop_get(?BT_META_EXTRA_VALIDATION, [], Props),
+            compare_bucket_types(Type, RemoteHash, Extra);
         false ->
             %% This is not a typed bucket. Check if the remote
             %% side is also untyped.
             undefined =:= prop_get(?BT_META_PROPS_HASH, undefined, Props)
     end.
 
+compare_bucket_types(Type, -1, Extra) ->
```
JeetKunDoug: Why "-1" as a value here? the atom `ts` is, in term_to_binary, the exact same length as the integer `-1` and would be much easier to understand what was going on. Realize this causes changes in a few places, but I'd much prefer `ts` as an atom unless there was some reason -1 was chosen I don't get.

macintux: It’s more general than timeseries (it’s for any use case where we need to do bucket type comparison that the current overly-simplistic hash mechanism can’t handle), and it’s supposed to be an invalid integer hash value. I could definitely see defining a macro to abstract the -1.

-John

src/riak_repl_bucket_type_util.erl
```diff
@@ -21,13 +21,37 @@ bucket_props_match(Props) ->
...
+    %% We just need one hash out of each list to match
+    item_matches(RemoteHashes, LocalHashes);
+test_prop_extra(UnknownProp, Type) ->
+    lager:info("repl does not know what to do to compare ~p for ~p",
+               [UnknownProp, Type]),
+    %% We don't know what to do with unknown properties, so to be safe
+    %% default to a failed comparison
+    false.
+
+item_matches(L1, L2) ->
```
JeetKunDoug: Can we rename this to something that describes what we're really doing - `at_least_one_match` perhaps?

src/riak_repl_keylist_server.erl
```diff
@@ -602,14 +605,38 @@ diff_bloom({diff_obj, RObj}, _From, #state{client=Client, transport=Transport,
...
+
+            EncodedObj = encode_robj(RObj, V),
+
+            _ = riak_repl_tcp_server:send(Transport, Socket, EncodedObj),
             ok
     end,
     {reply, ok, diff_bloom, State}.
 
+-spec encode_robj(riak_object:riak_object(), atom()) -> binary().
+encode_robj(O, WireVersion) ->
```
JeetKunDoug: Nit-picky, but reading {Partition, O} in this function makes O look like a 0 - can we rename `O` to `Obj` or `RObj` to make this easier to read?

src/riak_repl2_fssink.erl
```diff
@@ -151,6 +153,8 @@ handle_info({Proto, Socket, Data},
     case decode_obj_msg(Data) of
```
JeetKunDoug, line 153: NOTE: Did this refactor and pushed it, so no longer valid.

Trying to clean up the passing of {error, unknown_wire_format} all the way to `riak_repl_util:do_repl_put`, and found that refactoring decode_object_msg a bit, plus a change in this case statement, would allow us to isolate `do_repl_put` from the error condition.

If we rewrite decode_obj_msg a bit:

``` erlang
decode_obj_msg(Data) ->
    Msg = binary_to_term(Data),
    catch case Msg of
        {fs_diff_obj, BObj} when is_binary(BObj) ->
            RObj = from_wire_or_throw(BObj),
            {fs_diff_obj, RObj};
        {fs_diff_obj, _RObj} ->
            Msg;
        {PartitionIdx, W3ObjList} when is_binary(PartitionIdx) ->
            {ts, PartitionIdx,
             ts_to_robj(from_wire_or_throw(W3ObjList))};
        Other ->
            Other
    end.

from_wire_or_throw(BObj) ->
    case riak_repl_util:from_wire(BObj) of
        {error, unknown_wire_format} = Error ->
            throw(Error);
        Value -> Value
    end.
```

We can then add a new case to the case statement that starts where this line comment lives:

``` erlang
case decode_obj_msg(Data) of
        {fs_diff_obj, RObj} ->
            riak_repl_util:do_repl_put(RObj);
        {ts, _Partition, _RObj} = TSMsg ->
            riak_repl_util:do_repl_put(TSMsg);
        {error, unknown_wire_format} ->
            %% error has already been logged
            ok;
        Other ->
            gen_fsm:send_event(State#state.fullsync_worker, Other)
    end,
```

This removes the need for `riak_repl_util:do_repl_put` to handle the {error, _Error} case at all, and isolates the condition somewhere closer to where the error occurred.

src/riak_repl_util.erl
```diff
@@ -112,6 +115,12 @@ get_partitions(Ring) ->
     lists:sort([P || {P, _} <-
             riak_core_ring:all_owners(riak_core_ring:upgrade(Ring))]).
 
+do_repl_put({ts, _PartitionId, {error, _Error}}) ->
```
JeetKunDoug: With refactor to `riak_repl2_fssink` we can remove this case I believe. - I'll take care of this in the refactor commit.

src/riak_repl_util.erl
```diff
@@ -860,35 +878,36 @@ deduce_wire_version_from_proto({_Proto, {_CommonMajor, CMinor}, {_CommonMajor, H
...
+%% Typically, Cmd :: fs_diff_obj | diff_obj | <partition id>
+encode_obj_msg(ts, {Partition, RObj}) ->
+    encode_obj_msg(w3, {Partition, RObj}, undefined);
 encode_obj_msg(V, {Cmd, RObj}) when is_binary(RObj) ->
     encode_obj_msg(V, {Cmd, RObj}, undefined);
 encode_obj_msg(V, {Cmd, RObj}) ->
     encode_obj_msg(V, {Cmd, RObj}, riak_object:type(RObj)).
 
+encode_obj_msg(w3, {PartitionIdx, RObj}, _) ->
+    BObj = riak_repl_util:to_wire(w3,{PartitionIdx, RObj}),
```
JeetKunDoug: remove module name here - just call to_wire.

macintux:
It appears that all build failures are thanks to the `riak_kv` branch in question. Once that PR is merged, these should pass.

JeetKunDoug:
@borshop retry

JeetKunDoug:
:+1: e142d7d

macintux:
@borshop merge

PR #742, opened 2016-05-06
Fix supervisor PID leak in do_repl_put function

nickelization:
There are some issues with the riak_kv API which will be fixed in a
future release, but long story short, riak_kv_get_fsm:start_link doesn't
always actually end up linking the caller to the new FSM process. This
would cause the supervisor to end up with an endlessly growing list of
workers, since it had no way of seeing when a child died.

To fix this resource leak, we can simply switch to starting the process
directly, instead of using the supervisor API call. There's really no
reason to run these processes under a supervisor anyway since they're
normally started up under sidejob.

Note that in 2.2 this fix may be reworked again to use whatever new
APIs we end up adding/changing in riak_kv.

For more info, see basho/riak_kv#1178 (RIAK-2157)

### DIFF
--- a/src/riak_repl_util.erl
+++ b/src/riak_repl_util.erl
@@ -164,7 +164,7 @@ do_repl_put(Object, B, true) ->
             case riak_kv_util:is_x_deleted(Object) of
                 true ->
                     lager:debug("Incoming deleted obj ~p/~p", [B, K]),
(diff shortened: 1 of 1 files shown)


JeetKunDoug:
+1 1e3a569

nickelization:
@borshop merge

PR #744, opened 2016-06-15
Do not populate realtime queue if repl is not enabled

macintux:
Postcommit should bail if riak_repl2_rt has not validated that repl is enabled. Thanks to @erikleitch for identifying this problem.

Will create a test, so hold off on reviews.

### DIFF
--- a/src/riak_repl2_ts.erl
+++ b/src/riak_repl2_ts.erl
@@ -12,14 +12,20 @@
 postcommit({PartitionIdx, Val}, Bucket, BucketProps) when not is_list(Val) ->
     postcommit({PartitionIdx, [Val]}, Bucket, BucketProps);
 postcommit(PartitionBatch, Bucket, BucketProps) ->
-    maybe_postcommit(PartitionBatch, Bucket, proplists:get_value(repl, BucketProps, both)).
+    maybe_postcommit(PartitionBatch, Bucket,
+                     proplists:get_value(repl, BucketProps, both),
+                     application:get_env(riak_repl, rtenabled, false)).
 
-%% If `repl' is `false' or `fullsync', we skip realtime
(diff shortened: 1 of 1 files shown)


macintux:
Unit test failed, passes locally, not much information. Probably irrelevant.

macintux:
Ready for review.

src/riak_repl2_ts.erl
```diff
@@ -12,14 +12,20 @@
...
+    maybe_postcommit(PartitionBatch, Bucket,
+                     proplists:get_value(repl, BucketProps, both),
+                     application:get_env(riak_repl, rtenabled, false)).
 
-%% If `repl' is `false' or `fullsync', we skip realtime
-maybe_postcommit(_PartitionBatch, _Bucket, false) ->
+%% If `repl' is `false' or `fullsync', we skip realtime. Also if
+%% riak_repl2_rt has yet to set the `rtenabled' environment value to
+%% true, skip
+maybe_postcommit(_PartitionBatch, _Bucket, false, _Enabled) ->
```
JeetKunDoug, line 31: The order of parameters here is a bit difficult to discern - wondering if something like

``` erlang
maybe_postcommit(_PartitionBatch, _Bucket, _BucketRepl=false, _Enabled)
```

or, with typespecs:

``` erlang
-spec maybe_postcommit(PartitionBatch :: {index(), [term()], Bucket :: riak_core_bucket_type:bucket_type() ...)
...
```

Might make it a bit easier to see what's going on in each of the function heads. I'd vote for a spec with parameter names, just to get the spec in there (and maybe put the spec on postcommit/3 as well).

JeetKunDoug:
+1 8844c82

JeetKunDoug:
@borshop merge

PR #749, opened 2016-08-01
Switch to use newer get_fsm start API call

nickelization:
This ends up calling the same function in the end, but start_link does not actually link to the caller, and so shouldn't be used anymore since it has a misleading name.


bsparrow435:
+1 2a4ed90cb4ddc9cc5bc2bf0e6788d1a543cecf41

nickelization:
@borshop merge

PR #750, opened 2016-09-16
Rename old hashing algorithm to legacy

bsparrow435:


bsparrow435:
The dialyzer error is fixed with most recent riak_kv hashing PR.

PR #753, opened 2016-10-06
riak_repl2_pg: Fix invalid non-list status

ian-mi:
status output is expected to be a list. This commit changes the status output
format when the repl leader is undefined to match the format of the
other case.


bsparrow435:
merged.

PR #754, opened 2016-10-10
Revert "Integrate repl hook to avoid synchronizing expired objects [JIRA: RIAK-2695]"

nickelization:
Reverts basho/riak_repl#330 (RIAK-2438) - see that PR for more comments and information as to why we are reverting it.


bsparrow435:
:+1:

PR #770, opened 2017-02-24
Delete unused compose and multicompose functions

jvoegele:
The `compose/2` and `multicompose/1` functions in `riak_repl_ring` were not used anywhere. Better versions of these functions now live in `riak_core_util`.


JeetKunDoug:
@thumbot retry

Issue #775, opened 2017-12-22
Date conversion bug in _wm_stats


martincox:
In `riak_repl_wm_stats` a UTC datetime is passed into `httpd_util:rfc1123/1`, which expects a local date. It causes the stats to blow up when clocks go forward for DST. rfc1123 takes a local datetime and throws an exception when you pass a datetime that sits within the mystical non-existent hour when moving clocks forward. It should be handled gracefully, and there's a patch in OTP 20.1 for it. But we should be passing in a local time anyway. The net result is that we lose the ability to monitor for an hour during this time period.

```
(riak@127.0.0.1)1> Seconds = 63657709200. %% riak_core_util:moment() to produce current UTC seconds
63657709200
(riak@127.0.0.1)2> Date = calendar:gregorian_seconds_to_datetime(Seconds).
2017,3,26},{1,0,0
(riak@127.0.0.1)3> Formatted = httpd_util:rfc1123_date(Date).
exception error: no case clause matching []
in function httpd_util:rfc1123_date/1 (httpd_util.erl, line 344)
```

Fixed by converting to a local datetime with `calendar:universal_time_to_local_time/1`:

```
(riak@127.0.0.1)13> Seconds = 63657709200. %% riak_core_util:moment() to produce current UTC seconds
63657709200
(riak@127.0.0.1)14> Date = calendar:gregorian_seconds_to_datetime(Seconds).
2017,3,26,1,0,0
(riak@127.0.0.1)15> DateLocal = calendar:universal_time_to_local_time(Date).
2017,3,26,2,0,0
(riak@127.0.0.1)16> Formatted = httpd_util:rfc1123_date(DateLocal).
"Sun, 26 Mar 2017 01:00:00 GMT"
```

Think this is the same as described in basho/riak_core#773.

PR #776, opened 2017-12-22
Fix bug when passing utc timestamps into httpd_util:rfc1123/1.

martincox:
This function expects a localtime and throws an exception otherwise. This is evident when the clocks go forward for DST and a time falling in the skipped hour is passed into the function.

The fact that an exception is thrown is down to a bug in OTP which was patched in 20.1.


martincox:
@russelldb @Vagabond @binarytemple-external needs a quick review / merge if anyone can spare a couple of mins?

ghost:
I don't have review rights, but I'll give it a +1 

```
5> httpd_util:rfc1123_date(calendar:universal_time_to_local_time(DateTime)).
"Mon, 01 Jan 2018 00:00:00 GMT"
6> f(DateTime).                
ok
7> DateTime = {{2013, 9, 19}, {20, 51, 7}}.                                 
{{2013,9,19},{20,51,7}}
8> httpd_util:rfc1123_date(calendar:universal_time_to_local_time(DateTime)).
"Thu, 19 Sep 2013 20:51:07 GMT"
```

I guess from your commit message we can dump it later when everything is running on 20.2+ but that's the future, better to think in the present.

ghost:
fixes #775

russelldb (approved):
LGTM +1

PR #780, opened 2018-03-22
Fix sometime failing test

russelldb:
The test assumed that the shuffled order of the list would
be X because it assumed no other calls to random took place.
Maybe some recent changes added a call to `random`, but the
test fails on my ubuntu machine when `make eunit` is run.

This change calls `seed` in the test to ensure that "determinsitic random"
behaviour occurs for the test.

### DIFF
--- a/src/riak_repl_app.erl
+++ b/src/riak_repl_app.erl
@@ -383,19 +383,11 @@ unmask_address(IP, Mask, Size) ->
             unmask_address(IP, Mask, Size - 1)
     end.
 
+
 %%%%%%%%%%%%%%%%
 %% Unit Tests %%
 %%%%%%%%%%%%%%%%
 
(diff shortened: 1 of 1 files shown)


gwelr (commented):
As discussed, if you method is list_shuffle, the only valid test would be to prove it is correctly shuffle each time. This test try to prove the opposite. I would remove this test. If you absolutly need a test, run twice the list_shuffle call with the same In and make sure that Out1 and Out2 are different.

russelldb:
Yeah, I think since running it twice is only probably going to pass. It's a bad test. To copy Goldblum in JP: I was so preoccupied with whether or not I could fix it, I didn’t stop to think if I should.

gwelr (approved):

PR #781, opened 2018-04-08
Change ETS queue table permissions to protected

binarytemple:
Sometimes realtime replication stops working, the reason why it has stopped is unclear, diagnosis is further complicated by the fact that the ETS queue table (buffer of objects due to be replicated to the sink) is private (can only be seen by the `riak_repl2_rtq` `gen_server` instance).

If it were possible to examine the contents of that table the support engineer could examine the queued objects and perhaps identify a large object causing head of line blocking. At the moment the only option is to kill the process and let the supervisor restart it, destroying all useful information in the process. The diagnostic information is opaque and needlessly so. 

The table doesn't need to be changed to public as it's probably undesirable to modify state outside the public API for this gen_server. 

I propose that the permissions be changed to 'protected' in order to improve visibility.

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -13,7 +13,7 @@
         {ranch, "0.4.0-p1", {git, "git://github.com/basho/ranch.git", {tag, "0.4.0-p1"}}},
         {ebloom, ".*", {git, "git://github.com/basho/ebloom.git", {tag, "2.0.0"}}},
         {riak_kv, ".*", {git, "git://github.com/basho/riak_kv.git", {tag, "develop-2.2.5"}}},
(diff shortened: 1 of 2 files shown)


russelldb:
Not too late, was gonna yell "FREEZE!" on Monday. I'll double check it this afternoon, but it looks good to me, and makes good sense.

nickkeers:
This would be good, I think we've had the same frustration when trying to diagnose why the queue is blocking but haven't been able to do anything because the table is private

russelldb:
Just realised, this should be against 2.2, and 2.2.5 should have been deleted. It was just a temporary branch while I got the build working. Sorry @binarytemple, can you re-open against develop-2.2?

russelldb:
Closing in favour of #782 (branch target mess up, and @binarytemple is AFK for a few days so this is quicker)

PR #782, opened 2018-04-10
Change ETS queue table permissions to protected

russelldb:
NOTE: opened to replace #781 

Sometimes realtime replication stops working, the reason why it has stopped is unclear, diagnosis is further complicated by the fact that the ETS queue table (buffer of objects due to be replicated to the sink) is private (can only be seen by the `riak_repl2_rtq` `gen_server` instance).

If it were possible to examine the contents of that table the support engineer could examine the queued objects and perhaps discover that there is a large object causing head of line blocking. At the moment the only option is to kill the process and let the supervisor restart it, destroying all useful information in the process. The diagnostic information is opaque and needlessly so.

The table doesn't need to be changed to public as it's probably undesirable to modify state outside the public API for this gen_server.

I propose that the permissions be changed to 'protected' in order to improve visibility.

### DIFF
--- a/src/riak_repl2_rtq.erl
+++ b/src/riak_repl2_rtq.erl
@@ -13,7 +13,7 @@
 %% with the sequence number.  If multiple deliveries have taken
 %% place an ack of the highest seq number acknowledge all previous.
 %%
-%% The queue is currently stored in a private ETS table.  Once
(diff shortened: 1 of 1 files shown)


russelldb:
Reviewed and approved @ #781.

PR #784, opened 2018-05-28
removed failed connections from the pool

seanmcevoy:
Attempted fix for issue:
https://github.com/basho/riak_repl/issues/723
the pid of the dead process was being left in the 'pending' list.
i haven't figured out the original scenario that put it into this state but this bug would have been responsible for the repetitive retries and 'noproc' crashes.

### DIFF
--- a/.thumbs.yml
+++ b/.thumbs.yml
@@ -1,0 +1,11 @@
+minimum_reviewers: 2
+merge: true
+build_steps:
+  - make clean
+  - make deps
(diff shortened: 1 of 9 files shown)


martinsumner:
Does the `fail_request/3` call in L401 not remove the request from the pending list https://github.com/basho/riak_repl/blob/develop-3.0/src/riak_core_connection_mgr.erl#L654?

Issue #786, opened 2018-07-24
Incorrect return from `handle_info` calling log spam


russelldb:
Reported by @martinsumner https://github.com/basho/riak_repl/blob/develop/src/riak_core_service_conn.erl#L132-L134

The return is incorrect, should be `{stop, Reason, StateName}`

russelldb:
NOTE: there is another one on l128 too

martinsumner:
https://github.com/basho/riak_repl/blob/develop-2.9/src/riak_core_service_conn.erl#L128-L134

PR #787, opened 2018-07-24
GH786: Fix return values from handle_info

russelldb:
The incorrect stop message was causing log spam.

See #786


martinsumner:
+1

PR #789, opened 2019-01-18
Develop 2.9

martinsumner:
Adds

- Use of kvrequests refactor

- API for object touch (re-replicate an object)


martinsumner:
@martincox - develop-2.9 update for riak_repl

PR #790, opened 2019-02-12
Add OTP20/rebar3 support.

cmeiklejohn:
Replace use of random with rand.  Use `os:timestamp` where possible, encode request id's using the calling process's process identifier for uniqueness, replace usage of `rand_uniform`.  Use `gen_fsm_compat` for now to bridge `gen_fsm` compatibility changes before rewriting in `gen_statem`.

@martinsumner


src/riak_repl2_rtsource_conn.erl
```diff
@@ -463,7 +463,7 @@ send_heartbeat(State = #state{hb_timeout = HBTimeout,
 
     % Using now as need a unique reference for this heartbeat
     % to spot late heartbeat timeout messages
-    Now = now(),
+    Now = os:timestamp(),
```
llelf, line 466: The comment (two lines above) says it’s `now()` and it’s unique.

cmeiklejohn: If you look at the code, this is only used for estimating RTT for statistics.  I think it's fine, and it doesn't appear to require it to be monotonic nor globally unique.

Issue #799, opened 2019-10-07
Failure to complete full-sync on hitting soft retry limit


martinsumner:
Potentially related to - https://github.com/basho/riak_repl/issues/772

The test https://github.com/basho/riak_test/blob/develop-2.9/tests/repl_aae_fullsync_blocked.erl fails intermittently.

It fails when it uses an intercept to stop a full-sync from working on some vnodes, and checks the right number of vnode sync failures has occurred on completion.

When the test fails, it fails as full-sync is never considered complete.  The difference between success and failure is related to the ordering of the vnodes which the full-sync tries.  If the last vnode to be sync'd does sync OK (as it is not one with an intercepted function), then the test passes, and the correct number of vnodes failures are reported.  If the last vnode to be sync'd is one of those to not sync though, although the same work has completed/failed - the full-sync is never recorded as complete.

The cause of this appears to be that on hitting the soft retry limit https://github.com/basho/riak_repl/blob/24c6e8f408450cdf48b7259f9f0ef0778f94180d/src/riak_repl2_fscoordinator.erl#L561-L571 the function `maybe_complete_fullsync/2` isn't called.  Unlike on a hard failure - https://github.com/basho/riak_repl/blob/24c6e8f408450cdf48b7259f9f0ef0778f94180d/src/riak_repl2_fscoordinator.erl#L590 - and unlike on a success - https://github.com/basho/riak_repl/blob/24c6e8f408450cdf48b7259f9f0ef0778f94180d/src/riak_repl2_fscoordinator.erl#L447.

martinsumner:
https://github.com/basho/riak_repl/pull/800

PR #800, opened 2019-10-08
Mas 3.0 dialyzer

martinsumner:
Dialyzer fixes.

In gen_leader.erl the fix to dialyzer error on system_terminate/4 is clunky.  It is not clear why system_terminate/4 had no return, and how moving the code around in this way means there is no no problem.

Rolled in for src/riak_repl2_fscoordinator.erl is a fix to the intermittently failing test https://github.com/basho/riak_repl/issues/799.  This had been incorrectly ignored in previous releases (given that it was accepted there were known problems in aae full-sync in these test cases) - but know following this fix has predictable (passing) behaviour


src/gen_leader.erl
```diff
@@ -933,7 +934,7 @@ loop(#server{parent = Parent,
               _Msg when Debug == [] ->
                 handle_msg(Msg, Server, Role, E);
               _Msg ->
-                Debug1 = sys:handle_debug(Debug, {?MODULE, print_event},
+                Debug1 = sys:handle_debug(Debug, fun print_event/3,
```
ThomasArts: This is not exactly the same in the case of hot-code loading.

The original version would use the new print_event for each new event as soon as new code is loaded. The modified version will continue to use the original print_event function when new code is loaded. This means that it may continue doing so until the third version is loaded, after which the old function reference is removed and the loop crashes.

src/riak_repl2_fscoordinator.erl
```diff
@@ -566,9 +566,10 @@ handle_abnormal_exit(ExitType, Pid, _Cause, {value, PartitionWithSource, Running
                     Dropped = [Partition#partition_info.index | State4#state.dropped],
                     Purgatory = queue:filter(fun({P, _}) -> P =/= Partition end,
                                              State4#state.purgatory),
-                    {noreply, State4#state{error_exits = ErrorExits1,
-                                           purgatory = Purgatory,
-                                           dropped = Dropped}};
+                    maybe_complete_fullsync(Running,
```
ThomasArts, line 569: This is more than just a refactoring...

src/riak_repl2_fssink.erl
```diff
@@ -47,7 +47,7 @@
         cluster,
         fullsync_worker,
         work_dir = undefined,
-        strategy :: keylist | aae,
+        strategy :: keylist | aae | undefined,
```
ThomasArts, line 50: The documentation says:
```
%% If the protocol version >= 2.0, and the cluster has aae enabled and the replication
%% configuration stanza enables the aae strategy (the default is keylist),
```
In other words, it assumes a default. I would expect the init function to then implement that default by initialising the state with `keylist` alternatively to follow the logic.

At this moment there is a possible race condition when the server is started and a call `legacy_status` comes in, this may crash the server (line 107).  Dializer seems not to spot this, though.

src/riak_repl2_fscoordinator.erl
martinsumner: You are right, I was a bit lazy combining the two.  Shall I revert out the commits associated with https://github.com/basho/riak_repl/issues/799 and put in a separate PR?

src/gen_leader.erl
martinsumner: I don't think it would be a significant loss to not be able to hot load here.  Is there a better way though of resolving the dialyzer warning?

src/riak_repl2_fscoordinator.erl
ThomasArts: Thanks for the reference. This is fine.

src/gen_leader.erl
ThomasArts:
```suggestion
                Debug1 = sys:handle_debug(Debug, fun ?MODULE:print_event/3,
```

src/gen_leader.erl
```diff
@@ -1112,33 +1123,38 @@ reply({To, Tag}, Reply, #server{state = State} = Server, Role, E) ->
 handle_debug(#server{debug = []} = Server, _Role, _E, _Event) ->
     Server;
 handle_debug(#server{debug = Debug} = Server, _Role, E, Event) ->
-    Debug1 = sys:handle_debug(Debug, {?MODULE, print_event},
+    Debug1 = sys:handle_debug(Debug, fun print_event/3,
```
ThomasArts:
```suggestion
    Debug1 = sys:handle_debug(Debug, fun ?MODULE:print_event/3,
```

martinsumner: Thanks

ThomasArts (approved):

PR #806, opened 2019-10-31
Last gasp check

martinsumner:
Don't prompt changes on a ring_update is the updated ring is a last gasp of a local node closing down.

See basho/riak_core#945

This depends on this riak_core PR.


martinsumner:
@ThomasArts - this resolves another issue whereby there was incorrect action as a result of a lastgasp ring.  There were no known bad side effects other than log noise.

ThomasArts (approved):
I take it that the logic is the right way around, if not last gasp, then do the old thing.

ThomasArts (approved):

PR #812, opened 2021-06-04
Fix riak_repl_pb_get:get_cluster_id and :get, used in Repl communications with Riak CS

hmmr:
While doing renovation work on Riak CS, one test was found failing ([repl_v3_test](https://github.com/TI-Tokyo/riak_cs_test/blob/develop-3.0/tests/repl_v3_test.erl)). That test sets up two clusters and makes use of Repl proxy_get feature. Specifically, it makes two requests, `RpbReplCetClusterIdReq` and `RpbReplGetReq`, which are served by riak_repl_pb_get module. This PR clears up some bit rot to unbreak these requests.

There is an accompanying [PR in riak_pb](https://github.com/basho/riak_pb/pull/252).

### DIFF
--- a/src/riak_repl_pb_get.erl
+++ b/src/riak_repl_pb_get.erl
@@ -99,7 +99,7 @@ process(#rpbreplgetreq{bucket=B, key=K, r=R0, pr=PR0, notfound_ok=NFOk,
         make_option(basic_quorum, BQ) ++
         make_option(n_val, N_val) ++
         make_option(sloppy_quorum, SloppyQuorum),
-    case C:get(B, K, GetOptions) of
+    case riak_client:get(B, K, GetOptions, C) of
         {ok, O} ->
             make_object_response(O, VClock, Head, State);
         {error, {deleted, TombstoneVClock}} ->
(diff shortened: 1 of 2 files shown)


src/riak_repl_pb_get.erl
```diff
@@ -51,13 +51,23 @@ decode(_, Bin) ->
...
 %% @doc encode/1 callback. Encodes an outgoing response message.
-encode(#rpbgetresp{} = Msg) ->
+encode(Msg) ->
     {ok, riak_pb_codec:encode(Msg)}.
 
 %% Process Protocol Buffer Requests
 %%
 %% @doc Return Cluster Id of the local cluster
-process(#rpbreplgetclusteridreq{}, State) ->
+process(<<Code:8, Data/binary>>, State) ->
```
martinsumner, line 60: The pattern elsewhere is to do the decode at the decode step not by a pattern match at process e.g. https://github.com/basho/riak_kv/blob/develop-3.0/src/riak_kv_pb_aaefold.erl#L65-L66 or https://github.com/basho/riak_kv/blob/develop-3.0/src/riak_kv_pb_object.erl#L84-L98

Can this be made consistent with these other implementations?

hmmr: I am decoding it here because the message to be processed comes via [`riakc_pb_socket:tunnel`](https://github.com/TI-Tokyo/riak_repl_pb_api/blob/fd96ef21e2c9ab0d686e5a08c7d8944e9a2eb647/src/riak_repl_pb_api.erl#L39), which is [not decoded on receipt](https://github.com/basho/riak-erlang-client/blob/develop-3.0/src/riakc_pb_socket.erl#L2047) in riakc, hence the inconsistency.

martinsumner (approved):

PR #816, opened 2022-12-07
Mas i370 patch d

martinsumner:
Clean-up of old eqc tests of limited value.

As discussed with @ThomasArts there were two eqc tests that were largely testing either unused code, or just test code itself.  It was considered that the retained value of these tests were minimal, whereas the cost of repairing/updating the tests was significant.  On that basis, the tests, which have been failing for > 2 years of releases have been removed.


ThomasArts:
The tests in `rt_source_eqc` did not test anything sensible it seemed. The tests in `riak_repl_leader` do make a lot of sense. They start different Erlang nodes that participate in leader election and it is verified that the elected leader is indeed one of the expected leaders. Problem with these tests are that they dragged in a notion of `worker` that is not always connected. So tests fail.
The fact that these `riak_repl_leader` tests actually test an old version of the `riak_repl` and not `riak_repl2_leader`, they are of little value for the present setup. It would be valuable to create a similar, but working quickcheck model for leader section in the new version of the software. This however, can better be done from scratch than by adapting the existing, slightly old-fashion model.

ThomasArts (approved):

